PR #27, opened 2017-01-30
Fix issue 9

londumas:
fix issue 9, store the number of pairs in the cross-correlation

### DIFF
--- a/bin/do_xcf.py
+++ b/bin/do_xcf.py
@@ -188,7 +188,7 @@ if __name__ == '__main__':
     rp = (rps*wes).sum(axis=0)/wes.sum(axis=0)
     rt = (rts*wes).sum(axis=0)/wes.sum(axis=0)
     z = (zs*wes).sum(axis=0)/wes.sum(axis=0)
-    nb = (nbs*wes).sum(axis=0)/wes.sum(axis=0)
+    nb = nbs.sum(axis=0)
 
     out = fitsio.FITS(args.out,'rw',clobber=True)
     head = {}
--- a/bin/do_xcf.py
+++ b/bin/do_xcf.py
@@ -182,11 +182,13 @@ if __name__ == '__main__':
     rps=cfs[:,2,:]
     rts=cfs[:,3,:]
     zs=cfs[:,4,:]
+    nbs=cfs[:,5,:]
     cfs=cfs[:,1,:]
 
     rp = (rps*wes).sum(axis=0)/wes.sum(axis=0)
     rt = (rts*wes).sum(axis=0)/wes.sum(axis=0)
     z = (zs*wes).sum(axis=0)/wes.sum(axis=0)
+    nb = (nbs*wes).sum(axis=0)/wes.sum(axis=0)
 
     out = fitsio.FITS(args.out,'rw',clobber=True)
     head = {}
@@ -195,7 +197,7 @@ if __name__ == '__main__':
     head['NT']=xcf.nt
     head['NP']=xcf.np
 
-    out.write([rp,rt,z],names=['RP','RT','Z'],header=head)
+    out.write([rp,rt,z,nb],names=['RP','RT','Z','NB'],header=head)
     out.write([wes,cfs],names=['WE','DA'])
     out.close()
 
--- a/py/pylya/xcf.py
+++ b/py/pylya/xcf.py
@@ -34,6 +34,9 @@ def fill_neighs(pix):
             neighs = [q for p in npix for q in objs[p] if q.thid != d.thid]
             ang = d^neighs
             w = ang<angmax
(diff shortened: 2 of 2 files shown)


py/pylya/xcf.py
```diff
@@ -93,8 +98,9 @@ def fast_xcf(z1,r1,w1,d1,z2,r2,w2,ang):
     crp = sp.bincount(bins,weights=rp*we)
     crt = sp.bincount(bins,weights=rt*we)
     cz = sp.bincount(bins,weights=z*we)
+    cnb = sp.bincount(bins,weights=sp.ones(z.size))
```
ngbusca: remove keyword weights, they are 1 by default

py/pylya/xcf.py
```diff
@@ -54,20 +58,21 @@ def xcf(pix):
             zqso = [q.zqso for q in d.neighs]
             we_qso = [q.we for q in d.neighs]
 
-            cw,cd,crp,crt,cz = fast_xcf(d.z,d.r_comov,d.we,d.de,zqso,rc_qso,we_qso,ang)
+            cw,cd,crp,crt,cz,cnb = fast_xcf(d.z,d.r_comov,d.we,d.de,zqso,rc_qso,we_qso,ang)
             
             xi[:len(cd)]+=cd
             we[:len(cw)]+=cw
             rp[:len(crp)]+=crp
             rt[:len(crp)]+=crt
```
ngbusca: this should be rt[:len(crt)] += crt

py/pylya/xcf.py
```diff
@@ -54,20 +58,21 @@ def xcf(pix):
...
             z[:len(crp)]+=cz
```
ngbusca: this should be z[:len(cz)]+=cz

py/pylya/xcf.py
```diff
@@ -54,20 +58,21 @@ def xcf(pix):
...
+            nb[:len(crp)]+=cnb
```
ngbusca: this should be nb[:len(nb)]+=cnb

bin/do_xcf.py
```diff
@@ -182,11 +182,13 @@ def corr_func(p):
     rps=cfs[:,2,:]
     rts=cfs[:,3,:]
     zs=cfs[:,4,:]
+    nbs=cfs[:,5,:]
     cfs=cfs[:,1,:]
 
     rp = (rps*wes).sum(axis=0)/wes.sum(axis=0)
     rt = (rts*wes).sum(axis=0)/wes.sum(axis=0)
     z = (zs*wes).sum(axis=0)/wes.sum(axis=0)
+    nb = (nbs*wes).sum(axis=0)/wes.sum(axis=0)
```
ngbusca: remove the weights, the number of pairs should be an intenger. Then nb = nbs.sum(axis=0)

py/pylya/xcf.py
ngbusca: sorry, nb[:len(cnb)]+=cnb

londumas:
If we want to keep the fact that it "nb" is integer, we have to specify it in the FITS file, here
    out.write([rp,rt,z,nb],names=['RP','RT','Z','NB'],header=head)

py/pylya/xcf.py
```diff
@@ -43,6 +46,7 @@ def xcf(pix):
     rp = sp.zeros(np*nt)
     rt = sp.zeros(np*nt)
     z = sp.zeros(np*nt)
+    nb = sp.zeros(np*nt)
```
ngbusca: nb should be an int64: nb sp.zeros(np*nt,dtype=int64)

londumas:
It is declared as an integer scipy array but we don't store it has is in the FITS file, so it is store as a float array:
nb = vac[1]['NB'][:]
print nb

-> [  141.   345.   621. ...,  6724.  6904.  7010.]

ngbusca:
Can you test again whether it's saved as int or float?

londumas:
I just did.
It is saved as a float I think because:

nb = vac[1]['NB'][:]
print nb

-> [ 141. 345. 621. ..., 6724. 6904. 7010.]

londumas:
But sometime when dealing with big numbers, integers are tricky since they can loop around.
In my code I take only float.

ngbusca:
When f is a float you can have f + 1 = f. Ints go negative, so you'll know that you went overboard. But int64 is much bigger than the standard int in c, so it should be ok.

ngbusca:
I don't know how to specify the data type when writing with fitsio

londumas:
me neither, but it would be good to do it.

ngbusca:
I'll look into it

londumas:
I got it, it is because of the
cfs=sp.array(cfs)
we thus have to do
nbs=cfs[:,5,:].astype(sp.int64)

ngbusca:
indeed

Issue #18, opened 2017-01-25
Correct for flux calibration


londumas:
The pipeline has some mistakes due to calibration stars and skylines
(see figure 4 of auto-correlation paper: Bautista et al.).
A nice way to correct for this is to estimate it in the red side of the forest.

Do we want to have the possibility to correct the flux of forests by this effect.
Maybe by giving an asci or .fits file with:
lambda_obs correction err_correction
and correcting the flux and the ivar of the data

ngbusca:
This functionality is in the current baseline analysis but hasn't yet been ported to pyLyA (it's in the to-do list)

londumas:
ok, how do you see such a correction ?
Maybe feeding the code with another "iter.fits.gz" file.
This make more sense.

londumas:
It would take the field loglam and stack from a given file and correct with it.

Here what it looks like from the CIV forest
![correction_flux](https://cloud.githubusercontent.com/assets/3166436/22416974/f5e2c20a-e6d1-11e6-9d6b-3100cb896007.png)

ngbusca:
The correction should be read from the iter.fits.gz file that the user previously generated by running do_deltas on civ with the same choice of lambda-obs-min and lambda-obs-max, so as to have the same binning.

Then, as fluxes are read from the files, they should be divided by the correction. Note that this only corrects for multiplicative residuals.

londumas:
Agreed,
We could add a method to "forest", similar to "cont_fit".
i.e. using a interp1d to correct the flux and the invers variance (fl and iv) if the option is there.

What do you say ?

ngbusca:
In the __init__ of the class forest I'd add a keyword with the correction vector (None by default). If this vector is not None, then divide the flux by it (and multiply iv by this squared).

No interpolation, assume that the wavelength grids are the same (interpolating still leaves residuals).

londumas:
ok.
Is the grid for the stack the same as for the data ?
If yes then no problem, we just have to do a test on the  "iter.fits.gz" of calibration to see if it is on the same grid, with the same range of lambda_obs.

londumas:
We can also just add a correct_flux(self,correction) method to the class forest.
Where correction is the vector.

What do you prefer ?

londumas:
fixed in fix_issue_18

PR #39, opened 2017-02-03
fix issue 16, veto lines

londumas:
veto lines.

Tell me if that is what you want.

### DIFF
--- a/bin/do_deltas.py
+++ b/bin/do_deltas.py
@@ -120,10 +120,9 @@ if __name__ == '__main__':
     
 
     ### Veto sky lines
-    veto_line = None
     if (args.veto_line_file is not None):
-        veto_line = []
-        if True:
+        try:
+            veto_line = []
             with open(args.veto_line_file, 'r') as f:
                 loop = True
                 for l in f:
@@ -136,9 +135,8 @@ if __name__ == '__main__':
             for p in data:
                 for d in data[p]:
                     d.veto_lines(veto_line)
-        #except:
-        #    print(" Error while reading veto_line file {}".format(args.veto_line_file))
-        #    veto_line = None
+        except:
+            print(" Error while reading veto_line file {}".format(args.veto_line_file))
 
 
     if not args.dla_vac is None:
--- a/bin/do_deltas.py
+++ b/bin/do_deltas.py
@@ -84,8 +84,8 @@ if __name__ == '__main__':
     parser.add_argument('--nproc', type = int, default = None, required=False,
                     help = 'number of processors')
 
-    parser.add_argument('--veto-line-file',type = str,default=None,required=False,
-            help='Path to file to veto lines')
+    parser.add_argument('--mask-file',type = str,default=None,required=False,
+            help='Path to file to mask regions in lambda_OBS and lambda_RF. In file each line is: region_name region_min region_max  (OBS or RF)')
 
     args = parser.parse_args()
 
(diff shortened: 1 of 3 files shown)


londumas:
sory, I have to make some changes.

ngbusca:
where are these lines coming from?

londumas:
These are the sky lines and the Milky-Way lines.

londumas:
see plot in https://github.com/igmhub/pyLyA/issues/18

ngbusca:
I don't see any problem in #18

londumas:
In the plot of https://github.com/igmhub/pyLyA/issues/18,
you can see lines.
The problem is that these lines fluctuate, and so can't be exactly corrected
by the plot of issue_18.
That is why you still have a line in the red curve of figure 6 of Bautista et al.

A way to remove this effect is to remove some lines of the sky or of the MW.

londumas:
What is troubling you ?
In the analysis of the cross-correlation I veto some lines, so does the P_1D(k) analysis.

Some lines vary too much to be just corrected with the flux at the red side.

bin/do_deltas.py
```diff
@@ -84,6 +84,9 @@ def cont_fit(data):
     parser.add_argument('--nproc', type = int, default = None, required=False,
                     help = 'number of processors')
 
+    parser.add_argument('--veto-line-file',type = str,default=None,required=False,
```
ngbusca: I'd call this option mask-file. It can be used to avoid portions wider than just a line (e.g. the entire LyB peak)

bin/do_deltas.py
```diff
@@ -115,6 +118,27 @@ def cont_fit(data):
...
+
+    ### Veto sky lines
+    if (args.veto_line_file is not None):
+        try:
+            veto_line = []
+            with open(args.veto_line_file, 'r') as f:
+                loop = True
+                for l in f:
+                    l = l.split()
+                    veto_line += [ [float(l[1]),float(l[2])] ]
```
ngbusca: maybe usr_mask is a better name for veto_line. Also consider skipping lines starting with '#' so people can add comments

ngbusca: I'd read a fourth column indicating whether the mask is in lab or rest frame.

py/pylya/data.py
```diff
@@ -100,6 +100,18 @@ def __init__(self,h,thid,ra,dec,zqso,plate,mjd,fid,mode="pix"):
...
+            return
+
+        w = (self.ll<line[0,0]) | (self.ll>line[0,1])
+        for l in line[1:]:
+            w = w & ( (self.ll<l[0]) | (self.ll>l[1]) )
+
+        self.ll = self.ll[w]
+        self.fl = self.fl[w]
+        self.iv = self.iv[w]
+
```
ngbusca: treat also the case where the mask is in the rest frame

londumas:
I did apply your change request.
I had to change the "forest.mean_cont" (please see https://github.com/igmhub/pyLyA/pull/39/files).
because some bins can now empty and shouldn't be taken into account in interpolation.

londumas:
@ngbusca
for now if an exception is raised it continues without masking.

What do you prefere ?

py/pylya/data.py
londumas: done

ngbusca:
I think it's better to exit. If the user left the code running and went to have a coffee he might not realize there was a problem.

ngbusca:
Merci Hélion ! I think it looks good now. Is it ready to merge?

londumas:
I think so.
Do you agree with the correction to "forest.mean_cont" ?

ngbusca:
Yes, it's good

PR #52, opened 2017-02-18
fix issue 51

londumas:
definition of eta is inverse of papers

### DIFF
--- a/py/pylya/data.py
+++ b/py/pylya/data.py
@@ -157,7 +157,7 @@ class forest(qso):
 
         def chi2(p0,p1):
             m = model(p0,p1)
-            iv = self.iv/eta
+            iv = self.iv*eta
             we = iv/(iv*var_lss*m**2+1)
             v = (self.fl-m)**2*we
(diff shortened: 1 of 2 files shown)


ngbusca:
why do you want to change this? It was done on purpose.

londumas:
why was it done on purpose ?
Why change the definition of eta ?

ngbusca:
it's directly the slope of the fit sigma_forest vs 1/ivar.

londumas:
yes I understand that but why not call it iveta in the code and in the HDU ?

londumas:
sory ieta

ngbusca:
I don't see a point of changing this.

londumas:
This is just that then we have output that differ from precedent papers.
But this change nothing to the BAO obviously.
At the end it is your call

andreufont:
I'd just add a comment in the code saying
# this definition is the inverse of the one used in paper xxx

ngbusca:
I don't think it even makes sense to discuss this.

PR #58, opened 2017-03-10
fix_issue_45

londumas:
definitions of angmax are wrong


londumas:
Are you ok with this or you would like me to change some stuff ?

PR #68, opened 2017-03-16
fix issue 53

londumas:
https://github.com/igmhub/pyLyA/issues/53

range limit of sigma 2 LSS is too low


londumas:
Yes we can add a command line.
Do you want me to do it ?

ngbusca:
Sure

londumas:
done

londumas:
@ngbusca is it ok for you ?

bin/do_deltas.py
```diff
@@ -90,6 +90,16 @@ def cont_fit(data):
     parser.add_argument('--multiplicative-flux-calibration',type = str,default=None,required=False,
```
ngbusca: I should have caught this before, but can you change this option to --flux-calib ? It's too long now...

londumas:
done, you can merge

PR #75, opened 2017-03-17
Fix issue 71

londumas:
https://github.com/igmhub/pyLyA/issues/71

devide by zero in do_xdmat.py

### DIFF
--- a/bin/do_xdmat.py
+++ b/bin/do_xdmat.py
@@ -64,8 +64,8 @@ if __name__ == '__main__':
     parser.add_argument('--z-ref', type = float, default = 2.25, required=False,
                     help = 'reference redshift')
 
-    parser.add_argument('--rej', type = float, default = 1., required=False,
-                    help = 'reference redshift')
+    parser.add_argument('--rej', type = float, default = 0., required=False,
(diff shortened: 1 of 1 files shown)


bin/do_xdmat.py
```diff
@@ -64,8 +64,8 @@ def calc_dmat(p):
     parser.add_argument('--z-ref', type = float, default = 2.25, required=False,
                     help = 'reference redshift')
 
-    parser.add_argument('--rej', type = float, default = 1., required=False,
-                    help = 'reference redshift')
+    parser.add_argument('--rej', type = float, default = 0., required=False,
+                    help = 'fraction rejected: 0=no rejection, 1=all rejection')
```
ngbusca: this change would make the --rej option inconsistent with what's in do_dmat.... why do you want to do this change?

londumas:
It is just that I found more logic to have the code run by default instead of
cutting everything by default.
Anyway, I have put it back to 1.
If you agree, you can merge it.

ngbusca:
Ah, I see. I think it doesn't matter as long as the doc is clear. Is it not? If it's only the default value, I don't mind to change it.

PR #82, opened 2017-03-22
added handling for bi

ngbusca:
I've added the option  --bi-max in do_deltas to remove lines of sight with bi > bi-max. @londumas can you test before merging?


londumas:
@ngbusca couldn't you test it ? I am very busy right now.

ngbusca:
I did already, I wanted a second opinion ;^) But take your time

PR #89, opened 2017-03-29
add printing cuts

londumas:
add some comments printing the different cuts applied to the catalog.
Also look for BAL_FLAG_VI only if necessary


londumas:
tested and obviously doesn't change anything, but is safer, since possible to check the effect of all cuts.

PR #90, opened 2017-03-31
Fix issue 88

londumas:
fix issue 88,
i.e. remove mean_delta vs. lambda_obs
https://github.com/igmhub/pyLyA/issues/88

### DIFF
--- a/bin/do_cf.py
+++ b/bin/do_cf.py
@@ -143,9 +143,13 @@ if __name__ == '__main__':
     zs=cfs[:,4,:]
     cfs=cfs[:,1,:]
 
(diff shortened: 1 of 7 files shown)


ngbusca:
I agree that we need this fix, but maybe the code that does it could be more compact. Let me think about how to improve it

londumas:
seems compact to me, it only calls a method then remove the delta.
but if you find better, why not.

londumas:
any idea what could be changed ?

londumas:
@ngbusca could you take care of what we said last thursday ?
Or we could apply this merge and make it more beautiful another time

PR #93, opened 2017-04-03
fix crash when no object to correlate with pixels

londumas:
fix crash when no object to correlate with pixels given the range in r_perp and r_parallel

### DIFF
--- a/bin/do_xMetalGrid.py
+++ b/bin/do_xMetalGrid.py
@@ -151,7 +151,18 @@ if __name__ == '__main__':
     print("reading qsos")
 
     if (ra.size==0):
-        print(" No object to correlate with pixels")
+        out  = fitsio.FITS(args.out,'rw',clobber=True)
+        head = {}
+        head['RPMAX']          = xcf.rp_max
+        head['RTMAX']          = xcf.rt_max
+        head['NT']             = xcf.nt
+        head['NP']             = xcf.np
+        head['LAMBDAABS']      = args.lambda_abs
+        head['LAMBDAABSMETAL'] = args.lambda_abs_metal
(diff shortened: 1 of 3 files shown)


bin/do_xMetalGrid.py
```diff
@@ -150,6 +150,10 @@ def corr_func(p):
     pix = healpy.ang2pix(xcf.nside,th,phi)
     print("reading qsos")
 
+    if (ra.size==0):
+        print(" No object to correlate with pixels")
+        sys.exit(1)
```
ngbusca: Could you fix the error message to something more clear?

londumas:
"Given the range in rp, there are no objects to correlate with pixels, stop running"

is that better ?

ngbusca:
In fact, why is there a message needed at all? And why is the exit value = 1 (which implies there was an error) ? There's actually no problem, there are 0 pairs within the specified rp, rt, there's nothing special about it.

It's ok that the code exits, but it should result in an empty correlation function.

londumas:
I totally agree.

Currently it tries to read the list of objects and thus crash
sp.amin(zqso)

we can write en empty file if (ra.size==0) and then exit
would that be ok for you ?

ngbusca:
That makes much more sense. It should basically give the same thing as if we had actually run the whole code.

londumas:
ok, I take care of that

londumas:
The last changes work but ~20 lignes, we might think of a nicer and cleaner way to do the same thing

londumas:
I have found it, don't merge yet

londumas:
@ngbusca this solution is cleaner and works very well

PR #99, opened 2017-04-12
fix issue 57, storing number of pairs

londumas:
fix issue https://github.com/igmhub/pyLyA/issues/57 by storing number of pairs in do_cf.py


londumas:
@ngbusca , does it fit you ?

londumas:
is it ok ?

PR #111, opened 2017-06-15
X-correlation for forest

vserret:
Compute cf, cf1d, dmat and metal_dmat in the case of forest x-correlation. 
The codes are the same, just use options "in-dir2", "lambda-abs2" and "abs-igm_2" for the forest x-correlation.


ngbusca:
@vserret can you break this up into 4 separate pull requests with cf, cf1d, dmat and metal_dmat from 4 separate branches?

PR #131, opened 2017-06-29
replace constants.lya by constants.absorber_IGM[LYA], idem for lyb

vserret:


ngbusca:
@vserret I've also changed the lam_lya and lam_lyb in dla.py to use absorption_IGM["LYA"] etc.

PR #134, opened 2017-07-05
added rp_min. Permits to compute xi for rp <0

vserret:


py/pylya/cf.py
```diff
@@ -93,19 +94,20 @@ def cf(pix):
 def fast_cf(z1,r1,w1,d1,z2,r2,w2,d2,ang,same_half_plate):
     wd1 = d1*w1
     wd2 = d2*w2
-    rp = abs(r1-r2[:,None])*sp.cos(ang/2)
+    if rp_min<0 : rp = (r1-r2[:,None])*sp.cos(ang/2)
```
ngbusca: I'm not sure this is the right condition. You might want to have rp_min>0 and not take the abs()

vserret: if you have  rp_min>0 and you don't want to take the abs() there will be a problem since the rp grid will be positive. I'va done that to have the same results as before if rp_min>0. 
What do you think I should do ?

ngbusca: The problem is that the choice of sign of rp_min will change the result at rp>0... it doesn't quite make sense. You could use the x_correlation boolean that you defined earlier, no?

vserret: I could use : if (x_correlation) and (rp_min<0). Is that ok for you ?

ngbusca: why not just
```
 if x_correlation:
    rp = (r1-r2[:,None])*sp.cos(ang/2)
else:
    rp = abs(r1-r2[:,None])*sp.cos(ang/2)
```
? Why do you need the rp_min<0 condition?

vserret: because the computation of this correlation function is longer and for the moment I haven't modify the other tools ( dmat, do_fit, plotter...) so I want to be able to choose if I want abs() for the x-correlation

londumas:
Don't want to bring to much entropy but the auto-correlation of forests always use abs(rp) and the x-correlation of forest_1 - forest_2 should never use abs(rp).
So we should have:
 if x_correlation:
    rp = (r1-r2[:,None])*sp.cos(ang/2.)
else:
    rp = abs(r1-r2[:,None])*sp.cos(ang/2.)

vserret:
adjugé, vendu !

Issue #152, opened 2017-07-12
"print prior" not affected by verbose bool


londumas:
Even when verbose=False, we print information on the prior at each loop

londumas:
fixed in fix_issue_152

PR #154, opened 2017-07-12
print prior info only if verbose

londumas:


londumas:
https://github.com/igmhub/picca/issues/152

Issue #98, opened 2017-04-10
Add a dot to all float


londumas:
We have seen the bug with the wavelength, we should add a dot to all float values like "1." instead of "1".
Going to Python3 will solve the problem, but it is not tomorrow.

londumas:
fixed in fix_issue_98, only for files in bin/

PR #161, opened 2017-07-13
add a dot to floats in bin/*

londumas:


londumas:
https://github.com/igmhub/picca/issues/98

Issue #171, opened 2017-07-18
adding DLA is not verbose


londumas:
adding DLA doesn't give any information.
It would be nice to print some in order to see if everything goes well.

ngbusca:
Do you mean log the fact that there's a DLA catalog?

londumas:
no, I mean log something like:
"found a DLA in forest ...."

londumas:
I am trying to reproduce the DR12 cross result and had a bug linked to using the DR12_dla_catalog instead of the DR14

ngbusca:
sure, why not

londumas:
And it didn't tell me no DLA found, or something of the kind

ngbusca:
That's because the thing_ids in DR14 are different from those in DR12, so you won't find the DLAs

ngbusca:
It's not a bug ;^)

londumas:
I remeber that now.
I just wanted the code to say to me that no DLA were found

ngbusca:
Ok, that's a more general thing. E.g., one could use DR14Q and DR12Q for the delta field, or the DR12 vac, etc. We should find a way to issue a warning when different vacs are used (it might be on purpose).

ngbusca:
I meant DR14Q for the cross correlation and DR12Q for the delta field

londumas:
In the DLA catalog, some objects have a negative ThingId. Did you know that ?
Example:
['-00000001', '57006-07868-0875', '9.64039', '+0.6804', '2.584', '0', '0', '-1', '2.3', '2.3021', '20.27', '0.617', '-3.67', '0.36', '0.324', '0.246', '0.262', '0.603', '1.161', '01', '0.00', '0.00', '0.81', '-1.00', '0.95', '-1.00', '1.29', '-1.00', '1.48', '-1.00', '1.26', '-1.00', '1.31', '-1.00', '1.14', '-1.00', '3.21', '-1.00', '1.08', '-1.00', '0.00', '-1.00', '0.00', '-1.00']

ngbusca:
yes, that's life

londumas:
I find 24 of them. Did Pasquier told you what they are ?

ngbusca:
These are objects that have a spectrum (so you can decide whether there's a DLA or not) but that, for some reason, are lost after changes in the photometric extraction algorithms. So they "don't exist" anymore, so they don't have a RA, DEC.

londumas:
They do have a RA and dec:
[...]

PR #172, opened 2017-07-18
Fix issue 171

londumas:
add telling some information about DLA in do_delta
https://github.com/igmhub/picca/issues/171


londumas:
of curse.
Anyway I didn't change anything code wise

ngbusca:
It didn't look like that, but it's always good to test to make sure!

londumas:
Anyway I did test, because I agree, it is important.

Issue #174, opened 2017-07-19
strange code indentations


mblomqvist:
Some codes (for example cosmo.py) have strange indentations that mix tabs and spaces. Standard should be four spaces.

londumas:
indeed, could you fix that with a simple ctrl+F or a grep ?

ngbusca:
Unfortunately it's not that simple since there's a mixture of tabs and spaces...

londumas:
grep -P '\t' py/picca/*

ngbusca:
@londumas I don't think there's a simple solution, but if you think there's one, then it's all yours!

londumas:
fixed in fix_issue_174

PR #175, opened 2017-07-19
fix issue 174

londumas:
code indentation mixes tab and spaces
https://github.com/igmhub/picca/issues/174

Testing on full auto correlation and on full cross correlation

### DIFF
--- a/bin/do_fit
+++ b/bin/do_fit
@@ -17,15 +17,15 @@ parser = configargparse.ArgParser()
 
 parser.add('-c', '--config_file', required=False, is_config_file=True, help='config file path')
 for i in param.dic_init_float:
-	parser.add('--'+i,type=types.FloatType,required=False,help=param.help[i], default=param.dic_init_float[i])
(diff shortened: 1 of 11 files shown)


londumas:
I am testing on the full auto and cross fit.
I reveiwed the changes, but it would be good if you could also have a look @ngbusca .
I tell you when testing is complete

londumas:
gives the same result as master for cross fit and auto fit.
Ready to be merged

ngbusca:
did you look at the wedges as well?

londumas:
no, only to the chi2 and to the save.pars.
How could it be different if the both these files are the same ?

ngbusca:
I don't know...

ngbusca:
I meant the wedges as a means to verify the residuals.dat

londumas:
I just look at the wedgies, and the output are the same.
I can't be absolutely sure there is no bug introduced, but at least no major ones.

Issue #176, opened 2017-07-19
clone statment should have username in it


londumas:
To clone the repository, the best is to do:
```
 git clone https://<username>@github.com/igmhub/picca
```

ngbusca:
This is not how github recommends to clone, see:
https://help.github.com/articles/cloning-a-repository/

ngbusca:
Here's more info: https://help.github.com/articles/which-remote-url-should-i-use/

londumas:
The first link you give is when cloning a repository you own.

londumas:
If I do `https://github.com/igmhub/picca.git` then I will not be able to do `git push` since I am not 'igmhub' but 'londumas'

ngbusca:
@londumas they use your own repo as an example, but it applies to anything. What's the problem with the current recomendation from the readme?

ngbusca:
that is not true, when you push you will be asked for your username and password

ngbusca:
igmhub is not a user

londumas:
Indeed, this work.
I had a problem few days ago but it works now.

PR #177, opened 2017-07-19
update git clone instruction

londumas:
https://github.com/igmhub/picca/issues/176

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,8 +16,9 @@ requirements:
 
 download
 ```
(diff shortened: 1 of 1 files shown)


ngbusca:
Could you motivate this PR?

ngbusca:
@londumas do you agree? I don't understand the problem you are trying to fix...

londumas:
Yes I agree, I had a problem some days ago.
I tried from scratch and the problem disappeared

PR #184, opened 2017-08-03
Fix issue 183 b

londumas:
add the output files of the DR12+cross paper


londumas:
It would be nice to have the config files: fit.config

ngbusca:
@londumas we can discuss that later and add other things, but the priority now is to get these files in sync with the paper.

PR #187, opened 2017-08-31
Xdmat

vserret:


py/picca/cf.py
```diff
@@ -102,17 +102,21 @@ def fast_cf(z1,r1,w1,d1,z2,r2,w2,d2,ang,same_half_plate):
     w12 = w1*w2[:,None]
     z = (z1+z2[:,None])/2
 
-    w = (rp<rp_max) & (rt<rt_max) & (rp>rp_min)
+    if rp_min !=0:
+        w = (rp<rp_max) & (rt<rt_max) & (rp>rp_min)
+    else: 
+        w = (rp<rp_max) & (rt<rt_max) & (rp>=rp_min)
```
ngbusca: @vserret I don't understand this if, the second case works fine for any rp_min no?

vserret: no, if you want to have something really symmetric but I can change it

py/picca/cf.py
```diff
@@ -426,18 +440,21 @@ def fill_t123(r1,r2,ang,w1,w2,z1,z2,c1d_1,c1d_2,w123,t123_loc,same_half_plate):
...
     ba = bt + nt*bp
     we = w1[:,None]*w2
     zw = zw1[:,None]*zw2
 
-    w = (rp<rp_max) & (rt<rt_max)
+    w = (rp<rp_max) & (rt<rt_max) & (rp>=rp_min)
+
     if same_half_plate:
-        w = w & (bp>0)
+        w = w & (abs(rp)<(rp_max-rp_min)/np)
```
ngbusca: @vserret in this line you need a > not a <...

PR #191, opened 2017-09-11
fix issue 190

londumas:
https://github.com/igmhub/picca/issues/190


londumas:
add "configargparse" in readme

Issue #197, opened 2017-09-22
Implement alternative non-linear correction model


mblomqvist:
It would be good to add an option to the fitter to apply the Lya non-linear correction model of Arinyo et al 2015, even if the differences with the model of McDonald 2003 are relatively small.

mblomqvist:
Merged in master. Closed.

PR #198, opened 2017-09-27
Add alternative nlcorr

mblomqvist:
Add option to fitter to apply the non-linear correction model of Arinyo et al 2015 (default is the model of McDonald 2003). Model parameters are interpolated to any zref. Enhancement tested on auto and cross. Issue #197


mblomqvist:
Sure, the functions can be merged into one. Do you request any modifications of the top-level bool parameters (their names and/or implementation)?

ngbusca:
I would vote for changing the --dnl option to --dnl_{model} where model = "mcdonal" or "arinyo" (or some lower-case variant of that). This would break the backwards compatibility with the config files (I'm not too worried about it since we tagged the version for de Bourboux et al).

ngbusca:
@mblomqvist related: how do you feel about removing the dnl factor unless the --dnl_model flag is explicitly set?

mblomqvist:
Agreed. I will change the bool flags and make it such that no correction is the default, thus breaking backwards compatibility for old config files.

mblomqvist:
Branch ready to be merged with master pending review.

py/picca/fitter/cosmo.py
```diff
@@ -234,9 +270,11 @@ def getXiAuto(self,rp,rt,z,pk_lin,pars):
         sigmaNLper = pars["SigmaNL_perp"]
         sigmaNLpar = sigmaNLper*pars["1+f"]
         pk_nl = sp.exp(-(kp*sigmaNLpar)**2/2-(kt*sigmaNLper)**2/2)
-        pk_full*=pk_nl
-        if self.dnl :
-            pk_full*=self.DNL(k,muk)
+        pk_full *= pk_nl
+        if self.dnl_mcdonald :
+            pk_full *= self.DNL(k,muk,self.pk,self.q1_dnl,self.kv_dnl,self.av_dnl,self.bv_dnl,self.kp_dnl,"mcdonald")
```
ngbusca: here and in other places, wouldn't it be better to remove this if and call the function DNL with "self.dnl_method" as an argument?

mblomqvist: self.dnl_method is just a bool parameter, so one would have to pass two arguments: self.dnl_mcdonald and self.dnl_arinyo. I can do that.

ngbusca: you could simplify this by setting self.dnl_method to None by default and:

`
if self.dnl_method is not None:
    apply DNL(...,self.dnl_method)

ngbusca: and upgrade self.dnl_method to a string

mblomqvist:
Branch ready to be merged with master pending review.

ngbusca:
one last request :^P Can you replace the two options dnl_mcdonald and dnl_arynio by just one string  dnl_model which can be "mcdonald" or "arynio"? Finally, can you make the DNL function throw and exception if the model is unknown? Thanks!

mblomqvist:
Branch ready to be merged with master pending review.

ngbusca:
Here's a comparison of ap and chi2 

> master
> ap 1.02444734483 0.0331560336781
> chi2: 1596.1126238 1590 12 0.369550575394
> mcdonald 
> ap 1.02444734483 0.0331560336781
> chi2: 1596.1126238 1590 12 0.369550575394
> arinyo
> ap 1.02481116295 0.0336948717613
> chi2: 1593.17402388 1590 12 0.389316202408

ngbusca:
@mblomqvist I'll delete the branch on the server, you might want to do the same on your local repository

PR #201, opened 2017-10-04
set cont = 1e-10 when cont is negative

ngbusca:
This is a fix that effectively sets the continuum to 1e-10 if the continuum in a forest ever gets negative, so the forest effectively gets ignored.

### DIFF
--- a/bin/do_deltas.py
+++ b/bin/do_deltas.py
@@ -270,6 +270,8 @@ if __name__ == '__main__':
 
     log.close()
     for p in deltas:
+        if len(deltas[p])==0:
+            continue
         out = fitsio.FITS(args.out_dir+"/delta-{}".format(p)+".fits.gz",'rw',clobber=True)
         for d in deltas[p]:
             hd={}
--- a/bin/do_deltas.py
+++ b/bin/do_deltas.py
@@ -234,7 +234,6 @@ if __name__ == '__main__':
         for d in data[p]:
             assert hasattr(d,'ll')
 
-    log.close()
     for it in range(nit):
         pool = Pool(processes=args.nproc)
         print "iteration: ", it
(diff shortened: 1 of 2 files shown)


mblomqvist:
So the spectra are effectively ignored by setting their pixel weights to (near-)zero. It works for the end result, but the spectra are still in the sample, which means that correlation analyses include a bunch of zero-weight pixels. Would it not be better to completely remove these spectra after the final continuum iteration?

ngbusca:
@mblomqvist that's a good point, I'll remove them from the final sample and add that to the log.

ngbusca:
@mblomqvist do_deltas now prevents spectra with negative continua from being written in the output files. I've also added a check for failed fits (according to minuit). This info is logged in input.log file. In DR14 I get 1817 spectra rejected due to negative cont and 3010 rejected due to minuit failing.

ngbusca:
Summary of changes:

* in data.py:
  * added a check in cont_fit as to whether the continuum gets negative or minuit declares that the fit failed. In this case, the continuum is set to 10^-10 and p0 and p1 = 0
  * added a field bad_cont (set to None) to the data object. In any of the cases above, bad_cont is set to an error message that is used to log to the input.log file. Data objects with bad_cont not None are not written in the output delta files.

* in do_deltas.py:
  * when deltas are created from data objects, only data objects with bad_cont == None are used. 
  * the bad_cont error message is appended to the log file.

mblomqvist:
If a healpix p becomes empty due to bad continua, will it still be retained?

ngbusca:
@mblomqvist should be ok now

mblomqvist:
Looks good!

ngbusca:
Ok, I can't merge unless you start a review and approuve the changes!

PR #204, opened 2017-10-06
add order in bin/do_delta.py 

vserret:


ngbusca:
@vserret can you add a plot showing the correlation function for order = 1 and order = 0 both computed with this branch? On a few thousand spectra would be enough

vserret:
This is a plot showing the lya auto correlation computed with order 1 deltas (green) and order 0 deltas(blue). 

xi_order

londumas:
How come the blue is so different from figure 14 of https://arxiv.org/pdf/1702.00176.pdf ?

ngbusca:
@londumas I think it's just a few thousand spectra

londumas:
@vserret could you send with more forests so the difference can be better seen ?
Maybe all, so we can see what happens to the BAO peak.

ngbusca:
@londumas it might be better if you check out this branch and try to run it yourself on your side

londumas:
@ngbusca , obviously I don't want to add too much work on @vserret . I just mean that I am not sure that the last four plots are enough to see if there is a bug on this branch.

ngbusca:
@londumas : it's safer if someone else runs independent checks (one day we will have automatic tests...) I'll add you as a reviewer for this branch

vserret:
This is the comparison between master and delta_order branch. 
The correlation functions  are the same. 


xi_order_2

londumas:
I looked at the implementation of the order and it looks good.
The only problem is that currently the error on "p1" is zero. Could you change that by setting to "0.1" for example ?

ngbusca:
@londumas Thanks for catching that! It's a bug that was already fixed in master, strange that it came back. In master the error in p1 is set to p0/2.

londumas:
No, in master it is set to "p1/2", but p1=0, so it is set to 0.

ngbusca:
@londumas your master is not up to date

londumas:
I just looked at the master on the website

ngbusca:
@londumas sorry, you are right. I screwed that up in my latest merge. @vserret can you fix it?

londumas:
https://github.com/igmhub/picca/blob/master/py/picca/data.py

PR #207, opened 2017-10-10
Save errors on eta and vlss

mblomqvist:
Extract the errors on the parameters eta and vlss and save to the output fits file. Tested on subsamples of Lya and CIV spectra.

### DIFF
--- a/bin/do_deltas.py
+++ b/bin/do_deltas.py
@@ -248,14 +248,14 @@ if __name__ == '__main__':
         if it < nit-1:
             ll_rest, mc, wmc = prep_del.mc(data)
             forest.mean_cont = interp1d(ll_rest[wmc>0.], forest.mean_cont(ll_rest[wmc>0.]) * mc[wmc>0.], fill_value = "extrapolate")
-            ll,eta,vlss,nb_pixels = prep_del.var_lss(data,(args.eta_min,args.eta_max),(args.vlss_min,args.vlss_max))
+            ll,eta,eta_err,vlss,vlss_err,nb_pixels = prep_del.var_lss(data,(args.eta_min,args.eta_max),(args.vlss_min,args.vlss_max))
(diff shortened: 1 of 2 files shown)


ngbusca:
@mblomqvist the code looks good. Do the errors make sense ?

mblomqvist:
Here are the results for 10,000 CIV spectra. Errors are small but seem reasonable at least in a relative sense.
eta_civ
vlss_civ

londumas:
how does it compare to sub-sampling errors by dividing the ~300,000 CIV forests into ~100 samples ?

mblomqvist:
Actually, the errors seem rather small compared to what was derived using my own code on the full data set. Is there a particular reason why iminuit would estimate unreliable errors?

londumas:
yes, I think if the chi2 isn't defined well enough. For example if it doesn't take care of the correlation between pixels. Did you have a look at the result of the subsampling errors ?

mblomqvist:
No, I haven't looked at the scatter from subsampling. Feel free to have a crack at it. Since eta and vlss are quantities derived from the data that are applied to calculate the pixel weights, it seemed reasonable to check how well they are determined. Igmhub isn't really the right forum to interpret the results.

ngbusca:
@mblomqvist I agree with @londumas that if we want to add the uncertainties they should be meaningful. If you really want this change I'd suggest that you do a dedicated study on the chi2 of the eta and sigma_lls fits. In Busca et al (I think it's the only place where we showed uncertainties in these parameters) I calculated the uncertainties from the variance in the mocks....

mblomqvist:
Meaningful errors are obviously important. You know better than I do what are the limitations of the picca implementation and iminuit. The fact of the matter is that Busca et al is the only place where eta and vlss have been presented, which was a long time ago for DR9 and 15 mocks. An update of the results would be useful at some point.

PR #209, opened 2017-10-20
refactored eta-vlss fit

ngbusca:
This branch addresses and issue reported offline by @mblomqvist which showed that for qso x civ the variance from picca was worse than for other codes. I tracked the problem down to a bias in the estimation of eta and vlss and refactored this fit to correct for this bias: I essentially switched to a log binning of the weights instead of a linear binning.

### DIFF
--- a/bin/do_deltas.py
+++ b/bin/do_deltas.py
@@ -130,6 +130,7 @@ if __name__ == '__main__':
 
     forest.var_lss = interp1d(forest.lmin+sp.arange(2)*(forest.lmax-forest.lmin),0.2 + sp.zeros(2),fill_value="extrapolate")
     forest.eta = interp1d(forest.lmin+sp.arange(2)*(forest.lmax-forest.lmin), sp.ones(2),fill_value="extrapolate")
+    forest.fudge = interp1d(forest.lmin+sp.arange(2)*(forest.lmax-forest.lmin), sp.zeros(2),fill_value="extrapolate")
     forest.mean_cont = interp1d(forest.lmin_rest+sp.arange(2)*(forest.lmax_rest-forest.lmin_rest),1+sp.zeros(2))
 
     ### Fix the order of the continuum fit, 0 or 1. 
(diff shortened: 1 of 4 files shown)


ngbusca:
Here's the plot for the ratio of variance for this branch.

image

mblomqvist:
Excellent. What do you measure for vlss?

ngbusca:
@mblomqvist I've added the eta, vlls data to iter.fits so one can check that the fit worked out (check out hdu[4]). Could you test this branch on your side? I want to do a few more tests on my end before merging.

mblomqvist:
The code works well and gives consistent eta and vlss with my CIV analysis. Which additional tests are you considering before merging?

ngbusca:
I want to try other forests (lya, lyb, Si, etc.) to make sure it doesn't break somewhere else. I'm working on it...

ngbusca:
Since Busca et al. we've noticed that at low var-pipe the noise model is not very good, most likely due to the fact that the mean continuum doesn't capture the shape variability. So I've added a parameter to the noise model to capture that. The new model is:
`
var_del = eta * var_pipe + var_lls + fudge / val_pipe
`

Here's an example of this fit for lyman-alpha and the lowest loglam_obs bin:

image

mblomqvist:
Good. For CIV, fudge is about 1e-7, so a fairly small contribution.

mblomqvist:
What does the modification of do_xdmat.py do?

ngbusca:
@mblomqvist values for the fudge of 1e-7 are typical, the fudge dominates only at very large values of ivar.

The modification of do_xdmat.py has nothing to do with the pull request, but I couldn't live with that part of the code so poorly written ;^)

ngbusca:
@mblomqvist don't forget to click approve if you do :)

mblomqvist:
Didn't know that all the additional tests were done ;)

ngbusca:
I'm happy on my end but you are the reviewer (i.e. if the universe collapses after merging, it will be your fault)

PR #212, opened 2017-10-30
Fix issue 211

ngbusca:
This PR fixes issue 211. The problem was that the new fudge factor was extrapolated to negative values (unphysical) values. I changed the interpolation scheme for eta, var_lss and fudge to "nearest" which makes these interpolations more robust and fixes the issue.


mblomqvist:
The do_deltas analysis runs fine for the tested forests Lyb, Lya, SiIV, CIV, and eta values are OK. Note that the last lambda_obs bin with data is based on a comparatively small number of spectra, so its uncertainty is large. It should not be a problem.
eta_civ_siiv_lya_lyb

mblomqvist:
All cross-correlations look fine too.

PR #222, opened 2017-11-07
Fitter2

ngbusca:
This is a new version of the fitter that can handle multiple data sets. Some features:

* support for multiple data sets
* uses sparse matrices for metals and distorsion (requested offline by @julienguy ). This gives better performance both on cpu and memory
* options are handled by .ini files and the ConfigParse package
* output is packed in .h5 files instead of gazillion of ascii
* the code is much more compact and allows the user to define new models for pk or xi in a simple way
* It uses decorators for cacheing the calculation of the correlation function instead of the infamous recalc variable used in the previous fitter.


ngbusca:
To do:
* add SigmaNL
* add alpha_iso/epsilon fit
* add mcmc sampling
* add fast MC
* add cached version of xi_drp

londumas:
@ngbusca, do you think it would be possible to change "QSO" in names of variables into "OBJ", since we try to have a general fitter. Do you use the "lya" for variables ?

ngbusca:
@londumas In the new fitter the user specifies the names of tracer1 and tracer2. So you can have anything you want.

ngbusca:
@londumas the current change in xmetal dmat is not the final state. What I want is that the user slecifies the name of tje OBJ tracer, e.g. DLA, QSO, GAL, ...

londumas:
OK,  i thought it was ready to be merged. I leave you to it.

Le 10 nov. 2017 10:08 PM, "Nicolas Busca" <notifications@github.com> a
écrit :

ngbusca:
@vserret can you explain what's breaking for you forest1 x forest2 analysis? Your commit breaks the fitter on qso x lya so we'll have to fix that...

vserret:
When we write the metal dmat for forest x forest, we use : "RP_{}_{}".format(met1,met2)
and this was not the case before
We could write "if (self.tracer1==self.tracer2) or (self.tracer2=='QSO')" ?

ngbusca:
@vserret ok, I'll take a look. We might have to change the way we write the met_dmat, the ifs are a bad solution.

mblomqvist:
A few comments about (possible) modifications/additions:
* Change the name "lls" -> "hcd"
* Missing functions for cross-correlations: HCD, UV and combinations with QSO velocities (Lorentzian/Gaussian)
* DNL restricted to the Lya auto-correlation. That might be reasonable but breaks backwards compatibility with DR12.
* DNL Arinyo model uses parameters set by the user. Is that better than tabulating them?
* Get zref from the Pk template header
* Change the name "alpha" -> "gamma"

ngbusca:
@mblomqvist 

> Change the name "lls" -> "hcd"

agreed

> Missing functions for cross-correlations: HCD, UV and combinations with QSO velocities (Lorentzian/Gaussian)

agreed

> DNL restricted to the Lya auto-correlation. That might be reasonable but breaks backwards compatibility with DR12.

I've had mixed answers from the "experts" regarding this issue. Truth is that no one has done simulations to measure DNL from halos cross-correlated with LyA. Our anzats "sqrt(DNL)" doesn't have any physical basis.

> DNL Arinyo model uses parameters set by the user. Is that better than tabulating them?

Also true for McDonald. Might we ever want to fit for those parameters one day ? I'm still thinking about how to handle that (e.g. have an option to set them to the paper values by default).

> Get zref from the Pk template header

good idea.

> Change the name "alpha" -> "gamma"

why not!

londumas:
@ngbusca, in do_metal_dmat, it should not be `"RP_QSO_"+ai`, but `"RP_{}_{}".format(args.obj,m)` as in py/picca/fitter2/data.py. So we would add the name of the object when calculating do_metal_dmat

londumas:
@ngbusca, could you explain why it starts fitting (migrad) only the bias, then print everything, then refit every parameter (migrad)? Is it a trick to go faster?

ngbusca:
@londumas I agree, and this is how it's done in the fitter2 branch.

ngbusca:
> in do_metal_dmat, it should not be "RP_QSO_"+ai, but "RP_{}_{}".format(args.obj,m) as in py/picca/fitter2/data.py. So we would add the name of the object when calculating do_metal_dmat

I agree. This is a quick hack that I'll improve

ngbusca:
> could you explain why it starts fitting (migrad) only the bias, then print everything, then refit every parameter (migrad)? Is it a trick to go faster?

Yes, the biases are fit very fast since the FFTs are calculated only once. The the full minimization starts with the amplitudes of the correlation function more or less correct.

londumas:
@ngbusca, Could you add a hess method to the fitter2/chi2 ?
I get very different best fit results when changing the starting value for the fitted parameter
I am not asking to put it in bin/fitter2 however.

Also I like the fact to have a `zref` parameter (maybe it make more sense in chi2.ini however).
With this I can get the result of the measured bias at the given redshift.

What do you think?

ngbusca:
> @ngbusca, Could you add a hess method to the fitter2/chi2 ?
I get very different best fit results when changing the starting value for the fitted parameter
I am not asking to put it in bin/fitter2 however.

can you give an example of this?

> Also I like the fact to have a zref parameter (maybe it make more sense in chi2.ini however).
With this I can get the result of the measured bias at the given redshift.

I don't understand what you mean by this.

londumas:
Ok, I found the problem. It was because beta and bias were 1-1 correlated in the case of very low signal.

For the zref, basically I would like to be able to set it to "0.5", so I get the results at that redshift, not at the one of the input power spectrum.

ngbusca:
@londumas the zref is the redshift of the power spectrum, it's not a free parameter (I actually need to change that). I can add a parameter to rescale the power spectrum to another redshift. But note that the growth function at z~0.5 is non-trivial, since Omega_L becomes important, so we need to think how that changes things.

londumas:
>I can add a parameter to rescale the power spectrum to another redshift.

This would be very usefull, something like zeff

>But note that the growth function at z~0.5 is non-trivial, since Omega_L becomes important, so we need to think how that changes things.

Yes, I know this, I will have a look at the possible parametrizations

londumas:
It seems there are only two cases where we should replace `/` by `//` in fitter2:
```
xi[ell/2,:]*=L(mur,ell)
xi[ell/2,:]=sp.interpolate.splev(sp.log(ar),spline)
```
Could you still check with a `grep "/" -r py/picca/fitter2/*`?
And apply these changes?

londumas:
@ngbusca, I think we should remove the array absorber_IGM in /py/picca/fitter2/utils.py?
What do you think?

londumas:
@ngbusca, a useful thing would be giving not only the chi2 of each set of data, but also the "ndata", and why not the number of free parameters for this set: "npar". Indeed it would help see how fval is divided into the different chi2

londumas:
@ngbusca, why not add `if not hasattr(self, "fast_mc"): return` in `def fastMC(self)` ?
Currently, using bin/fitter2 without fastMC in chi2.ini file craches

ngbusca:
@londumas it doesn't in my local (uncommited) version. Be patient!

ngbusca:
> @ngbusca, why not add if not hasattr(self, "fast_mc"): return in def fastMC(self) ?
Currently, using bin/fitter2 without fastMC in chi2.ini file craches

@londumas can you update your branch and try again? It doesn't crash on my side...

londumas:
@ngbusca, I am on the latest commit.
It seems to be an issue regarding an aspect of the data. The code fails to perform the cholesky decomposition

Here is the error:
```
  File "/uufs/astro.utah.edu/common/home/u6011908/Programs/igmhub/picca/bin/fitter2", line 10, in <module>
    chi.fastMC()
  File "/uufs/astro.utah.edu/common/home/u6011908/Programs/igmhub/picca/py/picca/fitter2/chi2.py", line 86, in fastMC
    d.cho = cholesky(d.co)
  File "/uufs/chpc.utah.edu/sys/installdir/python/2.7.11-c7/lib/python2.7/site-packages/numpy-1.12.1-py2.7-linux-x86_64.egg/numpy/linalg/linalg.py", line 612, in cholesky
    r = gufunc(a, signature=signature, extobj=extobj)
  File "/uufs/chpc.utah.edu/sys/installdir/python/2.7.11-c7/lib/python2.7/site-packages/numpy-1.12.1-py2.7-linux-x86_64.egg/numpy/linalg/linalg.py", line 93, in _raise_linalgerror_nonposdef
    raise LinAlgError("Matrix is not positive definite")
numpy.linalg.linalg.LinAlgError: Matrix is not positive definite
```

ngbusca:
@londumas yep, if cholesky fails it means that the covariance is not positive definite. If the number of subsamples is small this will certainly be the case. If you are looking for a qualitative fit you could just use the diagonal of the covariance (provided all the diagonal elements are strictly positive)

londumas:
@ngbusca, ok, that what I was thinking on my bike. Should you check if the matrix is positive definite when loading the data?

ngbusca:
@londumas the fastest way to check is by doing a cholesky decomposition. If it crashes then it's not positive definite. Picca assumes the user knows what she's doing, meaning that she would (among many other things) check that her covariance makes sense.

mblomqvist:
@ngbusca What is the status of the new fitter? Can you post an updated to-do list?

ngbusca:
@mblomqvist There are two main things missing: the chi2 scans and the fact that I've twicked the calculation of the metal matrix that needs to be uniformized with another branch being developed by @vserret .

mblomqvist:
@ngbusca The latest version seems to still be missing certain pk functions, such as HCD for the cross-correlation etc.

mblomqvist:
@ngbusca I was just wondering if there is an estimated time frame for when the cross-correlation pk functions can be added, and the chi2 scan too. Do you know?

ngbusca:
I should have time to work on this during the Garching meeting...

ngbusca:
should we merge and finish the missing changes later?

londumas:
@ngbusca, yes.

ngbusca:
@londumas can you give me a review?

py/picca/fitter2/chi2.py
```diff
@@ -0,0 +1,141 @@
+from __future__ import print_function
+import numpy as np
```
londumas: To be consistent with the rest of picca, could you use scipy as sp?

py/picca/fitter2/chi2.py
```diff
@@ -0,0 +1,141 @@
...
+import iminuit
+import time
+import h5py
+from numpy.linalg import cholesky
```
londumas: same with scipy?

py/picca/fitter2/data.py
```diff
@@ -0,0 +1,213 @@
...
+
+        r_min = dic_init['cuts']['r-min']
+        r_max = dic_init['cuts']['r-max']
+
+        mu_min = dic_init['cuts']['mu-min']
+        mu_max = dic_init['cuts']['mu-max']
+
+        ## select data within cuts
+        mask = (rp > rp_min) & (rp < rp_max)
+        mask = mask & (rt > rt_min) & (rt < rt_max)
```
londumas: `mask &= (rt > rt_min) & (rt < rt_max)` is more beautiful

py/picca/fitter2/utils.py
```diff
@@ -0,0 +1,117 @@
...
+
+from . import myGamma
+
+nmuk = 1000
+muk=(sp.arange(nmuk)+0.5)/nmuk
+dmuk = 1./nmuk
+muk=muk[:,None]
+
+def sinc(x):
+    return sp.sin(x)/x
```
londumas: why not using scipy.sp?
It is more stable in x~=0

py/picca/fitter2/utils.py
```diff
@@ -0,0 +1,117 @@
...
+    emm=N*fft.fftfreq(N)
+    r=r0*sp.exp(-emm*l/N)
+    dr=abs(sp.log(r[1]/r[0]))
+    s=sp.argsort(r)
+    r=r[s]
+
+    xi=sp.zeros([ell_max/2+1,len(ar)])
+
+    for ell in range(0,ell_max+1,2):
+        pk_ell=sp.sum(dmuk*L(muk,ell)*pk,axis=0)*(2*ell+1)*(-1)**(ell/2)
```
londumas: be careful with `/` versus `//`.
It will not work in Python3

py/picca/fitter2/utils.py
```diff
@@ -0,0 +1,117 @@
...
+        xi_loc[-1]=0
+        spline=sp.interpolate.splrep(sp.log(r)-dr/2,sp.real(xi_loc),k=3,s=0)
+        xi[ell/2,:]=sp.interpolate.splev(sp.log(ar),spline)
+
+    return xi
+
+def Pk2Xi(ar,mur,k,pk,ell_max=None):
+    xi=Pk2Mp(ar,k,pk,ell_max)
+    for ell in range(0,ell_max+1,2):
+        xi[ell/2,:]*=L(mur,ell)
```
londumas: same here `/` versus `//`

py/picca/fitter2/utils.py
```diff
@@ -0,0 +1,117 @@
...
+    beta2 = kwargs["beta_{}".format(tracer2)]
+    bias2 = kwargs["bias_{}".format(tracer2)]
+    bias2 *= growth_rate/beta2
+
+    return bias1, beta1, bias2, beta2
+
+    
+
+
+### Absorber names and wavelengths
```
londumas: maybe no need to have these, we can remove it and only read from `py/picca/constants.py`

py/picca/fitter2/chi2.py
ngbusca: ok

py/picca/fitter2/chi2.py
ngbusca: agreed. The scipy.linalg should be slightly more performant.

py/picca/fitter2/data.py
ngbusca: beauty!! essential

py/picca/fitter2/utils.py
ngbusca: A while ago I did profiling tests (with the cProfile module)  on the fitter. The results:
  * the fitter spends half the time calculating the sinc function, so even more than the fft (which is almost the rest of the time)
  * I've done tests a while ago at Utah and found that sinc was slower than sin(x)/x by about 50%, probably because it tests whether the argument is zero. 
 * We never evaluate sinc at exactly zero

I plan to redo some profiling tests at nersc, but for now I prefer to leave my sinc function.

py/picca/fitter2/utils.py
ngbusca: What does a simple `/` do in python3 when the operands are integer and the divisor is a factor of the dividend (as is the case here since ell is always even)? Does it cast the result to float? I can change to `//` but I'd prefer the code to break rather than spitting out the wrong result...

py/picca/fitter2/utils.py
ngbusca: same as before, `ell` is always even.

py/picca/fitter2/utils.py
ngbusca: yes, you are right.

ngbusca: actually this wasn't used for anything, I've erased it.

ngbusca:
@londumas I think I've introduced all the requested changes. Can you take a look again?

py/picca/fitter2/data.py
```diff
@@ -0,0 +1,213 @@
+import fitsio
+from functools import partial
+import numpy as np
+from numpy import linalg
```
londumas: same here with numpy

py/picca/fitter2/data.py
```diff
@@ -0,0 +1,213 @@
+import fitsio
+from functools import partial
+import numpy as np
```
londumas: same here with numpy

py/picca/fitter2/parser.py
```diff
@@ -0,0 +1,87 @@
+from __future__ import print_function
```
londumas: no need, your print is compatible with Python3

py/picca/fitter2/pk.py
```diff
@@ -0,0 +1,158 @@
+import numpy as np
```
londumas: same here for numpy

py/picca/fitter2/utils.py
londumas: ok, thnk you for the answer

py/picca/fitter2/xi.py
```diff
@@ -0,0 +1,131 @@
+import numpy as np
```
londumas: same for numpy

py/picca/fitter2/parser.py
ngbusca: in python2 it will print the parenthesis. With this import it will have the same behaviour in python2 and python3

py/picca/fitter2/pk.py
ngbusca: done

py/picca/fitter2/xi.py
ngbusca: sure thing!

ngbusca:
@londumas can you check again?

londumas:
@ngbusca, only one `numpy` in `py/picca/fitter2/xi.py` and then ok to merge.
Seems good!

ngbusca:
@londumas there's only one instance of numpy left (from numpy import fft), but the fft package in scipy has a different syntax than that from numpy. So I'll leave the numpy fft.

ngbusca:
@londumas you need to approve the PR

Issue #225, opened 2017-11-08
allow to run on one file instead of all


londumas:
The correlation I try to compute is very long and takes a lot of memory.
A solution is to compute it individually on each delta-100.fits.gz and then combine the results.
For that I would need to replace 
```
fi = glob.glob(args.in_dir+"/*.fits.gz")
```
by
```
if (len(args.in_dir)>8) and (args.in_dir[-8:]==".fits.gz"):
    fi = glob.glob(args.in_dir)
 else:
    fi = glob.glob(args.in_dir+"/*.fits.gz")
```

would that be ok @ngbusca ? The current behavior wouldn't change by this commit.

londumas:
merged

PR #226, opened 2017-11-08
fix issue 225

londumas:
https://github.com/igmhub/picca/issues/225
add the possibility to run on a single file.

Tested and works


londumas:

Issue #228, opened 2017-11-08
cross correlation in do_cf1d doesn't work


londumas:
The cross correlation of two forests from the same quasar doesn't work.
It gives the 1d auto correlation of the first file.

londumas:
fixed

Issue #227, opened 2017-11-08
do_cf1d is computing twice


londumas:
it seems to me that the code is run twice on line 134 (or 135) and on line 138
```
    if x_correlation: 
        keys = []
        for i in data.keys(): 
            if i in data2.keys(): 
                keys.append(i)
        cfs = map(cf1d,keys)
    else: cfs = map(cf1d,data.keys())

    #pool = Pool(processes=args.nproc)
    cfs = map(cf1d,data.keys())
#pool.close()
```

ngbusca:
Indeed. Also, why is the multiprocessing commented out ?

⁣Sent from BlueMail ​

londumas:
I don't know, it was there when I arrived :)
What should we put back ?

londumas:
fixed

PR #229, opened 2017-11-08
fix issue 227 and 228

londumas:
fix issue 227 and 228:
https://github.com/igmhub/picca/issues/228
https://github.com/igmhub/picca/issues/227

do_cf1d computes only one time the correlation and works for the cross corelation

Here is the SiIV x CIV 1d cross correlation

capture d ecran de 2017-11-08 12-03-52

### DIFF
--- a/bin/do_cf1d.py
+++ b/bin/do_cf1d.py
@@ -137,8 +137,8 @@ if __name__ == '__main__':
         for i in data.keys(): 
             if i in data2.keys(): 
                 keys.append(i)
-        cfs = map(cf1d,keys)
-    else: cfs = map(cf1d,data.keys())
+        cfs = pool.map(cf1d,keys)
+    else: cfs = pool.map(cf1d,data.keys())
 
     pool.close()
 
--- a/bin/do_cf1d.py
+++ b/bin/do_cf1d.py
@@ -97,6 +97,7 @@ if __name__ == '__main__':
 
     cf.npix = len(data)
     cf.data = data
+    cf.ndata = ndata
 
     x_correlation=False
     if args.in_dir2: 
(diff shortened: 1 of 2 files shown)


londumas:
still some changes to be done

londumas:
The 1d auto and the 1d cross LYA give the same results:
in green the cross and in blue the auto
capture d ecran de 2017-11-08 13-40-53

londumas:
The only thing is that for the cross of different forests, the diagonal is zero, so all the 1d arrays are zero "v1d','wv1d','nv1d','c1d','nc1d','nb1d".
The data are then only in the 2d arrays.

Here is an example of the SIIV_CIV 1d cross correlation
capture d ecran de 2017-11-08 13-46-59

londumas:
ready to be merged, can you review @vserret

londumas:
@ngbusca, what are the lines I should put back ?

ngbusca:
@londumas the map should be a pool.map to use multiprocessing. Otherwise it runs on a single thread. So you need to uncomment the definition of pool = Pool(...) and move it to before the if x_correlation block. The replace the two map calls by pool.map. Please remove any commented out lines as well. Please test again before pushing that you get the same results as above.

ngbusca:
@londumas thanks for catching this.

londumas:
@ngbusca, how does this look like ?
I tested and it still gives the same results.
If ok, you can merge.

bin/do_cf1d.py
```diff
@@ -139,6 +140,8 @@ def cf1d(p):
         cfs = map(cf1d,keys)
     else: cfs = map(cf1d,data.keys())
```
ngbusca: @londumas you still need to replace map() by pool.map() otherwise it doesn't change anything ;^)

londumas: just did and indeed now it multithread

ngbusca: still getting the same results?

londumas:
yes, still getting the same results, and I did test it.

Issue #232, opened 2017-11-10
add iminuit errors for eta, var_lss and fudge


londumas:
add iminuit errors for eta, var_lss and fudge.
It is useful to be able to get rid of invalid results in presenting the different results.
Also it will be useful for comparing with the sub-samplig errors if we need to.

londumas:
fixed is fix_issue_232

londumas:
merged

PR #233, opened 2017-11-10
fix issue 232

londumas:
add errors of eta, var_lss, fudge
https://github.com/igmhub/picca/issues/232

### DIFF
--- a/bin/do_deltas.py
+++ b/bin/do_deltas.py
@@ -257,7 +257,7 @@ if __name__ == '__main__':
     res = fitsio.FITS(args.iter_out_prefix+".fits.gz",'rw',clobber=True)
     ll_st,st,wst = prep_del.stack(data)
     res.write([ll_st,st,wst],names=['loglam','stack','weight'])
-    res.write([ll,eta,vlss,fudge,nb_pixels,err_eta,err_vlss,err_fudge],names=['loglam','eta','var_lss','fudge','nb_pixels'])
+    res.write([ll,eta,vlss,fudge,nb_pixels],names=['loglam','eta','var_lss','fudge','nb_pixels'])
     res.write([ll_rest,forest.mean_cont(ll_rest),wmc],names=['loglam_rest','mean_cont','weight'])
     var = sp.broadcast_to(var.reshape(1,-1),var_del.shape)
     res.write([var,var_del,var2_del,count,nqsos,chi2],names=['var_pipe','var_del','var2_del','count','nqsos','chi2'])
--- a/bin/do_deltas.py
+++ b/bin/do_deltas.py
@@ -249,7 +249,7 @@ if __name__ == '__main__':
         if it < nit-1:
             ll_rest, mc, wmc = prep_del.mc(data)
             forest.mean_cont = interp1d(ll_rest[wmc>0.], forest.mean_cont(ll_rest[wmc>0.]) * mc[wmc>0.], fill_value = "extrapolate")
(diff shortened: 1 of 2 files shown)


mblomqvist:
This is the same modifications I proposed a few weeks ago...

ngbusca:
Isn't this exactly what @mblomqvist added a few weeks ago? If you guys want this option, that's fine with me, but It would be good to take the chance to study a bit the errors, not just add blindly what minuit reports. @mblomqvist what do you think?

londumas:
I added a hess calculation, it gives very different results than the minos one.
Also it is important to remove bins where the fit failed when plotting.

londumas:
@ngbusca, I agree. But studying the error isn't a part of picca.

ngbusca:
@londumas if picca reports an error the burden is on picca to make sure it's not senseless...

ngbusca:
@londumas otherwise it's very simple, we just return random numbers for everything.

londumas:
@ngbusca, the code gives the minos error. A user could decide to see the sub-samples errors.
This is up to him, the code will not give this to him.
I imagine you would like me to compare these two types of errors.
Am I right ?

ngbusca:
@londumas The first question is whether the chi^2 values are good. If they aren't, I wouldn't trust any estimation (Minuit, Hess, Minos...). So it would be great to look at subsamples and compare the errors from subsample to what minuit reports ( I wouldn't include minos or hesse since they are less stable and we don't want that to crash).

Once we have a good idea of how the subsample errors compare to minuit, then we just add a cautionary note on the README. But we need to do this study.  If you don't want to/can't do it and @mblomqvist agrees to add this, then I'll just merge.

ngbusca:
What I don't understand is that you want to include this option, but don't want to/don't have time (etc). do the study. This means that you don't really need the errors on these parameters, so why bother adding that to the code? I know it's easy, but there are better ways to use your time ;^)

londumas:
@ngbusca, I don't need the error on the parameters, however I need to remove bins where the fit failed.
That is why I did this commit. For the part where I have:
```
+            eta[i] = 1.
+            vlss[i] = 0.1
+            fudge[i] = 1.*fudge_ref
+            err_eta[i] = 0.
+            err_vlss[i] = 0.
+            err_fudge[i] = 0.
```

londumas:
So what about merging this and adding an issue to picca in order to study this error, and give cautions if need be into README.md.

ngbusca:
I wan't to hear also from @mblomqvist

mblomqvist:
If one wants to show a plot of eta(z) or vlss(z) in a paper, having reasonable error bars is important. I assumed that they were, but understood from your comments that they might not be. I guess I have a mild preference for adding the errors to the output file, but agree that they need to be studied.

ngbusca:
I'm actually leaning towards not having them. This way, the next person who actually "needs" them will have to do the study (and then we can put them in in a well understood way). Otherwise, she might just blame the eta/vlss errors on picca and move on.

mblomqvist:
Fair enough.

londumas:
ok, then just let me leave the part that does:
```
+            eta[i] = 1.
+            vlss[i] = 0.1
+            fudge[i] = 1.*fudge_ref
```
this is useful to remove bins where the fit failed

ngbusca:
@londumas ok, I'll take a look

londumas:
@ngbusca, just tested this version, and it works.

londumas:

Issue #234, opened 2017-11-10
Correct var_pipeline before running do_delta


londumas:
The idea is to do something similar as we do for flux calibration:
use another forest to correct for the errors in the definition of the var_pipeline.

londumas:
Here is the correction from the MgII forest (see defintion at the end)

![eta_correction_from_mgii](https://user-images.githubusercontent.com/3166436/32681287-0052e4b0-c62c-11e7-83b8-e2f50d678565.png)


```
--in-dir          /uufs/astro.utah.edu/common/home/u6011908/Data/Spectra/ForetFusion/Repeats/
--drq             /uufs/astro.utah.edu/common/home/u6011908/Data/Catalogs/DR14Q_v3_1.fits
--out-dir         /uufs/astro.utah.edu/common/home/u6011908/Run_programs/igmhub/picca/DR14/Delta_MGII/Delta/
--iter-out-prefix /uufs/astro.utah.edu/common/home/u6011908/Run_programs/igmhub/picca/DR14/Delta_MGII/Log/delta_attributes
--log             /uufs/astro.utah.edu/common/home/u6011908/Run_programs/igmhub/picca/DR14/Delta_MGII/Log/input.log
--flux-calib      /uufs/astro.utah.edu/common/home/u6011908/Run_programs/igmhub/picca/DR14/Delta_calibration_rebin2/Log/delta_attributes.fits.gz
--mask-file       /uufs/astro.utah.edu/common/home/u6011908/Programs/igmhub/picca/etc/dr14-line-sky-mask.txt
--rebin 1
--mode pix
--lambda-min 3600.0
--lambda-max 7235.0
--lambda-rest-min 2000.0
--lambda-rest-max 2600.0
--nit 10
--bi-max 0.

```

londumas:
Here is the result for the SiIV forest:

![eta_corrected_from_mgii_to_siiv](https://user-images.githubusercontent.com/3166436/32682377-8599742c-c631-11e7-8b0e-3ac24f32fc63.png)

londumas:
fixed in fix_issue_234

londumas:
merged

PR #236, opened 2017-11-10
Fix issue 234

londumas:
Correction of the ivar before running do_delta

https://github.com/igmhub/picca/issues/234

### DIFF
--- a/bin/do_deltas.py
+++ b/bin/do_deltas.py
@@ -94,7 +94,7 @@ if __name__ == '__main__':
     parser.add_argument('--flux-calib',type = str,default=None,required=False,
             help='Path to previously produced do_delta.py file to correct for multiplicative errors in the pipeline flux calibration')
 
-    parser.add_argument('--ivar-pipeline-calib',type = str,default=None,required=False,
+    parser.add_argument('--ivar-calib',type = str,default=None,required=False,
             help='Path to previously produced do_delta.py file to correct for multiplicative errors in the pipeline inverse variance calibration')
 
     parser.add_argument('--eta-min',type = float,default=0.5,required=False,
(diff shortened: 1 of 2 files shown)


bin/do_deltas.py
```diff
@@ -154,6 +155,18 @@ def cont_fit(data):
...
             sys.exit(1)
 
+    ### Correct multiplicative pipeline inverse variance calibration
+    if (args.ivar_calib is not None):
+        try:
+            vac = fitsio.FITS(args.ivar_calib)
+            ll  = vac[2]['LOGLAM'][:]
+            eta = vac[2]['ETA'][:]
+            forest.correc_ivar = interp1d(ll,eta,fill_value="extrapolate")
+
```
ngbusca: @londumas note that the eta interpolation was recently changed to kind="nearest". Can you change that here too?

londumas: whar is the difference ? Do you want me to change it too for What is the difference ? Do you want me to change it too for flux-calib ?

londumas: is there a reason why it is not done here ?
```
forest.mean_cont = interp1d(ll_rest[wmc>0.], forest.mean_cont(ll_rest[wmc>0.]) * mc[wmc>0.], fill_value = "extrapolate")
```

londumas: I am not sure it is better to have "nearest", it seems more natural to have "linear"

ngbusca: I agree it's more natural, but the nearest option is more stable when the nieghboring bins are empty or very noisy, and we need stability there. That's not that much of an issue for the mean cont or the mean abs because bins have a very large occupation and the edges are not very important.

londumas:
ok, the change is done.
In the future we could remove empty bins before defining the interpolation

ngbusca:
I like this option, but I'd like input from @mblomqvist and @vserret . What bothers me is that we can use the CIV forest to correct LyA but then what do we use to correct CIV? To me, the fact that one can correct the LyA forest with the CIV forest is a "proof of concept" of the extraction. Once that's settled, one can use any forest to correct any other forest, and the easiest is to correct each forest with itself.

londumas:
I agree, this can be viewed as a proof of concept but:
    -> To correct the forests, you can use the MgII or even the red of MgII where there seems to be no absorption, when looking at the xi1d.
    -> Also, I like the fact to correct a forest before running do_delta. Because corrections from the forest itself gets very small.

mblomqvist:
Applying flux-calib or ivar-calib is fine, but the result is not a flat line after the correction anyway. So what is really accomplished?

londumas:
It removes a first order correction.

mblomqvist:
But picca fits for eta again, so the initial calibration step seems superfluous. Unless there is a numerical reason which makes eta more reliable/stable when it is closer to unity.

ngbusca:
I think @londumas should try again with the kind="nearest" fix, otherwise the interpolation schemes are not compatible between the calculation and the correction

londumas:
Here it is with nearest:

capture d ecran de 2017-11-13 09-05-41

londumas:

Issue #235, opened 2017-11-10
fudge is not in py/picca/data.py/from_forest


londumas:
At the end of do_delta.py, the delta are given by from_forest(),
"fudge" is missing from the definition of the data

```
    def from_forest(cls,f,st,var_lss,eta):

        de = f.fl/f.co/st(f.ll)-1.
        mst = st(f.ll)
        co = f.co
        ll = f.ll
        iv = f.iv/eta(f.ll)
        we = iv*(co*mst)**2/(iv*(co*mst)**2*var_lss(f.ll)+1)
        co = f.co
        return cls(f.thid,f.ra,f.dec,f.zqso,f.plate,f.mjd,f.fid,ll,we,co,de,f.order)
```

londumas:
The definition should use:
```
def variance(var,eta,var_lss,fudge):
    return eta*var + var_lss + fudge/var
```

londumas:
```
    def from_forest(cls,f,st,var_lss,eta,fudge):

        ll = f.ll
        mst = st(ll)
        var_lss = var_lss(ll)
        eta = eta(ll)
        fudge = fudge(ll)
        co = f.co
        de = f.fl/(co*mst)-1.
        var = 1./d.iv/(co*mst)**2
        we = 1./variance(var,eta,var_lss,fudge)

        return cls(f.thid,f.ra,f.dec,f.zqso,f.plate,f.mjd,f.fid,ll,we,co,de,f.order)
```

londumas:
fixed in fix_issue_235

PR #237, opened 2017-11-11
fix issue 235

londumas:
fudge is not in py/picca/data.py/from_forest
https://github.com/igmhub/picca/issues/235

### DIFF
--- a/bin/do_deltas.py
+++ b/bin/do_deltas.py
@@ -266,7 +266,7 @@ if __name__ == '__main__':
     deltas = {}
     data_bad_cont = []
     for p in data:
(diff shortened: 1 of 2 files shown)


londumas:
do_delta.py runs normally, I will just check the 1d auto-correlation for any changes

londumas:
It produces a slightly higher variance.
Here is a plot of the 1d correlation for the Lya forest with the diagonal and the first two diagonals.


xi1d_lya_before_now_diag012

londumas:
Here is for the CIV forest, doesn't change much...

xi1d_civ_before_now_diag012

PR #240, opened 2017-11-15
Fix issue 238

londumas:
add reading spframe and spcframe.
Plots for comparison are coming

### DIFF
--- a/py/picca/data.py
+++ b/py/picca/data.py
@@ -65,20 +65,22 @@ class forest(qso):
         qso.__init__(self,thid,ra,dec,zqso,plate,mjd,fid)
 
         ## cut to specified range
-        w = (ll>forest.lmin)
+        bins = sp.floor((ll-forest.lmin)/forest.dll+0.5).astype(int)
+        ll = forest.lmin + bins*forest.dll
+        w = (ll>=forest.lmin)
         w = w & (ll<forest.lmax)
         w = w & (ll-sp.log10(1.+self.zqso)>forest.lmin_rest)
         w = w & (ll-sp.log10(1.+self.zqso)<forest.lmax_rest)
         w = w & (iv>0.)
         if w.sum()==0:
             return
-        ll=ll[w]
-        fl=fl[w]
-        iv=iv[w]
+        bins = bins[w]
+        ll = ll[w]
+        fl = fl[w]
+        iv = iv[w]
 
         ## rebin
-        bins = ((ll-forest.lmin)/forest.dll+0.5).astype(int)
-        ll = forest.lmin + sp.arange(bins.max()+1)*forest.dll
+        cll = forest.lmin + sp.arange(bins.max()+1)*forest.dll
         cfl = sp.zeros(bins.max()+1)
         civ = sp.zeros(bins.max()+1)
         ccfl = sp.bincount(bins,weights=iv*fl)
@@ -86,22 +88,13 @@ class forest(qso):
         cfl[:len(ccfl)] += ccfl
         civ[:len(cciv)] += cciv
         w = (civ>0.)
-        cfl[w] /= civ[w]
-        fl = cfl
-        iv = civ
-
-        ## cut to specified range
-        w = (ll>forest.lmin)
-        w = w & (ll<forest.lmax)
-        w = w & (ll-sp.log10(1.+self.zqso)>forest.lmin_rest)
-        w = w & (ll-sp.log10(1.+self.zqso)<forest.lmax_rest)
-        w = w & (iv>0.)
         if w.sum()==0:
             return
-        ll=ll[w]
-        fl=fl[w]
-        iv=iv[w]
+        ll = cll[w]
+        fl = cfl[w]/civ[w]
+        iv = civ[w]
 
+        ## Flux calibration correction
         if not self.correc_flux is None:
             correction = self.correc_flux(ll)
             fl /= correction
@@ -122,7 +115,7 @@ class forest(qso):
         fl = sp.append(self.fl,d.fl)
         iv = sp.append(self.iv,d.iv)
 
-        bins = ((ll-forest.lmin)/forest.dll+0.5).astype(int)
+        bins = sp.floor((ll-forest.lmin)/forest.dll+0.5).astype(int)
         cll = forest.lmin + sp.arange(bins.max()+1)*forest.dll
         cfl = sp.zeros(bins.max()+1)
         civ = sp.zeros(bins.max()+1)
--- a/py/picca/data.py
+++ b/py/picca/data.py
@@ -103,6 +103,25 @@ class forest(qso):
         self.iv = iv
         self.order=order
 
+    def __add__(self,d):
+
+        ll = sp.append(self.ll,d.ll)
+        fl = sp.append(self.fl,d.fl)
+        iv = sp.append(self.iv,d.iv)
+
+        bins = ((ll-forest.lmin)/forest.dll+0.5).astype(int)
+        cll = forest.lmin + sp.unique(bins)*forest.dll
+        civ = sp.bincount(bins,weights=iv)
+        civ = civ[bins.min():]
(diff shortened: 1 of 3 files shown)


londumas:
The code gives exactly the same results in pix mode

londumas:
I am looking at the spframe and spcframe now

londumas:
To read spfram I would have to to use the interpolation to get the wavelength solution.
I changed the code to no longer read spframe.

londumas:
Here are some plots for 10,000 CIV forests. I am currently running the full data set, but it takes a lot of time.
The spcframe data are a lot more noisy:

flux_vs_lobs



eta_vs_lobs

varlss_vs_lobs

fudge_vs_lobs

flux_vs_lrf


cf1d_var

cf1d_cor

londumas:
After correction of the bug we find similar results, apart for eta.

flux_vs_lobs

eta_vs_lobs

varlss_vs_lobs

fudge_vs_lobs

flux_vs_lrf

cf1d_var

cf1d_cor

londumas:
@ngbusca, it is ready to be merged.
We can add spframe another time if need be.

py/picca/data.py
```diff
@@ -63,25 +63,25 @@ class forest(qso):
...
     def __init__(self,ll,fl,iv,thid,ra,dec,zqso,plate,mjd,fid,order):
         qso.__init__(self,thid,ra,dec,zqso,plate,mjd,fid)
-        ## rebin
 
+        ## rebin
         bins = ((ll-forest.lmin)/forest.dll+0.5).astype(int)
         w = bins>=0
         fl=fl[w]
         iv =iv[w]
         bins=bins[w]
```
ngbusca: let's take the opportunity to clean up this code a bit. Can you add a cut on forest.lmin and forest.lmax earlier? 
e.g.:

    w = (ll>forest.lmin) & (ll<forest.lmax)
    ll = ll[w]
    fl = fl[w]
    iv = iv[w]

    bins = ((ll-forest.lmin)/forest.dll+0.5).astype(int)

etc...

londumas: We can't remove it from after rebining since sometime the last bin has its original ll<lmax, but its rebined one ll>lmax. However, we can add cutting also before rebining. Do you want this ?

ngbusca: right, in the snippet of code above, the cut is before the definition of `bins = ...`, so yes, before rebinning. Thanks!

londumas:
Hi updated the plot with all sample of 307132 qsos

londumas:
I have updated the plots with the full sample.

londumas:
@ngbusca, I did your changes and it works.
However I have a trouble with `ll = forest.lmin + sp.arange(bins.max()+1)*forest.dll` because then the first bin has forest.lmin wavelength. Shouldn't it be something like `ll = forest.lmin + (0.5+sp.arange(bins.max()+1))*forest.dll`

ngbusca:
I don't think so, because you are adding 0.5 to the bin number:

    bins = ((ll-forest.lmin)/forest.dll+0.5).astype(int)

so the way the lambda grid is setup already gives you the bin centers. But I realize now that there's a problem now with the first and last bin which only get contributions from a half bin. To fix that we would have to move the selection lmin lmax selection to after the bins are calculated and use floor to correctly handle the edges:

    bins = floor((ll-forest.lmin)/forest.dll+0.5).astype(int)

and then select:
    nbins = floor((forest.llmax-forest.lmin)/forest.dll+0.5).astype(int)
    w = (bins >= 0) & (bins < nbins)

what do you think?

londumas:
> the way the lambda grid is setup already gives you the bin centers

I am not sure about that, because we store ll in fits file and then the redshift is defined as:
`z = 10**d.ll/args.lambda_abs-1.`

ngbusca:
right, so that gives you the redshift of the bin center, what's the problem?

londumas:
The wavelength are given by `ll = forest.lmin + sp.unique(bins)*forest.dll`,
so `ll` isn't the bin center it is the bin min edge

ngbusca:
The way the binning is done:

    bins = floor((ll-forest.lmin)/forest.dll+0.5).astype(int)

(in particular adding the 0.5) means that the pixel centers are at 
`ll = forest.lmin + sp.unique(bins)*forest.dll`

londumas:
with `ll = forest.lmin + sp.unique(bins)*forest.dll`, the first pixel has bins=0, so ll = forest.lmin.
so the first pixel is at the bin min edge.

ngbusca:
yes, the first pixel has bins=0 but it gets contributions from abs(ll - forest.lmin)<dll/2

londumas:
ok, I see.

py/picca/data.py
```diff
@@ -63,31 +63,41 @@ class forest(qso):
...
-        cfl = sp.bincount(bins,weights=iv*fl)
-        cfl = cfl[bins.min():]
-        w=civ>0
-        cfl[w]/=civ[w]
-        iv = civ
-        fl = cfl
+        iv=iv[w]
 
+        ## rebin
+        bins = ((ll-forest.lmin)/forest.dll+0.5).astype(int)
```
londumas: add floor

ngbusca:
That's why I think we should calculate the bins first and the do the `lmin, lmax` cuts....

py/picca/data.py
```diff
@@ -103,6 +113,31 @@ def __init__(self,ll,fl,iv,thid,ra,dec,zqso,plate,mjd,fid,order):
...
+    def __add__(self,d):
+        
+        if not hasattr(self,'ll') or not hasattr(d,'ll'):
+            return self
+
+        ll = sp.append(self.ll,d.ll)
+        fl = sp.append(self.fl,d.fl)
+        iv = sp.append(self.iv,d.iv)
+
+        bins = ((ll-forest.lmin)/forest.dll+0.5).astype(int)
```
londumas: add floor

londumas:
for pix, the difference with master is very small. I am currently looking for spcframe
capture d ecran de 2017-11-16 11-18-38

londumas:
here is with spcframe, looks good:
capture d ecran de 2017-11-16 11-34-43

@ngbusca, ready to be merged

py/picca/data.py
```diff
@@ -63,35 +63,38 @@ class forest(qso):
...
-        w = w & (iv>0)
-        if w.sum()==0:return
-        
-        ll=ll[w]
-        fl=fl[w]
-        iv=iv[w]
+        bins = sp.floor((ll-forest.lmin)/forest.dll+0.5).astype(int)
+        ll = forest.lmin + bins*forest.dll
+        w = (ll>=forest.lmin)
+        w = w & (ll<forest.lmax)
```
ngbusca: here we should make the selection on the bins instead of ll:

    w = (bins >= 0) & (bins < sp.floor((forest.lmax-forest.lmin)/forest.dll+0.5))

londumas: why is the selection on ll not good ?

ngbusca: the difference is on the edge bins. Since forest.lmin is a new bin center, we need to collect all pixels with abs(ll-forest.lmin) < dll/2. If we cut ll > forest.lmin we only get half. The same thing happens near forest.lmax...

londumas: no because I changed the definition of `ll`: `ll = forest.lmin + bins*forest.dll`

ngbusca:
I think this is problematic if you rebin: ll will have repeated values

londumas:
> I think this is problematic if you rebin: `ll` will have repeated values

no because after the cuts, `ll` is defined as `cll = forest.lmin + sp.arange(bins.max()+1)*forest.dll`, `ll = cll[w]`

py/picca/data.py
ngbusca: Ok, I understand. I think it's correct. Did you do a test with rebinning just to double check?

londumas: yes and it works with rebin=5

Issue #241, opened 2017-11-20
add ploting a spectrum


londumas:
given a plate, mjd, fiber, plot a spectrum and its best fit continuum

londumas:
merged

PR #242, opened 2017-11-20
fix issue 241

londumas:
https://github.com/igmhub/picca/issues/241

given a plate, mjd, fiber, plot a spectrum and its best fit continuum


londumas:
For the standard qso we use in auto and cross papers:
spec_6972-56426-0912

```
do_plotSpec.py
--plate 6972
--mjd 56426
--fiberid 912
--drq /uufs/astro.utah.edu/common/home/u6011908/Data/Catalogs/DR14Q_v3_1.fits
--spectrum /uufs/astro.utah.edu/common/home/u6011908/Run_programs/igmhub/picca/DR14/Delta_LYA/Delta/
--in-dir /uufs/astro.utah.edu/common/home/u6011908/Data/Spectra/ForetFusion/Repeats/
--mode pix
--rebin 1
--lambda-rest-min 0.00001
--lambda-rest-max 10000000.
--lambda-max 7235.
--mask-file /uufs/astro.utah.edu/common/home/u6011908/Programs/igmhub/picca/etc/dr14-line-sky-mask.txt
--flux-calib /uufs/astro.utah.edu/common/home/u6011908/Run_programs/igmhub/picca/DR14/Delta_calibration/Log/delta_attributes.fits.gz
```

ngbusca:
@londumas it's great to see developments like this. Could you instead create a "tutorials" folder and put your executable there? It would be even better to have a python notebook to show how to run it...

londumas:
It is now in a tutorials folder.

ngbusca:
@londumas thanks. I'm a bit confused as to why you need so many arguments. Can't you just open the spectrum file (pix, spcframe, etc.) plot the spectrum, then open the delta file and plot the continuum? Why do you need rebin or drq, etc. etc.?

londumas:
The code finds the pix to open from plate, mjd, fiber.
It allows to go faster than reading all files (anyway these arguments are the same as for do_delta)
rebin is needed to open the data file

londumas:
the only and very small problem with my code is that I need to know the nside of the delta-*.fits file, and I don't think we store this info anywhere. I found that it is most of the time 8.

ngbusca:
@londumas I think you could greatly simplify this code if:

* if it's spec or corrected spec:
    * open the file with fitsio, plot the spectrum
    * read RA, DEC, THING_ID from the header => find out the pixel number (unsing healpy) => read the delta file => plot the continuum
* if it's a spCFrame:
    * open with fitsio, plot the spectrum.
     * Read RA, DEC from header [5] => find out the pixel number => read the data file => plot the continuum

* if mode is pix: 
    * this is the only case where you need drq to go faster, you could add a test: 
    if mode == pix and drq == None:
          print "drq required with pix mode"
          sys.exit(1)

This way, the user doesn't need to decide all the other arguments, which are not actually needed to just make a plot... does it make sens to you?

ngbusca:
@londumas yes, we don't store the NSIDE anywhere, it's a problem.

londumas:
> yes, we don't store the NSIDE anywhere, it's a problem.
We should add storing it in the do_delta output files, in the hdu=0
What do you say ? Or we can add a master.txt file that gives the nside.
I prefer the first solution

ngbusca:
you can't add header keywords to hdu0 unless you store an image. Let's add the nside probelm as a separate issue. If you are happy with your code I'll just merge and we will improve/simplify once we figure out where to store nside.

londumas:
ok, let's do that.
Ready to be merged.

Issue #245, opened 2017-11-22
d.neighs in do_<correlation> acts as a memory leak


londumas:
When a correlation is calculated, at each step of the loop, the delta class receives an attribute called `neighs`, with all other objects at a given distance.
After completion of the code, this attribute is still there. This produces a sort of memory leakage.
In the following plot I show the memory usage as a function of calculation step.
In blue for the current calculation, in green if after completion of the loop step I set `d.neights` to None 

![memory_vs_calculation_step](https://user-images.githubusercontent.com/3166436/33150468-d26fef5c-cf90-11e7-96ff-bc8d408d8a0c.png)

The first line is the load of delta, the second the load of QSOs, the third is the xcf_correlation calculation

londumas:
fixed in fix_issue_245

londumas:
merged

PR #246, opened 2017-11-24
fix issue 245

londumas:
Remove useless neighs after computation to avoid memory leak.
https://github.com/igmhub/picca/issues/245

### DIFF
--- a/py/picca/cf.py
+++ b/py/picca/cf.py
@@ -84,6 +84,7 @@ def cf(pix):
                 rt[:len(crp)]+=crt
                 z[:len(crp)]+=cz
                 nb[:len(cnb)]+=cnb
+            setattr(d1,"neighs",None)
 
     w = we>0
     xi[w]/=we[w]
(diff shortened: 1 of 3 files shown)


londumas:
Tested and works perfectly

ngbusca:
@londumas good idea to clean up variables that are not used anymore. Just out of curiosity, what's the difference between e.g. setattr(d,'neighs', None) and d.neighs = None ?

ngbusca:
@londumas by the way, why did you tag this as s bug? There was nothing wrong with the previous code.

londumas:
> @londumas good idea to clean up variables that are not used anymore. Just out of curiosity, what's the difference between e.g. setattr(d,'neighs', None) and d.neighs = None ?

There are no differences, the good thing with `setattr` is that we can get all variable with a dictionnary, as we do for do_xcf since delta are no longer use afet

> @londumas by the way, why did you tag this as s bug? There was nothing wrong with the previous code.

Because it made a memory leak that produces crass for some computation

ngbusca:
@londumas I see your point but note that you can't have memory leaks in python (unlike c++) since the garbage collector will always release memory that is not pointed by anything. Your changes make the code more memory efficient!

londumas:
Yes, it isn't really a memory leak in a C++ way.

PR #250, opened 2017-11-27
Modified py/picca/data.py to correct bugs

jfarr03:
Bugs described in issue number 249.


londumas:
I don't think that hardcoding the `order` is a good idea. Could we request the order in the input or put it in the image header ?
When do we use these image files ?

Issue #252, opened 2017-11-27
removed info of delta in do_metal_xdmat


londumas:
https://github.com/igmhub/picca/blob/e8f1c07ff9e846bf3030af14d860c8f977bab47b/py/picca/xcf.py#L317

in `do_metal_xdmat` we use at multiple times the delta so we should remove only `neighs` and not the other attributes

londumas:
The `Neighs` attributes is calculated only once, so we can't remove it.
@vserret, could you explain to me why do we give a list of metal to the code instead of running the code with different metals every time ?

ngbusca:
@londumas we want to have all the matrices in the same output file no? In any case, this functionality was there in the beginning, I think we want that.

vserret:
This would be better to have just one file at the end. Especially if you want to put it in the fitter.

londumas:
@ngbusca, I never used the file, so I can't tell.
The solution is then to calculate neighs at every step of the loop over metals.

londumas:
Could you do that @vserret ?

ngbusca:
@londumas that's very slow. The best is to not remove the neighs attribute in this function.

ngbusca:
@londumas please remove lines 317, 318 from this file.

ngbusca:
@londumas since you made this change, can you take care of it? Also, please make sure that a similar bug wasn't introduced at other places.

londumas:
> @londumas that's very slow. The best is to not remove the neighs attribute in this function.

If we do that then we put back the issue of filling the memory of the computer.

londumas:
This issue should also be present in do_metal_dmat (ie for cf)

PR #253, opened 2017-11-27
fix issue 252

londumas:
https://github.com/igmhub/picca/issues/252

Fix issue due to missing neighs

### DIFF
--- a/bin/do_metal_dmat.py
+++ b/bin/do_metal_dmat.py
@@ -19,9 +19,9 @@ from multiprocessing import Pool,Process,Lock,Manager,cpu_count,Value
 
 def calc_metal_dmat(abs_igm1,abs_igm2,p):
     if x_correlation: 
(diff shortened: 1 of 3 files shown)


londumas:
@vserret, could you test it?

londumas:
Cool, thank you @vserret. Did you try with two metals?
Did you try the do_metal_dmat (i.e. for cf)?

vserret:
there is a problem on do_metal_dmat.py
problem with identation level on line 23

londumas:
Thank you for spoting this can you test now that I corrected this error ?

vserret:
this works

londumas:
Thank you @vserret.
@ngbusca, ready to be merged.

londumas:
@ngbusca, I agree with you but for the time being I don't have any idea how to do that differently.

PR #254, opened 2017-11-28
avoid pairs of pixels from the same qso for forest xcf

vserret:


ngbusca:
@vserret the `d.ra != d1.ra` condition is not enough?

londumas:
@ngbusca, I asked @vserret to use thingid because it is more stable.
In simulations we can have ra1==ra2 but dec1!=dec2.
Also it make more sens to use thingid than ra

ngbusca:
@londumas I agree it makes more sense, but I was wondering whether this change was a just a matter of principle or there was an actual problem

londumas:
I don't think it introduces a bug now since `ra` is a double not even a float, but in the future it could. Better remove this now than too late.

vserret:
Not a problem, just a matter of principle

Issue #260, opened 2017-11-30
definition of angmax should be after removing delta


londumas:
In do_xcf.py, do_xMetalGrid.py, do_metal_xdmat.py, do_xdmat.py, the definition of angmax should be after removing delta too far from first or last pixel

ngbusca:
do we gain anything? Is it really worth complicating the code for this?

ngbusca:
The probability of introducing a bug is proportional to the number of lines. If we gain 0.1% I prefer not to do it.

londumas:
Then let's define `angmax=2*sp.pi`, this will take one line, it is simple, less chance of bugs, but very long.

ngbusca:
then you lose many many %, not 0.1%

londumas:
fixed in https://github.com/igmhub/picca/pull/261

PR #261, opened 2017-11-30
fix issue 260

londumas:
https://github.com/igmhub/picca/issues/260#issuecomment-348294038

definition of angmax should be after removing delta


londumas:
tested and get exactly the same result for do_xcf

ngbusca:
Should we get exactly the same? I thought we were losing pairs...

londumas:
You are mixing issues, this one is just an optimisation

Issue #257, opened 2017-11-30
Loosing pairs because of definition of angmax with forest-forest cross-correlation


londumas:
https://github.com/igmhub/picca/blob/d2ee8f2f5189f3f265d6d70519f8e364722a6d64/bin/do_cf.py#L187

In the case of cross-correlation:
```
cf.angmax = 2.*sp.arcsin(cf.rt_max/(cosmo.r_comoving(z_min_pix_forest1)+cosmo.r_comoving(z_min_pix_forest2)))
```

londumas:
This issue concerns: do_cf.py, do_dmat.py, do_metal_dmat.py

londumas:
with `cf.angmax = 2.*sp.arcsin(cf.rt_max/(2.*cosmo.r_comoving(z_min_pix)))`, I get for LYA:
`0.0564767370242`
with `cf.angmax = 2.*sp.arcsin(cf.rt_max/(cosmo.r_comoving(z_min_pix)+cosmo.r_comoving(z_min_pix2))`, I get 
`0.063247298498`

londumas:
fixed in https://github.com/igmhub/picca/pull/262

PR #262, opened 2017-11-30
fix issue 257

londumas:
fix https://github.com/igmhub/picca/issues/257
Loosing pairs because of definition of angmax with forest-forest cross-correlation

### DIFF
--- a/bin/do_cf.py
+++ b/bin/do_cf.py
@@ -162,9 +162,10 @@ if __name__ == '__main__':
         if not args.no_project:
             d.project()
 
+    cf.angmax = 2.*sp.arcsin(cf.rt_max/(2.*cosmo.r_comoving(z_min_pix)))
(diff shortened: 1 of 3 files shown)


londumas:
@ngbusca, I don't see any. Where ?

ngbusca:
@londumas you are right, there are no tabs. Before merging, have you checked that this bug is not present in other places?

londumas:
@ngbusca, I corrected in do_cf, do_dmat and do_metal_dmat.
Should I look elswhere for forest1-forest2 correlation ?

ngbusca:
@londumas I was thinking the xcf files

londumas:
no the xcf files are ok.
See https://github.com/igmhub/picca/search?utf8=%E2%9C%93&q=angmax&type=

PR #268, opened 2017-12-03
Coadd obs

ngbusca:
This PR refactors the way spCFrames are read:

* each spCFrame file is open once (so there might be a slight gain in performace)
* only spCFrames participating to an spPlate file are used (I don't know whether all spCFrames can be rejected, but if they are, then they won't be used)
but mainly:
* the user can choose to use the spCFrames participating to the best obs or all (reproducing what ForetFusion does but directly at the spCFrame level and without the intermediate wavelength resampling)

### DIFF
--- a/bin/do_deltas.py
+++ b/bin/do_deltas.py
@@ -252,7 +252,8 @@ if __name__ == '__main__':
         pool = Pool(processes=args.nproc)
         print "iteration: ", it
         nfit = 0
-        data_fit_cont = pool.map(cont_fit, data.values())
+        #data_fit_cont = pool.map(cont_fit, data.values())
+        data_fit_cont = map(cont_fit, data.values())
         for i, p in enumerate(data):
             data[p] = data_fit_cont[i]
 
--- a/bin/do_deltas.py
+++ b/bin/do_deltas.py
@@ -176,8 +176,10 @@ if __name__ == '__main__':
     nit = args.nit
 
     log = open(args.log,'w')
-    data, ndata = io.read_data(args.in_dir,args.drq,args.mode,\
-                              zmin=args.zqso_min,zmax=args.zqso_max,nspec=args.nspec,log=log,keep_bal=args.keep_bal,bi_max = args.bi_max,order=args.order, best_obs = args.best_obs, single_exp = args.single_exp)
+    data, ndata = io.read_data(args.in_dir, args.drq, args.mode,\
+                              zmin=args.zqso_min, zmax=args.zqso_max, nspec=args.nspec, log=log,\
+                              keep_bal=args.keep_bal, bi_max=args.bi_max, order=args.order,\
+                              best_obs=args.best_obs, single_exp=args.single_exp)
    
     ### Get the lines to veto
     usr_mask_obs    = None
--- a/bin/do_deltas.py
+++ b/bin/do_deltas.py
@@ -252,8 +252,7 @@ if __name__ == '__main__':
         pool = Pool(processes=args.nproc)
         print "iteration: ", it
         nfit = 0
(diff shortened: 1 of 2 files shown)


bin/do_deltas.py
```diff
@@ -170,10 +176,9 @@ def cont_fit(data):
     nit = args.nit
 
     log = open(args.log,'w')
-    data,ndata = io.read_data(args.in_dir,args.drq,args.mode,\
-                              zmin=args.zqso_min,zmax=args.zqso_max,nspec=args.nspec,log=log,keep_bal=args.keep_bal,bi_max = args.bi_max,order=args.order)
-    
-
+    data, ndata = io.read_data(args.in_dir,args.drq,args.mode,\
+                              zmin=args.zqso_min,zmax=args.zqso_max,nspec=args.nspec,log=log,keep_bal=args.keep_bal,bi_max = args.bi_max,order=args.order, best_obs = args.best_obs, single_exp = args.single_exp)
```
londumas: Could you split this line on more lines?

py/picca/io.py
```diff
@@ -178,19 +200,12 @@ def read_data(in_dir,drq,mode,zmin = 2.1,zmax = 3.5,nspec=None,log=None,keep_bal
...
-            pix_data = read_from_spcframe(in_dir,thid[w], ra[w], dec[w], zqso[w], plate[w], mjd[w], fid[w], order, mode=mode, log=log)
-            read_time=time.time()-t0
         if not pix_data is None:
             sys.stderr.write("{} read from pix {}, {} {} in {} secs per spectrum\n".format(len(pix_data),pix,i,len(upix),read_time/(len(pix_data)+1e-3)))
         if not pix_data is None and len(pix_data)>0:
             data[pix] = pix_data
             ndata += len(pix_data)
 
-        if not nspec is None:
-            if ndata > nspec:break
```
londumas: why do you remove this?

py/picca/io.py
```diff
@@ -262,51 +277,111 @@ def read_from_pix(in_dir,pix,thid,ra,dec,zqso,plate,mjd,fid,order,log=None):
...
 
-                index = fib_list.index(f)
+            flux = spcframe[0].read()
+            ivar = spcframe[1].read()*(spcframe[2].read()==0)
+            llam = spcframe[3].read()
+            
+            ## now convert all those fluxes into forest objects
+            for index, (t, r, d, z, p, m, f) in enumerate(zip(thid[wfib], ra[wfib], dec[wfib], zqso[wfib], plate[wfib], mjd[wfib], fid[wfib])):
+                index =(f-1)%500
+                wlam = (10**llam[index]/(1+z) > 1220) & (10**llam[index]/(1+z)<1230)
```
londumas: why 1220 and 1230? and I don't see where you use this.

py/picca/io.py
ngbusca: this is now done between l. 116 and l.124:

    if nspec != None:
        ra = ra[:nspec]
        dec = dec[:nspec]
        zqso = zqso[:nspec]
        thid = thid[:nspec]
        plate = plate[:nspec]
        mjd = mjd[:nspec]
        fid = fid[:nspec]

bin/do_deltas.py
ngbusca: sure

py/picca/io.py
ngbusca: woops, leftover from debugging...

bin/do_deltas.py
ngbusca: done

py/picca/io.py
ngbusca: removed

PR #269, opened 2017-12-05
Fix issue 258

vserret:


ngbusca:
what bothers me from this is that the fill_neighs is the coupled to the fiducial cosmology. I'd prefer to keep a pure angle selection even if this results in some lose of performance. I think we should even remove this cut from the xcf.fill_neighs function.

londumas:
@vserret, what is your gain in terms of time?
@ngbusca, why don't you want to couple fill_neighs to the cosmology ? Any way we already suppose flatness.

ngbusca:
@londumas because we might want to do the analysis in observable coordinates where there's no fiducial cosmology.

londumas:
@ngbusca, if we gain a lot from this commit, we could have different version of fill_neight. If we don't gain enough, no need to do anything.

ngbusca:
@londumas yep, as with any optimization!

londumas:
>@londumas yep, as with any optimization!

Agree

londumas:
@vserret, can we remove this PR?

vserret:
yes

PR #273, opened 2017-12-06
convert lambda-abs argument in string not float 

vserret:


mblomqvist:
Please add CIV(1549) 1549.06 to constants.py if you're going ahead with this modification.

ngbusca:
@mblomqvist can we call it CIV(eff) instead? So that we make it clear that it's an effective line and not an actual transition...

mblomqvist:
OK, I guess that's reasonable.

londumas:
Thanks @vserret  to take care of this
`--lambda-abs2` should be `None` by default I think

vserret:
I don't agree, I think this is better not to have to write "--lambda-abs2" in all the cases.

londumas:
@vserret, the problem is that currently to o the auto-correlation on the SiIV forest I need to set `lambda-abs2` because of the line `elif args.lambda_abs != args.lambda_abs2:`.
When doing an auto-correlation, I should have to worry about this. Don't you agree?

ngbusca:
I don't know when it happened but the current do_metal_dmat is broken for the autocorrelation. Guys, you *must* be extra careful when doing changes and make sure to run thorough tests.

londumas:
@ngbusca, what is the bug?

ngbusca:
@londumas I patched it now in master. When alpha2 was introduced, the indentation was done with a tab which translated into a wrong indentation on my system. When lambda-abs2 got defaulted to None instead of LYA, the code thought that I wanted to do a cross-correlation since `args.lambda_abs != args.lambda_abs2`.

vserret:
This is strange because I run the code before pulling your commits and it was working.

ngbusca:
@vserret it's also my fault since I didn't to test runs before approving. 

If you look at here:

https://github.com/igmhub/picca/commit/f25a0fafe341718ac077224020e4dee27b04a042#diff-949bb7e0439bbe951b2904b194bad6ad

You can see the indentation problems in do_cf l. 175 and do_metal_dmat, l. 176. The problem with the tab is that what works in one system might not work in another, which is why we need to be extra careful.

The other problem is still in master actually:

        elif args.lambda_abs != args.lambda_abs2:
             x_correlation=True

since `args.lambda_abs2` is now `None`, `x_correlation = True` when args.lambda_abs2 is not specified.

vserret:
I don't see that change
Do you want me to implement it and change the if conditions or should we let lambda-abs2 ='LYA'  by default ?

londumas:
@vserret. We should have lambda2=none by default, then test it with 'if not lambda2 is None'. When doing an autocorrelation, we shouldnt have to scecify lambda2.

bin/do_cf.py
```diff
@@ -101,6 +101,10 @@ def corr_func(p):
 
     cosmo = constants.cosmo(args.fid_Om)
 
+    lambda_abs  = constants.absorber_IGM[args.lambda_abs]
+    if (args.lambda_abs2) : lambda_abs2 = constants.absorber_IGM[args.lambda_abs2]
+    else: lambda_abs2 = constants.absorber_IGM[args.lambda_abs]
```
londumas: please remove this line. if no input it should be None

bin/do_cf.py
```diff
@@ -101,6 +101,10 @@ def corr_func(p):
 
     cosmo = constants.cosmo(args.fid_Om)
 
+    lambda_abs  = constants.absorber_IGM[args.lambda_abs]
+    if (args.lambda_abs2) : lambda_abs2 = constants.absorber_IGM[args.lambda_abs2]
```
londumas: `if not args.lambda_abs2 is None`

bin/do_cf.py
```diff
@@ -145,14 +149,14 @@ def corr_func(p):
                     if ndata2>args.nspec:break
         else:
             dels2 = delta.from_image(args.in_dir2)
-    elif args.lambda_abs != args.lambda_abs2:   
+    elif lambda_abs != lambda_abs2:
```
londumas: replace by `if not  args.lambda_abs is None and lambda_abs != lambda_abs2`

bin/do_dmat.py
```diff
@@ -106,7 +106,9 @@ def calc_dmat(p):
 
     cosmo = constants.cosmo(args.fid_Om)
 
-
+    lambda_abs = constants.absorber_IGM[args.lambda_abs]
+    if args.lambda_abs2: lambda_abs2 = constants.absorber_IGM[args.lambda_abs2]
+    else: lambda_abs2 = constants.absorber_IGM[args.lambda_abs]
```
londumas: same here

londumas: and all that follows

bin/do_cf.py
vserret: I have seen that you put this in other places in the code, why ? 
This is shorter to write "if (args.lambda_abs2)", there is a difference with "if not args.lambda_abs2 is None" ?

bin/do_cf.py
vserret: If you want to compute the cross-correlation for Lya in lya and lya in lyb. You need to have something for lambda-abs2. 
If I remove that, I will have to write "--lambda-abs2" everytime I want to compute a xcf.

bin/do_cf.py
londumas: no difference, do as you want

bin/do_cf.py
londumas: > If I remove that, I will have to write "--lambda-abs2" everytime I want to compute a xcf.

It is logical to have to write `--lambda-abs2` everytime you compute a cross-forest-forest correlation
It is not logical to write `--lambda-abs2` when you compute an auto-forest correlation.

vserret: I agree but here you don't write it for the auto-correlation

londumas: Just test doing the auto-correlation of SiIV or CIV, you will see the problem

vserret: I've tested it 
command : "do_cf.py --in-dir deltas-civ --lambda-abs "CIV(1551)" --out auto_civ.fits "
I don't understand your problem. Did you use the last version of this branch ?

londumas: perfect, if it works and if it is not flagged as cross as before, it is perfect.

londumas:
I just looked ot the commits, it seems good @ngbusca. However, I didn't try tu run all possible code.
@vserret, did you try running all the `do_` ?
Thank you @vserret.

vserret:
Yes I've try all of them 
But It would be better to check again to avoid identation problems

londumas:
@vserret, for indentation just search for tab with "ctrl+f"

londumas:
@vserret, could you remove the space and tell me how you did to find `'CII(1035)'   : 1334.53,` ?

londumas:
@vserret, is it ready to be merged?

vserret:
Yes

londumas:
@ngbusca, @vserret PR is ready to be merged.

londumas:
@vserret, very nice! Do this PR passes the unitest?

londumas:
@ngbusca @vserret just told me it passes the tests. Ready to be merged.

PR #275, opened 2017-12-08
Read desi files

londumas:
@ngbusca, could you review this commit?
Please don't merge yet. We are testing the full pipeline.

### DIFF
--- a/py/picca/io.py
+++ b/py/picca/io.py
@@ -105,94 +105,6 @@ def read_drq(drq,zmin,zmax,keep_bal,bi_max=None):
 
     return ra,dec,zqso,thid,plate,mjd,fid
 
-def read_truth(drq,zmin,zmax,keep_bal,bi_max=None,spectype="QSO"):
-    '''
-        read desi truth catalog
-
-    '''
-    vac = fitsio.FITS(drq)
-    vacTargets = fitsio.FITS(drq.replace("truth.fits","targets.fits"))
-
-    ## Info of the primary observation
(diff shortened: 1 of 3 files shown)


londumas:
works very well. It only requires to build by hand a fake drq from the desi pipeline.

py/picca/io.py
```diff
@@ -156,17 +155,11 @@ def read_data(in_dir,drq,mode,zmin = 2.1,zmax = 3.5,nspec=None,log=None,keep_bal
...
-        typ = h[1]["TRUESPECTYPE"][:]
-        w = ["QSO" in t for t in typ]
-        w = sp.array(w)
-        ztable = {t:z for t,z in zip(tid[w],zs[w]) if z > zmin and z<zmax}
-        sys.stderr.write("Found {} qsos\n".format(len(ztable)))
-        return read_from_desi(nside,ztable,in_dir)
+        sys.stderr.write("Found {} qsos\n".format(len(zqso)))
+        return read_from_desi(nside,in_dir,thid,ra,dec,zqso,plate,mjd,fid,order)
+
+        return read_from_desi(nside,ztable,in_dir,order)
```
ngbusca: it looks like this line is never reached!

londumas: Could you show me an exemple of code that reproduces your bug ? It witks very well fir me

ngbusca: it's just that l. 162 is never reached since there's a return statement in l. 160, no?

londumas: indeed, sorry. Just corrected it.

Issue #220, opened 2017-11-06
replace "print"  by sys.stderr.write or sys.stdout.write or print()


londumas:
print has changed in Python3, should be replaced to either sys.stderr.write or sys.stdout.write or print()

mblomqvist:
Can you elaborate on the problem? Python 2 is forward-compatible with the print parenthesis used in Python 3. Error reporting should be done with sys.stderr.write.

londumas:
there are print without parenthesis, those are not forward-compatible

mblomqvist:
I know, so that could be added instead of moving away from using print (when appropriate)

londumas:
@mblomqvist, you are right.

ngbusca:
I'm not too worried about compatibility with python3 (for now), but we could one day systematically do 
`
from __future__ import print_function
`
and use parenthesis. This is very very very *very* low priority

londumas:
fixed in PR https://github.com/igmhub/picca/pull/278

londumas:
fixed in PR https://github.com/igmhub/picca/pull/279

PR #278, opened 2017-12-14
replace print by print()

londumas:
https://github.com/igmhub/picca/issues/220


londumas:
@ngbusca, tested and works

PR #279, opened 2017-12-19
Convert to python3

londumas:
Works both in Python 2 and 3.
Needs an important and serious review


londumas:
Also, currently all the `do_metal_*dmat.py` don't work in Python3 because of this issue:https://github.com/esheldon/fitsio/issues/149

ngbusca:
@londumas if sorting fixes the problem, I would go with that

londumas:
Gives the same results with Python2, working on Python3.

londumas:
@ngbusca, the current PR works very well in Python2.
In Python3 it fails for cf1d,  cf1d_cross, all dmat and all metal_dmat.
It seems to be linked to different ways to round numbers between Python2 and Python3.
Have you heard of such a thing? It also could be something else.
For example the line 189 of do_cf1d.py: ` cor[w]/=norm[w]` gives slightly different numbers in Python2 and in Python3.

ngbusca:
@londumas can you try using the numpy.allclose function instead of exact equality? If this solves the problem, we might as well go with that. numpy.allclose is quite a strong test of equality, testing both absolute and relative differences.

londumas:
@ngbusca, works very well. Tested on Python2 and Python3.

PR #281, opened 2018-01-05
add 3d forest-correlation (l1/l2,ang)

londumas:
This PR adds the 3d forest-correlation (l1/l2,ang).
It is very useful to understand metals and other systematics.
Obviously @ngbusca, don't hesitate if you don't like some names of variables and or files or any other features.


londumas:
Here is the result for the lya auto correlation on DR14
lya_lya_correlation

ngbusca:
@londumas this is great! Do you think you could add a third axis with the mean redshift of pairs?

londumas:
@ngbusca, it is already included since it copies the code of do_cf.

ngbusca:
@londumas I meant a real third axis with bins in pair redshift. We don't do that in do_cf

londumas:
do you mean allow for different redshift bins ? I could look at that. Then the arrays would be 3d instead of 2d.

londumas:
@ngbusca, what about doing this PR for the time being and leave redshift bins for another time?

ngbusca:
Thanks for implementing this, it's a great feature to have. Let's see if we can make some progress on #284 before merging.

londumas:
@ngbusca, ready to be merged. Do you need me to do something else?

ngbusca:
@londumas did you run the tests?

londumas:
Just did and it works. Added a test for do_cf_angl

Issue #286, opened 2018-01-16
Add tests


londumas:
In the light of many PR, we want standard tests to be run before merging.
The idea is to create fake data with given seed, maybe 10,000 forest would be enough.
Create a random DRQ, then random pixel.

londumas:
Here is a code to generate random catalog and forest. What do you think of this ?

```
#
# Produce fake data for a full Lya-Lya and QSO-Lya study
#   Copy the range of PLATE, MJD, ... from SDSS DR14Q
#

import scipy as sp
import numpy.random
from scipy.interpolate import interp1d
import fitsio
import healpy

def produce_cat(path_DRQ,nObj):

    print("Create cat with number of object = ", nObj)

    ### Create random catalog
    ra    = (360.-0.)*numpy.random.random_sample(nObj) + 0.
    dec   = (90.+20.)*numpy.random.random_sample(nObj) - 20.
    plate = numpy.random.randint(266,   high=10001, size=nObj )
    mjd   = numpy.random.randint(51608, high=57521, size=nObj )
    fid   = numpy.random.randint(1,     high=1001,  size=nObj )
    thid  = sp.arange(1,nObj+1)
    zqso  = (5.8-1.7)*numpy.random.random_sample(nObj) + 1.7
   
    ### Save
    out = fitsio.FITS(path_DRQ,'rw',clobber=True)
    cols=[ra,dec,thid,plate,mjd,fid,zqso]
    names=['RA','DEC','THING_ID','PLATE','MJD','FIBERID','Z']
    out.write(cols,names=names)
    out.close()

    return
def produce_forest_in_pixel(path_PIX,path_DRQ,nside):

    ### Load DRQ
    vac = fitsio.FITS(path_DRQ)
    ra    = vac[1]["RA"][:]*sp.pi/180.
    dec   = vac[1]["DEC"][:]*sp.pi/180.
    thid  = vac[1]["THING_ID"][:]
    plate = vac[1]["PLATE"][:]
    mjd   = vac[1]["MJD"][:]
    fid   = vac[1]["FIBERID"][:]
    vac.close()

    ### Get Healpy pixels
    pixs  = healpy.ang2pix(nside, sp.pi/2.-dec, ra)

    ### Save master file
    path = path_PIX+"/master.fits"
    head = {}
    head['NSIDE'] = nside
    cols  = [thid,pixs,plate,mjd,fid]
    names = ['THING_ID','PIX','PLATE','MJD','FIBER']
[...]

PR #288, opened 2018-01-19
first commit for unittest

londumas:
This is an ongoing work in creating unittest for picca.
I still lack some knowledge on how to make it run on github.
Linked to issue https://github.com/igmhub/picca/issues/286

@ngbusca. For the time being, I create 1000 Qsos/forests. The correlations are calculated on the default grid (rmax=200). This produces big firs.gz file. I could change this by calculating only for rmax=20 or some other value. What do you think?


ngbusca:
@londumas thanks for all this work, it's pretty impressive. I've run it and it passed the tests ;^)
I've introduced a mistake in do_cf.py that was caught by the code, excellent!

Is it ready to merge?

londumas:
@ngbusca, it is.

PR #292, opened 2018-01-26
Add obj name

londumas:
Add tests for metal grid and add object name to do_metal_xdmat fits file entry


ngbusca:
is there any other file that needs this change, other than do_metal_xdmat.py ?

londumas:
@ngbusca, I don't know. Maybe do_metal_dmat.py. Depends on how fitter2 read the data.
By the way I run the tests and it is perfect

PR #294, opened 2018-01-31
run test on 1 proc instead on 4

londumas:


ngbusca:
incidentally, it doesn't fix the tests on my mac. Still I'll merge since it's better to run the tests on a single thread on small computers. I've restored the defaults on allclose.

Issue #284, opened 2018-01-09
excess correlation for lambda1==lambda2


londumas:
The physical correlation between two pixels of the same forests or of different forests has a contribution from a non-physical correlation.
This non-physical correlation could be related to the sky. It still needs to be understood.
It is a function of theta: something like xi = theta^parameter

Here is the x(l1/l2,theta) correlation of the MgII_1 and MgII_2 forests. You can see the strong line at l1/l2==1.

![capture d ecran de 2018-01-05 22-43-50](https://user-images.githubusercontent.com/3166436/34701166-de9f7b20-f4a3-11e7-8b69-3ab138d36af7.png)

Here is the comparison between the Lya auto-correlation and the MgII_2 auto-correlation.

![capture d ecran de 2018-01-08 18-43-12](https://user-images.githubusercontent.com/3166436/34701167-dec03fe0-f4a3-11e7-82d6-c45ab7912860.png)

in the code, in def fast_cf(z1,r1,w1,d1,z2,r2,w2,d2,ang,same_half_plate):

The safe and conservative way is to remove all these pairs:
```
    if ang_correlation:
        w = (rp==1.)
        wd12[w] = 0.
        w12[w]  = 0.
    else:
        w = (rp==0.)
        wd12[w] = 0.
        w12[w]  = 0.
```

The more realistic way is to let the user decide, since for Lya it is small and could be estimated redward, or maybe moddeled.
```
    if remove_same_wave:
        if ang_correlation:
            w = (rp==1.)
            wd12[w] = 0.
            w12[w]  = 0.
        else:
            w = (rp==0.)
            wd12[w] = 0.
            w12[w]  = 0.
```

ngbusca:
@londumas are you using coadds of multiple observations or the best observation?

londumas:
@ngbusca, I use `/uufs/astro.utah.edu/common/home/u6011908/Data/Spectra/ForetFusion/Repeats/.

londumas:
@ngbusca, I have noticed no difference when using spcframe.
A good test I could do would be to use skyfibers with a random  redshift for the fake "QSO". It would help see if it is the sky or something else. What do you think?

ngbusca:
@londumas the spcframe option still coadds all the available observations. You want to make sure that you use only one plate per spectrum.

ngbusca:
@londumas actually one plate-mjd combination per spectrum

londumas:
[...]

PR #296, opened 2018-01-31
add option to reject pairs with same wavelength

londumas:
add option to reject pairs with same wavelength.
fix issue https://github.com/igmhub/picca/issues/284


londumas:
@ngbusca, yes it passes the tests

PR #299, opened 2018-01-31
Add xcf angl

londumas:
add do_xcf_angl.py to compute xcf(theta,l1/l2).
Works very well and tested with unitest.
Add also a unitest for this correlation

lyaxqso


londumas:
@ngbusca, This PR does multiple things:
    - add do_xcf_angl
    - add project by default when reading deltas
    - correct possible bug because of using `w` for cutting some pairs in cf.py
    - correct counting number of pairs in cf.py and xcf.py when weight of the pair is zero
    - remove useless `lambda-abs-name` in do_metal_xdmat.py
This PR passes the test with the new version of `cf_angl.fits.gz`, `cf.fits.gz` and `cf_cross.fits.gz`. I had to remake them since they were not no longer correct.

ngbusca:
@londumas thanks for implementing this, it's a nice feature to have.

There are explicit calls to delta.project in some `bin/do_...` files. Do you confirm that we don't project twice after your changes? At some poing we need to move to using read_deltas everywhere, but it's a TODO.

londumas:
@ngbusca; yes I confirm it is not applied twice.

PR #307, opened 2018-02-02
Cut zqso

vserret:
do a cut in zqso pairs during the computation of the cf


vserret:
@ngbusca : the test is ok. This can be merged

londumas:
@ngbusca, ready to be merged. @vserret ran the tests and it works

mblomqvist:
This pull request was never merged. Why was it closed?

ngbusca:
@mblomqvist after discussion (with you!) we realized that cutting on zqso_pair puts the lyb forest in the high-z bin while it's mostly low-z bins. @vserret is changing that to a cut on the redshift of the last forest bin which will fix it. It will be implemented for the cross as well...

vserret:
@mblomqvist as @ngbusca said I'm changing the zcut. I will put the new version in few hours on git

mblomqvist:
OK, very good.

PR #310, opened 2018-02-09
Initial commit of read_spplate

ngbusca:
I've cleaned up some leftover code in read_spcframes and added
the functionality to directly read spplates. The user can choose
to read all available spPlates or only the best observation.


py/picca/io.py
```diff
@@ -423,6 +430,54 @@ def read_from_spcframe(in_dir, thid, ra, dec, zqso, plate, mjd, fid, order, mode
...
+def read_from_spplate(in_dir, thid, ra, dec, zqso, plate, mjd, fid, order, log=None, best_obs=False):
+    pix_data={}
+    unique_plates = sp.unique(plate)
+    print "reading {} plates".format(len(unique_plates))
+
+    for p in unique_plates:
+        wplate = plate==p
+        plate_mjd = "{}-*".format(p)
+
+        spplates = glob.glob(in_dir+"/{}/spPlate-{}.fits".format(p, plate_mjd))
```
londumas: We need to compare the list `spplates` to the list of mjd in `sp.unique(mjd[wplate])` in order to find missing plates

ngbusca:
@londumas I've added an INFO message to the screen and the log when mjds are missing

PR #311, opened 2018-02-13
add storing nside

londumas:
add the input `nside` to correlation header.


londumas:
@ngbusca, run testes and works

PR #315, opened 2018-02-14
Add unitest for Pk1D

Cyeche:


py/picca/test/test_cor.py
```diff
@@ -63,7 +66,7 @@ def produce_folder(self):
         """
 
         print("\n")
-        lst_fold = ["/Products/","/Products/Spectra/",
+        lst_fold = ["/Products/","/Products/Spectra/","/Products/Pk1D/",
         "/Products/Delta_LYA/","/Products/Delta_LYA/Delta/",
         "/Products/Delta_LYA/Log/","/Products/Correlations/",
```
londumas: I would add instead
````
"/Products/Delta_Pk1D/","/Products/Delta_Pk1D/Delta/",
         "/Products/Delta_Pk1D/Log/"
```

py/picca/test/test_cor.py
```diff
@@ -230,6 +233,56 @@ def send_delta(self):
...
+
+
+    def send_delta_Pk1D(self):
+
+        print("\n")
+        ### Send
+        cmd  = " do_deltas.py"
+        cmd += " --in-dir "          + self._branchFiles+"/Products/Spectra/"
+        cmd += " --drq "             + self._branchFiles+"/Products/cat.fits"
+        cmd += " --out-dir "         + self._branchFiles+"/Products/Delta_LYA/Delta/"
```
londumas: `self._branchFiles+"/Products/Delta_Pk1D/Delta/"`

py/picca/test/test_cor.py
```diff
@@ -230,6 +233,56 @@ def send_delta(self):
...
+#        cmd += " --out-dir "         + "./"
```
londumas: remove comment

py/picca/test/test_cor.py
```diff
@@ -230,6 +233,56 @@ def send_delta(self):
...
+        cmd += " --iter-out-prefix " + self._branchFiles+"/Products/Delta_LYA/Log/delta_attributes"
```
londumas: `"/Products/Delta_Pk1D/Log/delta_attributes_Pk1D"`

py/picca/test/test_cor.py
```diff
@@ -230,6 +233,56 @@ def send_delta(self):
...
+#        cmd += " --iter-out-prefix " + "./delta_attributes"
```
londumas: remove comment

py/picca/test/test_cor.py
```diff
@@ -230,6 +233,56 @@ def send_delta(self):
...
+        cmd += " --log "             + self._branchFiles+"/Products/Delta_LYA/Log/input.log"
```
londumas: `"/Products/Delta_Pk1D/Log/input.log"`

py/picca/test/test_cor.py
```diff
@@ -230,6 +233,56 @@ def send_delta(self):
...
+        cmd += " --delta-format Pk1D --order 0 --use-constant-weight" 
+        cmd += " --lambda-min 3650. --lambda-max 7200.0 --lambda-rest-min 1050.0 --lambda-rest-max 1180" 
+        cmd += " --nproc 1"
+        subprocess.call(cmd, shell=True)
+
+        ### Test
+        if self._test:
+            path1 = self._masterFiles + "/delta_attributes_Pk1D.fits.gz"
+            path2 = self._branchFiles + "/Products/Delta_LYA/Log/delta_attributes.fits.gz"
```
londumas: "/Products/Delta_Pk1D/Log/delta_attributes_Pk1D.fits.gz"

py/picca/test/test_cor.py
```diff
@@ -230,6 +233,56 @@ def send_delta(self):
...
+            self.compare_fits(path1,path2,"do_deltas.py")
```
londumas: `self.compare_fits(path1,path2,"do_deltas.py for Pk1D")`

py/picca/test/test_cor.py
```diff
@@ -230,6 +233,56 @@ def send_delta(self):
...
+
+            path1 = self._masterFiles + "/delta_Pk1D.fits.gz"
+            path2 = self._branchFiles + "/Products/Delta_LYA/Delta/delta-337.fits.gz"
```
londumas: `"/Products/Delta_Pk1D/Delta/delta-337.fits.gz"`

py/picca/test/test_cor.py
```diff
@@ -230,6 +233,56 @@ def send_delta(self):
...
+        cmd += " --nproc 1"
+        subprocess.call(cmd, shell=True)
+
+        ### Test
+        if self._test:
+            path1 = self._masterFiles + "/delta_attributes_Pk1D.fits.gz"
+            path2 = self._branchFiles + "/Products/Delta_LYA/Log/delta_attributes.fits.gz"
+            self.compare_fits(path1,path2,"do_deltas.py")
+
+            path1 = self._masterFiles + "/delta_Pk1D.fits.gz"
```
londumas: `"/delta-337_Pk1D.fits.gz"`

py/picca/test/test_cor.py
```diff
@@ -230,6 +233,56 @@ def send_delta(self):
...
+            path2 = self._branchFiles + "/Products/Delta_LYA/Delta/delta-337.fits.gz"
+
+            self.compare_fits(path1,path2,"do_deltas.py")
```
londumas: `self.compare_fits(path1,path2,"do_deltas.py for Pk1D")`

py/picca/test/test_cor.py
```diff
@@ -230,6 +233,56 @@ def send_delta(self):
...
+
+        return
+
+    def send_Pk1D(self):
+
+        print("\n")
+        ### Send
+        cmd  = " do_Pk1D.py"  
+        cmd += " --in-dir "          + self._masterFiles + "/delta_Pk1D/"
+        cmd += " --out-dir "         + self._branchFiles+"/Products/Pk1D/"
```
londumas: `self._branchFiles+"/Products/Correlations/"

py/picca/test/test_cor.py
```diff
@@ -230,6 +233,56 @@ def send_delta(self):
...
+            self.compare_fits(path1,path2,"do_deltas.py")
+
+        return
+
+    def send_Pk1D(self):
+
+        print("\n")
+        ### Send
+        cmd  = " do_Pk1D.py"  
+        cmd += " --in-dir "          + self._masterFiles + "/delta_Pk1D/"
```
londumas: `self._branchFiles + "/delta_Pk1D/"`

py/picca/test/test_cor.py
```diff
@@ -230,6 +233,56 @@ def send_delta(self):
...
+        cmd += " --out-dir "         + self._branchFiles+"/Products/Pk1D/"
+        subprocess.call(cmd, shell=True)
+
+        ### Test
+        if self._test:
+            path1 = self._masterFiles + "/Pk1D.fits.gz"
+            path2 = self._branchFiles + "/Products/Pk1D/Pk1D-0.fits.gz"
```
londumas: `self._branchFiles + "/Products/Correlations/Pk1D-0.fits.gz"`

py/picca/test/test_cor.py
```diff
@@ -230,6 +233,56 @@ def send_delta(self):
...
+            self.compare_fits(path1,path2,"do_deltas.py")
```
londumas: ` self.compare_fits(path1,path2,"do_Pk1D.py")`

ngbusca:
@Cyeche your local master is different from the remote master. This makes it hard to compare your new code... This happened because you committed modifications to your local master before creating the Pk1D branch. To restore your local master to the remove please do:

    git checkout master
    git fetch origin
    git reset --hard origin/master

After that, go back to your Pk1D branch `git checkout Pk1D` and:

 * remove the useless commnents
 * add `from __future__ import print_function` as the first import and change your prints from a satement(i.e. without the parenthesis) to a function: `print("toto")` so that it works both on python2 and python3
 * commit and push

Thanks!!

Cyeche:
Should be fine now.... Except if you push faster in the master branch than I wrote

londumas:
@Cyeche, I asked for two more changes.
One is esthetic, the other is important.
Unitest should run on branch files. not on master files

ngbusca:
@Cyeche it's not a problem of your master being behind, it's that your master diverged from picca's master, because you committed changes to your master branch. We need to fix that before I can merge. Did you follow the steps from my previous comment?

Cyeche:
Actually I didn’t because I haven’t understood the problem… I restarted from scratch and I think that I haven’t fixed the problem...

ngbusca:
@Cyeche can you do a dummy commit and push?

ngbusca:
@Cyeche I can fix it from my end, but you will need to erase your local repository and clone again. Is that ok?

Cyeche:
Ok, but basically it was I did. I have already  started from a new clone.

Cyeche:
The point is that with the « spec » file produced by the unitest,  the resolution information, wdisp, seems to be different from the real data…. I will investigate more on this point 
> 
> @Cyeche <https://github.com/cyeche>, I asked for two more changes.
> One is esthetic, the other is important.
> Unitest should run on branch files. not on master files

londumas:
@Cyeche, maybe you have to reproduce your `self._masterFiles + "/Pk1D.fits.gz"` because before you could have been using Delta_lya files.

Cyeche:
Hi,

It would be very useful if someone may produce in test_control, forests with the ‘spec' format. Because, for the pk1D in test_control, we are currently testing against the ‘pix' format spectra and this is not what we are doing with ‘true’ data. 

To reproduce, our 1D analysis , in principle, we need both wdisp and several exposures for each objects. However, a minimal change would be to add just wdisp array in the ‘pix’ format.

Cheers, Christophe.

londumas:
@Cyeche, feel free to add a `produce_forests_spec(self)` to produce Monte Carlo in such a format.
Maybe we should wrap-up this PR before we do such a thing.

Cyeche:
BTW, what is heretic if I use true spectra instead of ‘simulated’ spectra?

londumas:
@Cyeche, why not but you would have to download them at each test because we can't store them on git. You have to be sure it wouldn't be too long. Also we can not store passwords on git.

Cyeche:
Actually, I just need one spectra, it is not so big.

londumas:
@Cyeche, you need more than one spectra because you need to test the estimation of the error.
And this can't be done with only one. Can it be?
I think the requirement is that no bins of the measure are empty and that there is a minimum of two entries per bin to get the variance.

Cyeche:
For the 1D spectrum, I really need a single ‘spec’ spectrum with a few exposures, the typical size is just 1.5M. I think that we can store it in git.  
Tell me, if you think that it is reasonable I can write it very fast.

londumas:
Can you test how long it is to download it from the public data release. Maybe just `time wget <spec>`

Cyeche:
Public release you mean from DR12 for instance?

londumas:
@Cyeche, yes.

ngbusca:
I'm ok with using a true spec file. You should just add it to git, it should be ok.

londumas:
@ngbusca, it would be better to download it. Let's see how long it makes.
If it is too long we add it to github instead.

ngbusca:
@londumas why? We already have binary files in the test directory, they take up 11MB, so adding another 1.5 should be ok.

Cyeche:
I you have the command line we can test it.

londumas:
@ngbusca, as you prefer.

ngbusca:
if you put it in git then I don't need to change anything in travis. If you download it it might cause problems (e.g. wget doesn't come by default in mac, we would have to make it a requirement, etc.) Let's just add it to git.

ngbusca:
@londumas @Cyeche has there been any evolution on this discussion?

Cyeche:
I did some tests and the problem is that we need a minimum number (~10) of spectra  to get a result that makes sense for the continuum and the mean flux. The simulation of the Blue and Red raw spectra is quite complicated and the best way is still to use real data. I’m considering the possibility to use real spectra reduced to the width of the lya-a forest to keep a reasonable storage size. Any other ideas are welcome!

Cyeche:
I added a test of do_deltas based ‘spec’ format spectra. Only the interesting regions of the spectra (~around Ly-a forest) are stored in the test directory.

PR #317, opened 2018-02-14
Protect i/o for pk1D

Cyeche:


londumas:
@Cyeche, did you run the tests?

mblomqvist:
The code mixes False and None. Should we make it uniform?

Cyeche:
Yes and it passed the unitest.

Cyeche:
> 
> The code mixes False and None. Should we make it uniform?

The default is None  in do_deltas and I suspect that the test is done over None in many places….

mblomqvist:
Disregard the earlier comment which was immediately deleted.

Issue #304, opened 2018-02-01
Add calculating the cross-covariance


londumas:
The idea is to add a script like `export.py` to get the cross-covariance from two data sets.

londumas:
It would be nice to add `nside` to the header of correlation output.
With this we can compare the nside of `data1` and `data2` when computing the cross-covariance

londumas:
Add `nside` to output correlation in `add_nside_in_cor`

PR #318, opened 2018-02-14
compare nside of both correlations

londumas:
This PR adds a comparison between the nside of the two correlations.
If the nside are different, the code can not compute a cross-covariance.
Fixes https://github.com/igmhub/picca/issues/304


londumas:
@ngbusca, passes the tests. Ready to be merged.

ngbusca:
@londumas it should compare both NSIDE and the pixelization scheme (NEST or RING), no ?

PR #320, opened 2018-02-14
add option minos and storing minimum attributes

londumas:
This PR adds in fitter2:
    - the storage of the `self.best_fit.get_fmin()` attributes. i.e. `is_valid`, `fval`, `has_pos_def_covar`, ...
    - an option to run minos on given parameters. As in fitter1, `_all_` runs on all
    This is done through adding the following lines in the file given to fitter2:
```
[minos]
sigma = 1.0 <or> 2.
parameters = _all_ <or> param1 param2
```

This PR aims at getting the same results we stored in fitter1.
The next step is running a chi2 scan.
@ngbusca, Don't hesitate to ask me to change anything

### DIFF
--- a/py/picca/fitter2/chi2.py
+++ b/py/picca/fitter2/chi2.py
@@ -109,7 +109,7 @@ class chi2:
         if not hasattr(self,"minos_para"): return
 
         sigma = self.minos_para['sigma']
-        if '_all_' in self.minos_para['parameters']:
+        if 'all' in self.minos_para['parameters']:
             self.best_fit.minos(var=None,sigma=sigma)
         else:
             for var in self.minos_para['parameters']:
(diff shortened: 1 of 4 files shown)


py/picca/fitter2/chi2.py
```diff
@@ -98,6 +105,22 @@ def fastMC(self):
                     self.fast_mc[p] = []
                 self.fast_mc[p].append([v, best_fit.errors[p]])
 
+    def minos(self):
+        if not hasattr(self,"minos_para"): return
+
+        sigma = self.minos_para['sigma']
+        if '_all_' in self.minos_para['parameters']:
```
ngbusca: why do you want `_all_` instead of `all` ?

py/picca/fitter2/chi2.py
```diff
@@ -127,7 +150,13 @@ def export(self):
             g.attrs['chi2'] = d.chi2(self.k, self.pk_lin, self.pksb_lin, self.best_fit.values)
             fit = g.create_dataset("fit", d.da.shape, dtype = "f")
             fit[...] = d.best_fit_model
-    
+
+        ## write down all attributes of the minimum
+        g = f.create_group("minimum")
```
ngbusca: The group `'best fit'` is already the `minimum`. Please add these extra attributes there, otherwise it will be confusing.

londumas:
@ngbusca, just applied your requests. Could you have another look?

py/picca/fitter2/chi2.py
```diff
@@ -122,12 +145,17 @@ def export(self):
         g.attrs['list of free pars'] = self.best_fit.list_of_vary_param()
         g.attrs['list of fixed pars'] = self.best_fit.list_of_fixed_param()
 
+        ## write down all attributes of the minimum
+        dic_fmin = utils.convert_instance_to_dictionary(self.best_fit.get_fmin())
+        for item, value in dic_fmin.items():
+            g.attrs[item] = value
+
```
ngbusca: When you write down all these attibutes, don't we get duplicated info? E.g., fval will be written twice. Could you remove the old instances of the duplicated output parameters?

londumas: It was not written twice but only once: 1st time written, 2nd overwritten.
Anyway it is more beautiful to write it once.

londumas:
@ngbusca, correction made.

py/picca/fitter2/chi2.py
ngbusca: I hadn't realize that, but thanks for caring about the beauty of the code!

Issue #290, opened 2018-01-22
do_xcf.py has no '--from-image' option


jfarr03:
Following from issue #289, I noticed that do_xcf.py has no --from-image option at all. Is this for a reason? 

If not, then I have simply copied across the option from do_cf.py (with my alterations, as specified in #289), and I can add this into a pull request too.

ngbusca:
Once we settle on #289 let's carry the behaviour over to do_xcf.py

Issue #289, opened 2018-01-22
do_cf.py input from '.fits' files using '--from-image'


jfarr03:
At the moment, it is only possible to use do_cf.py's '--from-image' option by specifying a directory or several directories. It would be useful to be able to specify one specific file, as often there are several '.fits' files in a directory, but only one is designed to be read by picca. Even more useful would be to specify multiple input files using wildcards, for example.

I have written a workaround which appears to enable the specifying of individual '.fits' files, which I can put into a pull request if you would like. 

Specifying multiple files can be done currently by putting the '--in-dir' argument into quote marks (so that the wildcards are expanded by glob.glob inside the script, rather than by the shell), though this seems a little unnatural, given that you would not have to do so normally.

ngbusca:
We settled on this behaviour to avoid the limitation that some systems have on the length of the arguments (assuming an alternative where the user would list the input files). It might be unnecessary in the case of `--from-image`. 

I also find unnatural to use quotes and it also complicates sending scripts to batch systems (you'd need to escape the quotes). 

I propose to change the `--from-image` behaviour to specifying a list of files (the system would expand the wildcards), would that solve the problem? If yes, feel free to PR it.

jfarr03:
That makes sense to me. In order to accept a list of arguments, the most logical method seems to be to use the " nargs='+' " option for the input directory. However that would result in args.in-dir being a list of strings (currently it just wants a single string), even if only one argument is provided. It would also change the behaviour when not using --from-image. 

It is fairly straightforward to work around this, by simply looping through the list of strings, and passing each item to the current process (using glob.glob), adding to the list of files each time. I think this should recreate exactly the same output as before - does it sound ok to you as a solution?

londumas:
It will come a little bit out of nowhere, but I would be in favor of removing any options to read different type of files. I think that picca correlations should only read pixels files. Then it is the user responsibility to modify its own file format to be readable by picca. If picca adapts to every file format, it will grow too complicated.
However, if you find a nice way to read your files from py/picca/io.py, why not.

ngbusca:
@jfarr03 I would prefer something that doesn't require touching the behaviour of --in-dir. What I would do, instead, is to change the --from-image option such that it accepts arguments instead of it being a flag (I use nargs="*", but "+" is fine with me), and override the argument of --in-dir if --from-image is present.

ngbusca:
[...]

PR #322, opened 2018-02-15
Modifications to do_cf and do_xcf as per issues 289 and 290

jfarr03:
Addresses the issues described in #289 and #290.

Also fixes a very minor issue discovered in py/picca/data.py: the project function defined in line 419 would cause NaN errors when running do_cf if a skewer contained only 1 cell with IVAR>0.


ngbusca:
Thanks for implementing this @jfarr03. Could you merge your branch with master:

    git checkout master
    git pull
    git checkout from_image_adjustments
    git merge master

so that travis runs the automatic tests? They should be meaningful now...

ngbusca:
@jfarr03 you will also have to `git push origin from_image_adjustments` after the last merge

ngbusca:
@jfarr03 your tests are failing. You can also run them locally to try to debug: `python py/picca/test/test_cor.py`

ngbusca:
@jfarr03 it's not just a matter of running the code, there are checks to make sure that the proposed code does not change the results of standard analyses.

jfarr03:
Sure that makes sense - the tests didn't run on my local machine but the in-browser ones now seem to have passed. The issue locally was coming from my version of Python, though, rather than anything in the picca code, so I don't think it's anything to worry about!

The interplay between --in-dir and --from-image is now as follows:

 - if just '--in-dir {directory name}' is specified, then everything works as before. This can only accept one directory or one file (.fits or .fits.gz) as an input
 - if '--from-image' is specified with no argument, then the from image option is followed using the argument of '--in-dir'. This is how I was using it previously
 - if '--from-image {directory name}' is specified, then the from image option is followed using the argument of '--from-image', and the argument of '--in-dir' is ignored. This can accept multiple directories or files as an input

Ideally it'd be great to make '--in-dir' able to accept multiple directories or files, so then '--from-image' could return to just being a flag, but that would require more work to check it doesn't cause any problems elsewhere, so I understand that you're not keen on changing '--in-dir' at the moment. Let me know if there's anything you think needs more work!

PR #324, opened 2018-02-15
Pk1 d

ngbusca:
I'm trying this PR again since the previous had many problems with git that I don't understand.  Sorry @londumas, can you take a look again?


ngbusca:
@Cyeche I've made some changes to make the code run in python3 and removed the dependency of fftw (replaced by scipy.fftpack.fft). It passes the automatic tests (so the new fft behaves just like the old one) but it would be good to check on data.

ngbusca:
@Cyeche @nathalieGit could you review this branch? Thanks!

Cyeche:
I checked it, it is OK.

I have an additional comment about test_cor for Pk1D (spec and pix format) Il will send it in another mail.

ngbusca:
Thanks @Cyeche !

Issue #327, opened 2018-02-16
unitests for fitter2


ngbusca:
@londumas any ideas what would be the best way to implement this?

londumas:
@ngbusca, yes. Why not simply do a combined fit with metals with one bias_lya and one bias_qso and one SiIII\(1207\). Then compare the h5pt output.
I can take care of this if you want.

londumas:
@ngbusca, doing a combined fit on `cf`, `xcf`, `cf_cross` works very well.
The only small issue is that some of the bins of the distortion matrix are empty.
We could set the rejection to lower and rerun `dmat`, `dmat_cross` and `xdmat` so that at least all bins of the matrix have been filled.
What do you think?

londumas:
@ngbusca, the other small issue is that because we only have 1000 forests the covariance matrix is not positive definite.
This is only an issue for fastMC

ngbusca:
I don't think we need a realistic distorsion matrix, we could just use the identity. I'm surprised that the covarinace is not positive definite, for the tests we go out to 60Mpc/h(?) so there are many less bins to calculate.

londumas:
@ngbusca, I think it is noce to give the fit the distortion matrix calculated in the unittest. It allows to test the export script.
For the covariance matrix it can be linked also to the fact that the forests distributed on a small patch of the sky

londumas:
@ngbusca, could you give me an exemple of running fast MC ? I could add it to the unit tests

ngbusca:
@londumas to run a fast mc you can add:

    [fast mc]
    niterations = 1

to your `chi2.ini` file. I agree that it's better to use the output of the export.py script, but we need to keep the running time for each test under  ~a minute.

PR #331, opened 2018-02-19
Add unittest fitter2

londumas:
Add unittest for fitter2.
Add also unittest for `export` and `export_cross_covariance`
fix issue https://github.com/igmhub/picca/issues/327

```
real	5m37.269s
user	5m19.952s
sys	0m14.319s
```


bin/do_cf.py
```diff
@@ -134,7 +134,6 @@ def corr_func(p):
         for i,f in enumerate(fi):
             sys.stderr.write("\rread {} of {} {}".format(i,len(fi),ndata))
             hdus = fitsio.FITS(f)
-            print('\n',hdus[1:],'\n')
```
ngbusca: @londumas thanks for removing this!

bin/export.py
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 
-from __future__ import print_funcion
+from __future__ import print_function
 import fitsio
```
ngbusca: This fix was already implemented on master, I think your master is not up to date. Could you 
    git checkout master
    git pull
    git checkout add_unitest_fitter2
    git merge master
    git push origin add_unitests_fitter2

londumas:
@ngbusca, impossible to test `fast mc` because the covariance matrix is not positive definite

bin/do_cf.py
londumas: @ngbusca, no problem.

ngbusca:
@londumas if the fit works, it means that the covariance is positive definite in the reduced separation range. We should modify `fast mc` to only generate data in the fitting range...

londumas:
@ngbusca, could you accept the review?
@ngbusca, I don't thnk it is true. See this code bugs on `cholesky`, not on `inv`.


```
from scipy import linalg
from scipy.linalg import cholesky
import fitsio

# first run this on your terminal
#export.py --data $PICCA_BASE/py/picca/test/data/cf.fits.gz --dmat $PICCA_BASE/py/picca/test/data/dmat.fits.gz --out test.fits.gz

path = 'test.fits.gz'
h = fitsio.FITS(path)
co = h[1]['CO'][:]

ico = linalg.inv(co)
cho = cholesky(co)
```

londumas:
@ngbusca, Travis failed because you need to add `h5py`

ngbusca:
@londumas thanks, you can add h5py to the requirements.txt and it should work.  But for the fits, python inverts the covariance, but if it's not positive definite the chi2 is meaningless. We cam test export.py with the output of do_cf from real data or simulations. If we insist in using the output of our pure noise skewers, we should at least add an artificial signal. And I'm still not sure how to dix the covariance.

londumas:
@ngbusca, which version of h5py? I am using `'2.6.0'` in Utah.

londumas:
@ngbusca, Travis test failed because it doesn't know `fitter2`:
```
  File "/home/travis/virtualenv/python3.6.3/bin/fitter2", line 4, in <module>
    __import__('pkg_resources').run_script('picca==0.9', 'fitter2')
  File "/home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages/pkg_resources/__init__.py", line 750, in run_script
    self.require(requires)[0].run_script(script_name, ns)
  File "/home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages/pkg_resources/__init__.py", line 1534, in run_script
    exec(script_code, namespace, namespace)
  File "/home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages/picca-0.9-py3.6.egg/EGG-INFO/scripts/fitter2", line 4, in <module>
    __import__('pkg_resources').run_script('picca==0.9', 'fitter2')
ModuleNotFoundError: No module named 'picca.fitter2'
```

ngbusca:
@londumas it fails on my local computer as well because I don't have PICCA_BASE set. Are we also making PICCA_BASE a requirement in this PR?

londumas:
@ngbusca, nice catch.
I just removed the reference to `PICCA_BASE` and changed it to `resource_filename()`.
I am testing if it works

londumas:
@ngbusca, passed tests on my computer. Waiting for results of Travis

londumas:
@ngbusca, still the same error.
```
  File "/home/travis/virtualenv/python3.6.3/bin/fitter2", line 4, in <module>
    __import__('pkg_resources').run_script('picca==0.9', 'fitter2')
  File "/home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages/pkg_resources/__init__.py", line 750, in run_script
    self.require(requires)[0].run_script(script_name, ns)
  File "/home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages/pkg_resources/__init__.py", line 1534, in run_script
    exec(script_code, namespace, namespace)
  File "/home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages/picca-0.9-py3.6.egg/EGG-INFO/scripts/fitter2", line 4, in <module>
    __import__('pkg_resources').run_script('picca==0.9', 'fitter2')
ModuleNotFoundError: No module named 'picca.fitter2'
```

ngbusca:
@londumas I think I know what the problem is, let me fix it on my side

londumas:
@ngbusca, ok. Could you explain it to me after?

ngbusca:
@londumas now it's finding the module, but it's giving a different error message. Can you take a look?

ngbusca:
@londumas I think you need to generate again the reference file for the fitter2 test after you changed the name of the reference power spectrum (but I'm not 100% sure this is the problem)

londumas:
@ngbusca, the Travis tests fails because they take too much time. This is because of the time required to set the environment, not because of the unittest itself.
The code has been tested in Python2 and in Python3

ngbusca:
@londumas ok, this was a tricky one. The output of `resource_filename` depends on whether you install using `python setup.py install` (that I use in travis) or by setting `PYTHONPATH`, etc. Basically, we can't use `resource_filename` if we want to keep the possibility of installing with `setup.py install` (which I think we want). Is it possible to write the tests without using `resource_filename`?

ngbusca:
@londumas ok, it's a bit more complicated. I can do a workaround in travis using pytest. pytest seems to handle correctly the clash between setup install and `resource_filename`. But the problem is that pytest doesn't write any output on the screen (unless the test fail), and travis kills the tests if they last more than 10 minutes and there's no screen output. I can think of two fixes: either remove the use of `resource_filename` so I can call `python setup.py test` to run the tests OR we split the test_cor.py into many faster test_X.py unittests so that `pytest` will print the output of each and travis won't kill the test.

ngbusca:
@londumas even more info: what I mentioned above won't work. We need to specify the location of PlanckDR12.fits without using `resource_filename`. That would fix all the problems that I'm seeing. Can you do that?

ngbusca:
Ok, I've magically fixed everything, nothing else for you to change ;^)

@londumas Are we ready to merge?

londumas:
@ngbusca, thank you very much! Ready to be merged.

Issue #332, opened 2018-02-20
Missing cross-correlation models in fitter2


mblomqvist:
Fitter2 does not support models for HCD, UV and combinations with veto_lorentz/gaus for the cross-correlation. In r-space, the QSO radiation model is not functioning properly.

ngbusca:
@mblomqvist thanks for the reminder. I'll work on the hcds and uv. 

Can you expand on how the radiation model is not working properly?

mblomqvist:
There is a QSO radiation model, but it 1) takes the wrong arguments 2) uses rp without being passed this argument 3) doesn't apply the drp shift in the actual model 4) asserts LYA even though the absorption tracer could be something else.

ngbusca:
@mblomqvist feel free to PR a fix

ngbusca:
@mblomqvist can we close this issue? Are there pieces still missing?

mblomqvist:
We can close this specific issue now.

PR #334, opened 2018-02-20
hcds in cross correlation (addresses issue #332)

ngbusca:
This PR adds the HCDs model to the cross correlation. It also allows to combine any Pk model with any velocity dispersion model.

This implies a change in the `cross.ini [model]` section:
    [model]
    model-pk = pk_hcd_cross
    velocity dispersion = pk_velo_lorentz

### DIFF
--- a/py/picca/fitter2/pk.py
+++ b/py/picca/fitter2/pk.py
@@ -150,9 +150,9 @@ def G2(k, pk_lin, tracer1, tracer2, dataset_name = None, **kwargs):
 def pk_velo_gaus(k, pk_lin, tracer1, tracer2, **kwargs): 
     assert tracer1 == "QSO" or tracer2 == "QSO"
     kp = k*muk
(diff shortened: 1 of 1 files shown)


ngbusca:
I have also renamed `lls` to `hcd`, so parameters in .ini files need to be updated accordingly (@vserret )

ngbusca:
I feel uneasy adding a UV model for the cross, I don't think there's a model that's supported by the literature.

mblomqvist:
@ngbusca We should discuss the generality of the UV model with the experts. For now, we are unable to reproduce the fit model of the DR12 cross paper which included UV.

londumas:
@ngbusca, how does it work for metals and `velocity dispersion`now?

ngbusca:
@mblomqvist I think we didn't give it enough thought in DR12. We can always reproduce the DR12 fits by checking out the DR12 tagged version and using fitter1.

@londumas currently we can't include these factors in the metal P(k) (but we probably should)

PR #335, opened 2018-02-20
QSO radiation model of fitter2

mblomqvist:
Updated the QSO radiation model of fitter2 to make it operational. Dependence on drp is assumed since it is a cross-correlation model. Can the unittest be comfortably run on a laptop?


londumas:
@mblomqvist, yes. Just do `python test_cor.py` in `py/picca/test/`

PR #340, opened 2018-02-22
Refactor diff

ngbusca:
The  fact that the diff field does not get coadded over multiple observations was causing a crash when reading from spplates. I have refactored the diff and reso fields such that they stay None as the default value. This avoids the crash but I don't know if it breaks something on the Pk1D side (it doesn't break the unittests).


ngbusca:
@Cyeche can you try this branch on a realistic Pk1D analysis?

londumas:
@ngbusca, I feel incompetent to review this PR.

Cyeche:
Ok it doesn’t break the true data...

> 
> @Cyeche <https://github.com/cyeche> can you try this branch on a realistic Pk1D analysis?

ngbusca:
Thanks @Cyeche

PR #342, opened 2018-02-27
Add correlation objects

londumas:
init PR for adding computation of the auto correlation of objects:
     - tested with simulations.
     - missing features like the covariance matrix
     - some issues when random have same THING_ID that data
     - a little bit too long but works fine for the moment


londumas:
@ngbusca, please don't merge right now. Let me do some more tests.

londumas:
@ngbusca, ready to be merged.

ngbusca:
@londumas it would be interesting to implement a Wick-like estimator for the covariance.

ngbusca:
@londumas I can take care of the Wick covariance. Where can I get the input catalogs (D and R?). Are you planning on adding a unittest?

londumas:
@ngbusca, what do you mean by "Wick covariance"? For the simulation I was thinking of getting the covariance from the covariance between the different realization.
I tried to get the covariance from the different healpix in export_co.py but I was stuck because RR is distributed on different RR than DD and DR.
You can get the data and the random from Jean-Marc.
Do you want me to build unitest? It wouldn't be long.

ngbusca:
@londumas to get the covariance from simulations you would need hundreds of thousands of them. I don't understand 

> RR is distributed on different RR than DD and DR

In a Wick approach, you just separate the contributions to the covariance from pairs of galaxies, triplets and quadruplets, just like we do in lyman-alpha. I know galaxy people don't do that, I don't know why, but I don't know if it was explored already.

londumas:
@ngbusca, are you ok to merge? and you can add the Wick covariance when you have the time to code it.

londumas:
@ngbusca, sorry. Let me build the unittest before.

londumas:
@ngbusca, ready to be merged.

ngbusca:
@londumas it's great to have this functionality, thanks! I'd like to test a but the code before merging.

londumas:
@ngbusca, don't hesitate. You can look at the unittest to see how to run it.
It is a little bit too slow for the time being. We could fix that by going from a class based on one quasar to a class of list of quasars. I am sure we could gain a lot that way.

londumas:
@ngbusca, by the way this PR removes `bin/do_xMetalGrid.py` and `bin/export_metal.py`.
It also add a function in `py/picca/piutils.py` to give the angmax for a given calculation of the correlation.

londumas:
@ngbusca, ok to merge?

PR #343, opened 2018-02-27
Add test for specific estimator of noise implemented in Pk1D analysis

Cyeche:
Read spectra with different exposures and compute the difference between the exposures
Add a test with do_deltas to check that the difference is correctly computed


londumas:
@Cyeche, perfect but could you put all the input files in `test/data/test_Pk1D/` or something with a similar name. It would help keep track of what the files are.
Where did you get the files from? Maybe add a README to tell where you downloaded the files from.

Cyeche:
Done. The code is self-explanatory ;-)

> 
> @Cyeche <https://github.com/cyeche>, perfect but could you put all the input files in test/data/test_Pk1D/ or something with a similar name. It would help keep track of what the files are.
> Where did you get the files from? Maybe add a README to tell where you downloaded the files from.

Cyeche:
Thanks, I will correct it! This shows that this test is useful!  And moreover it rings a bell for another issue that I got...

Issue #344, opened 2018-02-28
Add option to control the verbosity of fitter2


ngbusca:
I like to see the fit progressing in the screen, but we need an option to control the amount of output...

PR #345, opened 2018-02-28
Fitter2 verbosity

ngbusca:
Adresses issue #344 
To set the verbosity, add the following section to chi2.ini

    [verbostiy]
    level = 1

The current bevavior with full minimization verbosity corresponds to not having this section or level=1. Any other level suppresses the minimization output.


ngbusca:
I've also set the verbosity to zero in the travis tests (it was generating too much output).

PR #348, opened 2018-03-01
Fix convert desi data

londumas:
Gets RA and DEC for desi simulations correctly and fast.
Works if giving all the ztarget or only a part.
Please test before merge.

```
from picca import utils
path_truth = "/global/projecta/projectdirs/desi/datachallenge/dc17b/targets/truth.fits"
path_targets = "/global/projecta/projectdirs/desi/datachallenge/dc17b/spectro/redux/dc17b/ztarget-dc17b.fits"
path_drq       = "test.fits.gz"
utils.desi_from_truth_to_drq(path_truth,path_targets,path_drq)
```
```
from picca import utils
path_truth = "/global/projecta/projectdirs/desi/datachallenge/dc17b/targets/truth.fits"
path_targets = "/global/projecta/projectdirs/desi/datachallenge/dc17b/targets/targets.fits"
path_drq       = "test.fits.gz"
utils.desi_from_truth_to_drq(path_truth,path_targets,path_drq)
```


londumas:
@ngbusca, I can add an `assert` to be sure that the user is giving the good `ztarget` file.
The assert would compare the size of the targets size and the truth file. They should be exactly the same.

ngbusca:
@londumas that's not enough since they can be sorted differently no?

londumas:
@ngbusca, They are supposed to be in the same order.
But anyway in this PR we get RA and Dec from a dictionary.
Please have a look at the new code.
The assert will assure in a way that the user is using a file with the good format.

londumas:
@ngbusca, ok to merge?

py/picca/utils.py
```diff
@@ -76,34 +77,53 @@ def desi_from_truth_to_drq(truth,targets,drq,spectype="QSO"):
...
-    vacTargets = fitsio.FITS(targets)
 
-    ## Info of the primary observation
-    thid  = vac[1]["TARGETID"][:]
-    ra    = vacTargets[1]["RA"][:]
-    dec   = vacTargets[1]["DEC"][:]
-    zqso  = vac[1]["TRUEZ"][:]
+    w  = sp.ones(vac[1]["TARGETID"][:].size).astype(bool)
+    print(" start                 : nb object in cat = {}".format(w.sum()) )
+    w &= numpy.core.defchararray.replace(vac[1]["TRUESPECTYPE"][:].astype(str),' ','')==spectype
```
ngbusca: @londumas could you simplify this line by doing:

    vac[1]['TRUESPECTYPE'][:].strip()==spectype

ngbusca: you won't need to import numpy

londumas:
@ngbusca, changed applied. `array.strip()` doesn't exist. `strip()` acts on strings not on array.
I used `sp.char.strip(array)`

py/picca/utils.py
```diff
@@ -76,34 +76,53 @@ def desi_from_truth_to_drq(truth,targets,drq,spectype="QSO"):
...
+    for i,t in enumerate(thid):
+        if t not in keys_from_TARGETID_to_idx: continue
+        idx    = from_TARGETID_to_idx[t]
+        ra[i]  = raTargets[idx]
+        dec[i] = decTargets[idx]
+    if (ra==0.).sum()!=0 or (dec==0.).sum()!=0:
+        w  = ra!=0.
+        w &= dec!=0.
+        print(" and RA and DEC        : nb object in cat = {}".format(w.sum()))
+
```
ngbusca: Could you fix the vertical alignment according to https://www.python.org/dev/peps/pep-0008/ , please? I know we haven't used that consistently so far, but we'd better start at some point...

londumas: In what does it consists? I don't want to read all the doc.

ngbusca: More than one space around an assignment (or other) operator to align it with another.

    Yes:

    x = 1
    y = 2
    long_variable = 3
    No:

    x             = 1
    y             = 2
    long_variable = 3

londumas:
@ngbusca, done.

ngbusca:
@londumas is everything ok from your end? I don't think this is tested by the unittests (I don't argue that it should), but did you test that your changes do what you want them to do?

londumas:
Yes it works and do what you requested.

Issue #351, opened 2018-03-02
Unittest failure


mblomqvist:
The unittest was run in python 2.7 on a topic branch that is up-to-date with master and only has a minor change in fitter2, but stops with the failure
```
======================================================================
FAIL: test_cor (__main__.TestCor)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "test_cor.py", line 74, in test_cor
    self.send_xcf_angl()
  File "test_cor.py", line 612, in send_xcf_angl
    self.compare_fits(path1,path2,"do_xcf_angl.py")
  File "test_cor.py", line 243, in compare_fits
    self.assertTrue(allclose,"{}: Header key is {}, maximum relative difference is {}".format(nameRun,k,diff.max()))
AssertionError: do_xcf_angl.py: Header key is DA, maximum relative difference is 0.00279007921419
```

londumas:
could you do the same test on master?

mblomqvist:
Same failure on master. There are warnings for every code of the type
WARNING: do_xcf_angl.py: Header key is RP, arrays are not exactly equal, using allclose

londumas:
@mblomqvist, what happens if you comment the test of `do_xcf_angl.py`?

ngbusca:
@mblomqvist were the tests passing on your system before?

mblomqvist:
This is the first time I'm running it on my system

londumas:
@mblomqvist, It can be a difference of generation of random numbers.

ngbusca:
@mblomqvist can you confirm that all your package versions correspond to what's listed in the requirements.txt file?

mblomqvist:
I commented out
self.compare_fits(path1,path2,"do_xcf_angl.py")
but the unittest stopped at send_metal_xdmat with a segmentation fault: 11

ngbusca:
@londumas it's possible but a bit unlikely. The tests pass at nersc on python 3, at Utah on python 2, on my mac and on the linux console of travis...

ngbusca:
@mblomqvist that segfault seems like a separate problem...

mblomqvist:
The numbered requirements are scipy 0.17.0 and iminut 1.2 and my laptop fulfils both

ngbusca:
@mblomqvist then your master is not up-to-date:

scipy==0.19.1
numpy==1.13.1
iminuit==1.2
healpy==1.11
fitsio==0.9.11
numba==0.35.0
h5py==2.6.0
future
[...]

PR #352, opened 2018-03-02
Fitter2 BAO amplitude parameter

mblomqvist:
Add the BAO amplitude parameter bao_amp to fitter2. Unittest failed (also on the master branch) on my laptop, see issue #351


ngbusca:
I think we'd need to include the bao_amp parameter in the corresponding .ini file in the unittests. @londumas can you comment?

ngbusca:
@mblomqvist the failure here is not related to the ang_cf but to the fact that the unittest for fitter2 doesn't know the bao_amp parameter (is not listed in the right .ini file)

mblomqvist:
Right, I didn't think about that.

ngbusca:
I'll fix it on my end in a sec

ngbusca:
I've added the bao_amp parameter to the test/config_cf.ini but that's not enough.  The test finds a new parameter that isn't in the reference file so it's not happy. We need @londumas to tell us what's the best way forward, maybe we need to regenerate the reference file?

mblomqvist:
This pull request was supposed to be a trivial update but turned out to be head scratcher ;)

londumas:
@ngbusca, You need to regenerate the result file.

ngbusca:
@londumas can you do that directly on this PR? You are the unittest master ;^)

londumas:
@ngbusca, @mblomqvist. It isn't hard. Just change line 51 by:
```
self._test = False
self._branchFiles = './keep_tests/'
```
and do a:
```
mkdir ./keep_tests/
```
run the unittest on your laptop and copy the new result to the data and commit.

ngbusca:
Thanks @londumas. @mblomqvist I'll take care of that

ngbusca:
@londumas do I need to add the line

    self._branchFiles = './keep_tests/'

after line 51?

londumas:
@ngbusca, yes it allows to save the output to a given directory not to a random directory in `/tmp`

ngbusca:
@londumas thanks, that worked perfect. Let's see if travis likes it.

mblomqvist:
Thank you both for fixing it.

PR #355, opened 2018-03-03
add velocity dispersion for metals

londumas:
add velocity dispersion for metals in fitter2.


ngbusca:
I'm ok with this change, but you should remember that you can't use the cached_xi in this case (so the fit will be much much much slower). Do you confirm that?

londumas:
I do.

mblomqvist:
Should it not be set by 'velocity dispersion'? It seems a bit inconsistent to have a separate option for the metals.

londumas:
@ngbusca, changed so that if velo-dispersion is set on the model, then it is set to the metals.
Ok to merge?

ngbusca:
I agree that it makes more sense to add the dispersion to the metals if it's added to the cosmological model. For now, the user has to remember not tu use the cached version of xi if the dispersion is added. I'll work on improving that

Issue #336, opened 2018-02-21
Add isotropic BAO fit to fitter2


mblomqvist:
Fitter2 does not support BAO fits expressed in (alpha_iso, 1+epsilon).

PR #360, opened 2018-03-06
Add isotropic BAO fit

mblomqvist:
Add option to fit for (aiso,1+epsilon) instead of (ap,at). Set in the ini file by

[model]
bao dilation = aiso

Any other input defaults to fitting (ap,at).

### DIFF
--- a/py/picca/fitter2/chi2.py
+++ b/py/picca/fitter2/chi2.py
@@ -82,15 +82,26 @@ class chi2:
         for d in self.data:
             d.best_fit_model = self.best_fit.values['bao_amp']*d.xi_model(self.k, self.pk_lin-self.pksb_lin, self.best_fit.values)
 
-            ap = self.best_fit.values['ap']
-            at = self.best_fit.values['at']
-            self.best_fit.values['ap'] = 1
-            self.best_fit.values['at'] = 1
             snl = self.best_fit.values['sigmaNL_per']
             self.best_fit.values['sigmaNL_per'] = 0
-            d.best_fit_model += d.xi_model(self.k, self.pksb_lin, self.best_fit.values)
-            self.best_fit.values['ap'] = ap
-            self.best_fit.values['at'] = at
+            if d.bao_dilation == 'aiso':
+                aiso = self.best_fit.values['aiso']
+                eps = self.best_fit.values['1+epsilon']
+                self.best_fit.values['aiso'] = 1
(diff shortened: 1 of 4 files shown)


mblomqvist:
OK to merge?

ngbusca:
@mblomqvist Thanks for implementing this option. I might have a few requests to change the implementation of this option such that it follows the logic of fitter2, but I need some time (a few days) to digest it.

mblomqvist:
@ngbusca OK, thanks for the update. I tried to adhere to the logic by making it a model option and creating a dedicated function in utils.

londumas:
@mblomqvist, I agree with Nicolas, why not do something similar as `velocity dispersion`, where
the code goes to set the function according to the name of the function given in `model`.
```
         if 'velocity dispersion' in dic_init['model']:
             self.pk *= getattr(pk, dic_init['model']['velocity dispersion'])
```

londumas:
linked to issue https://github.com/igmhub/picca/issues/336

ngbusca:
@londumas I was thinking along those lines, but before @mblomqvist changes anything, let me do more specific suggestions. I'll need a bit of time...

ngbusca:
@mblomqvist ok, here's what I think would be an implementation more in line with the code style in fitter2. It would also simplify the code and remove all the ifs:

 * add the section `[cosmo-fit type]` to the chi2.ini file that will specify the name of the function that will take the full model parameters and return `ap, at`.
 * add a placeholder in utils.py to hold that function, e.g. `cosmo_fit_func`, initially `None`
 * wherever the code needs to know `ap`, `at`, call that function (`ap, at = utils.cosmo_fit_func(pars)`) instead of whatever there is now. 
 * implement a function that will return `ap` and `at` (e.g. `ap_at`) for the current fit type. That function will be totally trivial: `def ap_at(pars): return pars['ap'], pars['at']`
 * implement a function that will return `ap` and `at` for the `aiso`, `epsilon` fit, e.g. `def aiso_epsilon(pars): return ...`

Then you just need to specify the desired function in the `[cosmo-fit type]` section of the `chi2.ini`... makes sense ?

ngbusca:
I forgot to mention that you will need to modify `parser.py` to set `cosmo_fit_func` to the specified function in `chi2.ini`

mblomqvist:
OK, thanks for the input. I'll take a look.

mblomqvist:
Closing this PR to make way for an alternative implementation.

Issue #341, opened 2018-02-26
Protect do_xcf against empty in_dir


jmarclegoff:
If in_dir is empty, the user gets a cryptic message :
```
File "../../picca/bin/do_xcf.py", line 206, in <module>
    d_min_pix = cosmo.r_comoving(z_min_pix)
File "/global/u2/l/legoff/.conda/envs/myenv/lib/python2.7/sitepackages/scipy/interpolate/polyint.py", line 79, in __call__
    y = self._evaluate(x)
  File "/global/u2/l/legoff/.conda/envs/myenv/lib/python2.7/sitepackages/scipy/interpolate/interpolate.py", line 610, in _evaluate
    below_bounds, above_bounds = self._check_bounds(x_new)
  File "/global/u2/l/legoff/.conda/envs/myenv/lib/python2.7/sitepackages/scipy/interpolate/interpolate.py", line 642, in _check_bounds
    raise ValueError("A value in x_new is above the interpolation "
ValueError: A value in x_new is above the interpolation range.
```
I suggest adding after line 149 of do_xcf.py, something like : 
```
    if (len(fi) == 0) :
        print "no file found in in_dir"
        exit(0)
```

londumas:
@jmarclegoff, with @ngbusca we want to move reading the files to `py/picca/io.py` at some point.
That is where the test should be at the end.

PR #361, opened 2018-03-06
Fix loading data xcf

londumas:
Load data in `do_xcf` and in `do_xdmat` from io.py.
Might have broken loading simulations fron @jfarr03.
Correct some small bugs in `do_xcf_angl`


jfarr03:
@londumas Will the changes made here still work with the 'from-image' option? I can't see that it will but I may be missing something!

londumas:
@jfarr03, I don't know.
The issue is that this option should act on `io.read_deltas()`.
Could you test if it works? And if it doesn't commit to this branch the changes that would make it work.

jfarr03:
@londumas Sure I understand - io.read_deltas doesn't seem to have any reference to 'from-image' in it, so I guess it probably wouldn't work! I'll make the edits necessary and will commit. 

Are the changes made here also relevant to do_cf.py then? I know the reading of the deltas works exactly the same in do_cf and do_xcf as I edited it for both of them a couple of weeks ago.

jfarr03:
@londumas Oh yes I know, I just was wondering whether you might also want to implement the same changes to do_cf.py, as that works in the same way as do_xcf and do_xcf needed changing.

londumas:
@jfarr03, I don't intend to change anything to `do_cf`.
The reason is that I don't work currently on the auto-correlation and don't want to change anything for others.

jfarr03:
@londumas Ok cool no worries - just thought I'd mention it!

ngbusca:
@jfarr03 is do_cf broken in the same way? In principle it shouldn't since we haven't updated it to use read_deltas (I think...)

ngbusca:
by the way, does this fix your problem with the covariance?

londumas:
@ngbusca, we had a look together with @jfarr03, he was sending `do_xcf.py` with the wrong parameters. This must explains the issue with his covariance.
`do_cf.py` has not been touched.

mblomqvist:
@londumas Can you elaborate a bit on what this pull request does?

londumas:
@mblomqvist, There are two aspects:
    - a simple cleaning of `do_xcf.py` and `do_xdmat.py` to read data from `io.py`
    - correct two bugs in `do_xcf_angl.py`: definition of redshift range and removing mean delta versus lambda_obs

jfarr03:
@ngbusca Both do_cf and do_xcf worked for me previously, so as far as I was concerned neither was 'broken' as such. 

I was just wondering as it seemed like the new pull request deletes what I wrote a few weeks ago to include the '--from-image' option in do_xcf, and I know that do_cf works in the same way as I copied the do_xcf behaviour from what I wrote for do_cf!

ngbusca:
@jfarr03 do you confirm that the option `--from-image` works for this branch as it does for the current master?

londumas:
@ngbusca, it does not work yet. Some work head.

ngbusca:
@londumas ok... we might want to add a unittest at some point then...

jfarr03:
@ngbusca @londumas I can confirm that the --from-image option doesn't work yet - it should just be a case of copy and pasting (with a few extra subtleties) lines 128-184 from the old do_xcf into io.read_deltas though to fix the problem. I can do that now if that would be helpful?

londumas:
@jfarr03, yes you can do this now if you have the time.

ngbusca:
@londumas this PR breaks a functionality in the current master, I think the ball is more on your side of the field to make the necessary fixes ;^)

londumas:
@ngbusca, I agree, I will take care of this.

londumas:
@jfarr03, I just made the correction to this PR to read `from_image`. Could you test it?
Please do:
```
git checkout fix_loading_data_xcf
do_xcf.py <all_option> --from-image <more options>
```
This PR should be also fixing https://github.com/igmhub/picca/issues/341, could you test @jmarclegoff?

jfarr03:
@londumas It doesn't seem to be quite working for me yet - lines 602-609 are still only valid for the 'from-fitsio' option it seems

londumas:
@jfarr03, lines 602-609 of which file and I don't see the `from-fitsio` option, where is it?

jfarr03:
@londumas Sorry lines 602-609 in py/picca/io.py. In particular, the line 605:
`dels += [delta.from_fitsio(h) for h in hdus[1:]]`
The 'from_fitsio' method is defined in py/picca/data.py, under the 'delta' class I think.

londumas:
@jfarr03, could you test again? Thank you for your help.

jfarr03:
@londumas The loading of data certainly seems fine now!

However, do_xcf does now crash in line 133 when determining forest.dll (which seems to be a much more detailed process now than it was before). This happens as we have some 'skewers' which only have 1 cell in them due to a cut-off in the wavelength that we are imposing. This should be relatively easy to avoid, but it made me wonder: is there a particular reason for the new, longer method to determine forest.dll?

londumas:
@jfarr03, I had to change the way to get forest.dll because it is not given by `io`.
I am in favor of removing forest with less than 50 pixels in `from_image`. At least let's remove forest with only one pixel. What do you think @ngbusca?

jfarr03:
@ngbusca @londumas I think that seems sensible to me, it shouldn't be hard to implement. I'll write it into our processing tomorrow morning and will test again once that has completed

jfarr03:
@londumas This doesn't seem to work for me yet - having reprocessed our skewers to have at least 2 pixels, do_xcf.py now crashes at line 141:

`ll,st, wst   = prep_del.stack(xcf.dels,delta=True)`

The exact error I get is: 

`Traceback (most recent call last):
  File "/global/homes/j/jfarr/Programs/picca_prcheck/bin/do_xcf.py", line 136, in <module>
    ll,st, wst   = prep_del.stack(xcf.dels,delta=True)
  File "/global/homes/j/jfarr/Programs/picca_prcheck/py/picca_prcheck/prep_del.py", line 132, in stack
    nstack = int((forest.lmax-forest.lmin)/forest.dll)+1
TypeError: unsupported operand type(s) for -: 'NoneType' and 'NoneType'`

Clearly forest.lmax and forest.lmin are defined in lines 139 and 140, but it seems they are not being properly passed to the 'stack' function

londumas:
@jfarr03, could you print the values of `forest.lmax`, `forest.lmin`,`forest.dll`?

jfarr03:
@londumas Printing the values just before stack is called gives the following:

```
dll: 0.000340462
lmax 3.74182939421
lmin 3.55026836622
```

Printing the values in the first line of 'stack' returns the following:

```
dll: None
lmax None
lmin None
```

jfarr03:
@londumas @ngbusca This all seems to be working well for me now!

londumas:
@ngbusca, ready to be merged.

Issue #359, opened 2018-03-05
add galaxies to fitter2


londumas:
Currently fitter2 has some functions that `assert tracer==QSO`.
What would be the best way to add galaxies?
We could add a list of tracers in `py/picca/constants.py` like:
```
bool_tracer = ['QSO','LOWZ','CMASS','ELG','GALAXY']
```
and then in `py/picca/fitter2/pk.py:::pk_velo_gaus()` we could have:
```
assert tracer1 in bool_tracer or tracer2 in in bool_tracer
```
What do you think @ngbusca?

PR #362, opened 2018-03-09
add boolean tracer to fitter2

londumas:
Add the list of individual tracers to fitter2. Fix issue https://github.com/igmhub/picca/issues/359.
`tracer` is a very bad name but I don't really have other ideas.
We could also call them `boolean_tracer`. Any idea @mblomqvist and @ngbusca?
@mblomqvist, do you want me to add `SLA` or `LLS`.

This list will serve as a nomenclature as `constants.absorber_IGM` is doing.


ngbusca:
Don't you think it would be better to leave the names arbitrary but have a tracer type? The way to implement this would be to create a class tracer with two attributes: the name (arbitrary) and the tracer type (collapsed_tracer or absorber_tracer). Then, the Pk or xi functions would receive instances of this class instead of strings, and they would make the assertions on the tracer_type attribute. Makes sense ?

ngbusca:
The user would have to specify in the .ini both the tracer name and the tracer type

londumas:
@ngbusca, I am not sure to understand why would you want a complicated class when a dictionary or a list does the job.

ngbusca:
@londumas My point is that all the code needs to check is whether the tracer is a discrete tracer or not. If the user specifies that the tracer is discrete that's enough, otherwise you will have to add a tracer name to the list every time there's a new tracer (say `QSO_TYPE_X` or `ELG_HIGH_LUMINOSITY`, etc. etc.)

By the way, I prefer discrete tracer to collapsed tracer since the former is agnostic to the physics of the tracer (what if we want to use a discrete tracer that is not a collapsed object? e.g. a catalog of small lyman-alpha absorptions).

Finally, one very simple way out is to remove the assertions and trust the user knows what he's doing.

londumas:
@ngbusca, Thank you for the answer.
I added giving the type of tracers in the `config.ini` file, the issue now is that we have to pass it to `xi.py` and `pk.py` and this would change a lot of things.
An other thing we could do would be to remove the two assert. But this is dangerous.

ngbusca:
@londumas as for how to pass it, I would do a minimal change. All functions receive an argument tracer1 and tracer2, which is currently a string. If we used a dictionary instead of a string that would solve the problem. The dictionary would have a key `name` and a key `type`... what do you think?

ngbusca:
for the assertions, coadacy likes try-except blocks instead, because that works in compiled python, while the asserts are ignored. Since we are not compiling for now, I would leave the asserts and ignore the coadacy suggestion.

londumas:
@ngbusca, I actually just had the same idea of using a dictionary.
It still produces a lot of changes, are you ok with that?

ngbusca:
@londumas yes, I'm ok with that!

londumas:
@ngbusca, ready to be merged. Could you do a review?

Issue #328, opened 2018-02-16
fitter2 cuts bins according to the mean rp,rt grid and not to the center


londumas:
In fitter, we were cutting bins according to the` rp, rt, r, mu` value at the center of the grid. not at the mean.
This allows to always have the same number of bins even if some data sets have different `rp, rt` grids.
Do we want to put back this way of cutting?
If yes, we should keed track of `'NT', 'NP', 'RTMAX', 'RPMIN', 'RPMAX'` in `export.py` and get the center of the grid from these parameters in `py/picca/fitter2/data.py`

ngbusca:
I think the fix for this is quite simple, instead of doing, e.g., rmin=10 we could do rmin=9 or 9.5. Then we don't need to change anything in the code.

londumas:
@ngbusca, I would prefer the code to be based on the center of the grid. We did it for the previous papers.

ngbusca:
@londumas well, we did it in the previous papers just because we didn't realize at the time that we could have just change a little bit rmin.

mblomqvist:
But then, formally, we would need to quote different rmin (e.g. 10 and 9.5) for different fits in the papers.

ngbusca:
Yes, we would need a slightly lower rmin sich that it's equivalent to the previous fits.

londumas:
@ngbusca, why don't you want to do it as we did in DR12 papers? We would just have to change `export.py` and `fitter2/data.py`.

ngbusca:
Because I think that it really makes more sense to cut on the mean pixel separation than the pixel. Also it means we don't need to change anything in the code

londumas:
@ngbusca and @mblomqvist, the thing I don't like about "a slightly lower rmin" is that how to decide the value of this lower rmin? `0.1`, `1.0`? We would have to tune this parameter.
Cutting on the center of the grid is better, there are no parameter to tune.
Currently with my MgII fit I have some correlations that have `3179` bins, other that have `3180` bins.

However if you really don't like cutting center of the grid I can take care of finding the "slightly lower rmin".

mblomqvist:
I came across a fit where I had 3179 bins instead of 3180 bins, but the missing bin was at "slightly larger rmax" instead.

londumas:
I made a PR (https://github.com/igmhub/picca/pull/374) to give an idea of what I have in mind when I was talking of cutting on center of the bins. This PR gives the same behavior as in fitter.

[...]

PR #366, opened 2018-03-13
add storing grid parameter

londumas:
Add storing grid parameters (i.e. `RPMIN, RPMAX, RTMAX, NP, NT`) to exported files.
This is for two reasons:
- will help solving issue https://github.com/igmhub/picca/issues/328 if we decide to cut on the center of the grid
- thanks to this PR all useful infos of the correlation are in the exported file. This is useful for plotting, no need to read one file for the grid and another for the errors.

### DIFF
--- a/bin/export.py
+++ b/bin/export.py
@@ -43,7 +43,6 @@ if __name__ == '__main__':
     head = h[1].read_header()
     nt = head['NT']
     np = head['NP']
-    rt_min = 0.
     rt_max = head['RTMAX']
     rp_min = head['RPMIN']
     rp_max = head['RPMAX']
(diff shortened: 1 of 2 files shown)


bin/export.py
```diff
@@ -40,18 +40,20 @@
     we = sp.array(h[2]['WE'][:])
     hep = sp.array(h[2]['HEALPID'][:])
 
+    head = h[1].read_header()
+    nt = head['NT']
+    np = head['NP']
+    rt_min = 0.
```
ngbusca: what is rt_min used for? It doesn't seem to be written in the output...

londumas: I just removed this variable

londumas:
@ngbusca, ok to merge?

PR #368, opened 2018-03-15
Growth factor with Dark Energy

ngbusca:
This PR implements the growth_factor_de function to allow for growth_factors for arbitrary Om and OL.

There are two main changes:
  * I've changed the name `growth_function_no_de` to `growth_factor_no_de` of the old function (so users will have to update their .ini)
  * The new function `growth_factor_de` requires the values of `Om` and `OL`. The user would have to specify them in the list of parameters.

@londumas can you take a look? This should be useful for your low-z analyses.

### DIFF
--- a/py/picca/fitter2/xi.py
+++ b/py/picca/fitter2/xi.py
@@ -134,7 +134,7 @@ def cache_growth_factor_de(function):
             cache[pair] = (Om, OL)
             cache[1] = cached_growth_factor_de(*args, **kwargs)
         
-        return cache[1](args[0])
+        return cache[1](args[0])/cache[1](kwargs['zref'])
             
     return wrapper
 
@@ -168,10 +168,8 @@ def cached_growth_factor_de(z, zref = None, **kwargs):
         a = 1/(1+z[i])
         D1[i] = 5/2.*Om*hubble(z[i], *pars)*quad(dD1, 0, a, args=pars)[0]
 
-    D1_interp = interp1d(z, D1)
-    D1_ref = D1_interp(zref)
-    D1_interp = interp1d(z, D1/D1_ref)
-    return D1_interp
+    D1 = interp1d(z, D1)
+    return D1
 
 ### Lya bias evolution
 def bias_vs_z_std(z, tracer, zref = None, **kwargs):
--- a/py/picca/fitter2/xi.py
+++ b/py/picca/fitter2/xi.py
@@ -1,5 +1,7 @@
 import scipy as sp
 from .utils import Pk2Xi, bias_beta
+from scipy.integrate import quad
+from scipy.interpolate import interp1d
 
 def xi(r, mu, k, pk_lin, pk_func, tracer1=None, tracer2=None, ell_max=None, **pars):
     pk_full = pk_func(k, pk_lin, tracer1, tracer2, **pars)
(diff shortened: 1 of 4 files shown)


py/picca/fitter2/xi.py
```diff
@@ -119,9 +121,58 @@ def xi_qso_radiation(r, mu, k, pk_lin, pk_func, tracer1, tracer2, ell_max=None,
...
 ### Growth factor evolution
-def growth_function_no_de(z, zref = None, **kwargs):
+def growth_factor_no_de(z, zref=None, **kwargs):
     return (1+zref)/(1.+z)
 
+def cache_growth_factor_de(function):
+    cache = {}
+    def wrapper(*args, **kwargs):
+        Om = kwargs['Om']
+        OL = kwargs['OL']
```
londumas: Apparently `OL` is something in Python: It is in collored in blue.
Would it be possible to put it as `Ol`?
Obviously, this is not important.

londumas:
@ngbusca, Gracias.
I will try this PR and tell you if this work.

py/picca/fitter2/xi.py
ngbusca: I can't find anything that would explain why it's blue... and `Ol` looks much less pretty than `OL`. I'll keep looking while you review other things

ngbusca:
Here's a plot comparing the calculation from this PR (solid lines) and what I get by running directly `camb` at various redshifts (circles).

image

py/picca/fitter2/xi.py
londumas: I agree `Ol` dosn't look nice.

py/picca/fitter2/xi.py
```diff
@@ -119,9 +121,56 @@ def xi_qso_radiation(r, mu, k, pk_lin, pk_func, tracer1, tracer2, ell_max=None,
...
+def growth_factor_de(*args, **kwargs):
+    return cached_growth_factor_de(*args, **kwargs)
+
+def cached_growth_factor_de(z, zref = None, **kwargs):
+    '''
+    Implements eq. 7.77 from S. Dodelson's Modern Cosmology book
+    '''
+    print('Calculating growth factor for Om = {}, OL = {}'.format(kwargs['Om'], kwargs['OL']))
+
+    def hubble(z, Om=0.3, OL=0.7):
```
londumas: maybe it would be better to have the values `Om=0.315` and `OL=1.-Om=0.685` by default, to be consistent with the rest of picca.

ngbusca: good point. Actually I will remove the default values to force the user to specify them

ngbusca: In fact, we should use the same `Om`, `OL` as what was used to calculate the `P(k)`, shouldn't we? But I don't think this is written in the input p(k) file. I think we should write in the header of the P(k) fits file all the parameters from the .ini that was used to run camb...

mblomqvist: I can take care of adding those values to the P(k) fits file.

mblomqvist:
Would it be better to require 'Om' and 'Ok'?

londumas:
@ngbusca, thanks to the last merged branch, we can now read `Om` and `OL` from the power-spectrum. Do you want to take care of it? I can if you don't have the time.

Issue #367, opened 2018-03-14
redshift of parameters for combined fit


londumas:
Currently in `fitter2` there is a `zref` for each input correlation.
- it would be nice to put the input `zref` in the h5py file so the user knows at what redshift the parameters are give.
- At what redshift are the parameters given in the case of two input correlations with different `zref` ? maybe we should move `zref` from the config file to the `chi2` file. Or have this parameters unique in some way.

mblomqvist:
I suggested some time ago that zref should be read from the input power spectrum header. We have not followed up on that yet.

ngbusca:
The way I see it (it might not be what's implemented), the bias, beta, etc. should be given at the redshift of the measurement (call it zeff). We need to read the zref of the fiducial P(k) to know by how much we need to rescale it to zeff (but that's trivial). For parameters that are shared and need to be converted from zeff(measurement 1) to zeff(measurement 2) an assumption on the z evolution of that parameter is required.

ngbusca:
I agree with @londumas that we should propagate the zeff's to the output file.

londumas:
@ngbusca, do you mean that currently the result parameters are given at `zeff`, the redshift of the input power-spectrum?

mblomqvist:
Let's start by agreeing on the names: zref = reference redshift of the fiducial P(k), zeff = effective redshift of the measurement. Currently, fitter2 has no input for zeff (as far as I know).

londumas:
Here is how I see things:
- there should be a zeff (~<z_bins>) for each measure
- there should be a zoutput at which the parameters are given. It could be 0 or zeff if there is only one measure.
- zref (from P(k)) shouldn't be a parameter since it is given in the P(k) file. In the future it could also be exactly zoutput by reading P(k) from a grid in redshift. @mblomqvist, would it be hard to produce such a fits file: P(k,z) for z in [0,10] with dz=0.01 for example?

ngbusca:
@londumas we don't need to generate `P(k)` at different `z`, all we need is the growth factor.

ngbusca:
Here's a plot with the ratio of P(k) at z=2 and P(k) z = 1. Once correcting for the growth rate, the differences are smaller than 0.3%, and only at the largest scales.

![image](https://user-images.githubusercontent.com/4727509/37453746-f7d344c8-2838-11e8-9422-2c03ff5fef0f.png)

PR #371, opened 2018-03-16
read zref from power-spectrum

londumas:
Reads `zpk` from input power spectrum.
Remove reference to `zref`.
Initial commit for test, in the next commit, will add saving `zpk` to h5py file.
Fix issue https://github.com/igmhub/picca/issues/367

Gives the same result h5py. I will now adds storing this parameter to the file.


ngbusca:
@londumas can you take the chance to check that the fit parameters are given at zref from the masurement and not at zpk?

mblomqvist:
Apologies for being a stickler about the names ;) but zref and zpk are by definition the same quantity. The measurement is at zeff. We've been using the zref name for 5+ years, but if you prefer zpk in the code I fine with it.

ngbusca:
@mblomqvist I also like zref better than zpk for the redshift of the fiducial P(k)... 

By the way, if you add Om, etc. to the fiducial P(k) fits file, could you actually add all the parameters in the camb .ini file so we can reproduce the P(k) in 42 years if we need to?

mblomqvist:
@ngbusca Yes, I'd add more parameters. There are a few more obscure ones that I'm not entirely sure about though. In any case, CAMB has different versions and as a result the power spectra may not be exactly identical 42 years from now.

londumas:
@ngbusca, what is the use of `_wrap_chi2`? It doesn't seem to be used.

londumas:
@ngbusca, just changed the name from `zpk` to `zref`.
Do you want me to do anything else?

ngbusca:
@londumas with the current setup, the parameters (bias, etc) are given at `zref`. At some point we should change that so they are given at the redshift of the measurement... what do you think?

londumas:
@ngbusca, I totally agree. But then I have to add a common parameter for all data-set. Not one per data set. We could call it `zeff`? What do you think? I can take care of this.

londumas:
Here is the relation between `z` (redshift of a bin), `zeff` redshift where to give the parameters and `zref` the redshift of the input power spectrum: 
image
image

```
\xi_{biased}(z) = b_1(z_{eff})b_2(z_{eff})
\left(\frac{1+z}{1+z_{eff}}\right)^{\alpha_{1}}
\left(\frac{1+z}{1+z_{eff}}\right)^{\alpha_{2}}
\left(\frac{G(z)}{G(z_{ref})}\right)^{2}
\xi_{Pk}(z_{ref})
```

londumas:
Just as a test, here is the value of `log10(|bias_LYA|)` versus `redshift` for the cross-correlation assuming no evolution of bias_QSO and a `2.9` evolution of `bia_LYA`. Everything is perfect.

capture d ecran de 2018-03-19 13-28-44

londumas:
@ngbusca, ready to be merged.

ngbusca:
@londumas I don't like so much that one has to pass zref to the model functions, because it's a fixed number that doesn't change... but I don't see how else it can be done without more complication. But I really don't like passing zeff. Can you go back to using partial to define the z_evol functions with a fixed zeff? Then you only pass zref to the models.

londumas:
@ngbusca, I can. The issue is that I then need to pass the `zeff` parameter to the `data.data` constructor. Does such a line annoys you: `data.data(parse_data(os.path.expandvars(d)),zeff)` ?
It isn't very beautiful. I can try to pass it trough an other way.

londumas:
@ngbusca, I have another way. I can create a method that acts on the `zevol` and `growth_function`.
Let me do ca commit and you tell me what you think.

ngbusca:
@londumas I think it makes sense to add zeff as an attribute of data. You will need for the output file as well, won't you?

londumas:
@ngbusca, Here is another way to do it.
What do you think of it?

ngbusca:
The thing is that you still have to keep track of zeff in many many places while if you partial it in the beginning it's just only once. This is not minor, since zeff and zref are very similar names, so it's very mistake prone, and since it would cause only minor differences it would be very hard to catch.

londumas:
@ngbusca, did you look at the latest commit? I do `partial` it

londumas:
@ngbusca, sorry, partial them, both zref and zeff

mblomqvist:
I would prefer to save the effective redshift as 'zeff' instead of 'redshift', i.e. g.attrs['zeff'] = self.zeff

londumas:
@mblomqvist, done. @ngbusca, do you want me to change other things?

londumas:
Added fixing issue https://github.com/igmhub/picca/issues/375.
@ngbusca, Could you review the code for setting the parameters `Om` and `OL`?
I am not sure to understand how cached and uncached function works

londumas:
This is one solution, the other is to add the `Om=Om, OL=OL` parameters to `growth_factor_no_de`.
As you prefer @ngbusca.
```
        if dic_init['model']['growth function'] in ['growth_factor_de','cached_growth_factor_de']:
            self.growth_function = partial(getattr(xi, dic_init['model']['growth function']),zref=zref, Om=Om, OL=OL)
        else:
            self.growth_function = partial(getattr(xi, dic_init['model']['growth function']),zref=zref)
```

ngbusca:
@londumas note that `growth_factor_de` is already cached. `cached_growth_factor_de` should never be specified by the user. I prefer the solution with partial...

londumas:
@ngbusca, is that ok now?

py/picca/fitter2/parser.py
```diff
@@ -75,15 +82,17 @@ def parse_data(filename):
...
 
     dic_init['cuts'] = {}
     for item, value in cp.items('cuts'):
         dic_init['cuts'][item] = float(value)
 
     dic_init['model'] = {}
+    dic_init['model']['zeff'] = zeff
+    dic_init['model']['zref'] = fiducial['zref']
+    dic_init['model']['Om']   = fiducial['Om']
+    dic_init['model']['OL']   = fiducial['OL']
```
ngbusca: I'm surprised that coadacy doesn't complaint about these extra spaces. Could you please remove the double spaces here?

py/picca/fitter2/xi.py
```diff
@@ -142,21 +142,18 @@ def wrapper(*args, **kwargs):
 def growth_factor_de(*args, **kwargs):
     return cached_growth_factor_de(*args, **kwargs)
 
-def cached_growth_factor_de(z, zref = None, **kwargs):
+def cached_growth_factor_de(z, zref = None, Om = None, OL = None, **kwargs):
```
ngbusca: Same here. There shouldn't be spaces around the `=` in function parameters

py/picca/fitter2/parser.py
londumas: done

py/picca/fitter2/xi.py
londumas: done

londumas:
@ngbusca, just applied your request.

ngbusca:
@londumas there are conflicts in this branch, can you take a look?

londumas:
@ngbusca, corrected.

PR #373, opened 2018-03-19
Store cosmological parameters in fiducial P(k) file

mblomqvist:
The fiducial power spectrum fits files now store the basic cosmological parameters Om, Ok, OL, w, H0 and the derived parameters sigma8, zdrag, rdrag from CAMB in the header.


ngbusca:
I would prefer that the `pk2fits` would take the `camb` `.ini` as an input, instead of forcing the user to write all the parameters by hand (which might lead to mistakes). What do you think @mblomqvist ?

mblomqvist:
@ngbusca I thought about that but 1) the derived parameters are anyway not listed in the CAMB ini file; 2) users of CAMB for python don't use the ini file and parameters have different names.

mblomqvist:
@ngbusca I can add reading the CAMB ini file as an extra option to get Om, Ok, OL, w and H0, if you want. But specification by hand should also be allowed. What do you think?

ngbusca:
@mblomqvist thanks for the clarification, I didn't think about the derived parameters. Since we can't avoid passing them, I guess we could just leave it as is...

PR #374, opened 2018-03-19
Cut according to center of grid

londumas:
fixes issue https://github.com/igmhub/picca/issues/328.
Cuts according to the center of the grid and not according to the mean
(rp,rt) of the bin


ngbusca:
@londumas in #328 I thought we agreed this wasn't necessary. Can you provide an example where you can't reproduce a bin-center cut with the current master?

Issue #376, opened 2018-03-26
Binning effect  


vserret:
For the fit of the cf, the G(k) term of equation (21) (Bautista, 2017) is composed by the product of square of sinc but the Fourier transform of the rectangle function is just sinc. 
Should we erase the squares ?

ngbusca:
I agree with @vserret . @londumas @mblomqvist any objections?

londumas:
Do you mean that the last paragraph of page 11/22 of Bautista et al (https://arxiv.org/pdf/1702.00176.pdf) is wrong?
Could you give the math and the relevant assumptions?

vserret:
The Fourier transform of the rectangle function is sinc and since we are binning on the cf and not on the deltas, so I think we should not have the squares.

mblomqvist:
OK, that sounds reasonable. Was there a reason for using the square in previous analyses? Also, I noticed that fitter1 divided the argument with pi - why?

mblomqvist:
I might have answered my own question: the division by pi is probably because we used the scipy version of sinc.

ngbusca:
@mblomqvist The square is a heritage of the initial implementation which was relative to a smoothing of the delta field. I guess the transition to smoothing the correlation function was done without enough thought. The `pi` has to do with the definition of the `sinc` function of scipy.

mblomqvist:
@ngbusca thanks, yes I recall that the smoothing was initially for the delta field binning. Then it's clear that we should adopt @vserret's suggestion.

ngbusca:
Reworking the full expression, it's actually a bit more complicated than just removing the square. It seems to me that the transverse smoothing effect on `P(k)` can be treated exactly. I'll assign this issue to myself.

ngbusca:
![20180327_123440](https://user-images.githubusercontent.com/4727509/37962483-860e0ea4-31bb-11e8-9faa-d8de42d4424c.jpg)

ngbusca:
Here's my calculation. I'm a bit puzzled that the transverse `sinc` is exact, but I don't see the error...

ngbusca:
In the expressions above there are missing factors of `exp(i k_p r_p)` and `exp(i k_t r_t)`

PR #378, opened 2018-03-27
removed squares from sinc

ngbusca:
This branch fixes issue #376 reported by @vserret (thanks!). I've only removed the squares form the sinc functions as originally suggested by @vserret . The motivation can be found in the photo of my blackboard posted on the issue. It would be good if other people double-check the math.


londumas:
@ngbusca, you have to reproduce the h5py file.

londumas:
Could we imagine a simple test of this ? It should be simple.
Let's fill the auto-correlation with the power-spectrum where beta = 0 and bias = 1, on a 4x4 grid and fit it with picca. Then do the same with a grid 10x10 and 1x1. what do you think ?

ngbusca:
@londumas can you remind me how to regenerate the test reference files?

londumas:
@ngbusca, in `test/test_cor.py` change the boolean and the value of on attribute:
```
 self._test = False
self._branchFiles = './new_data/'
```
create the './new_data/' directory, run the test and then copy the new file to `data/`.

Issue #377, opened 2018-03-27
typo in tracer type


mblomqvist:
A tracer type can be "continuus" in the code but it should say "continuous".

ngbusca:
I agree. I also don't like "continuous", it's certainly not!

PR #379, opened 2018-03-27
correct typo in tracer type

londumas:
Correct typo in tracer type.
Fix issue https://github.com/igmhub/picca/issues/377


mblomqvist:
Looks good, but before approving should we brainstorm another name as requested by @ngbusca?

mblomqvist:
I have no principal issue with "continuous" (see description in e.g. Slosar et al 2011), but perhaps "forest" would be good?

ngbusca:
ideally, it would be something in line of whatever the tracer-type is for galaxies... I like forest, but it might be a bit restrictive in the sense that we might want to use a portion of the spectrum which is not a forest of any king. Let's keep `continuous` for now and see if we can come up with a better pair.

PR #384, opened 2018-03-28
replace chararray by char

londumas:
As reported by @Cyeche, there can a crash with using `chararray`. `char` works on all platforms.
The reason seems to be: https://docs.scipy.org/doc/numpy/reference/generated/numpy.chararray.html
@Cyeche, can you do this simple review?


Cyeche:
That's fine

PR #385, opened 2018-03-29
Isotropic fit in fitter2

mblomqvist:
This is a new implementation of the isotropic fit (aiso,1+epsilon), following some suggestions by @ngbusca. It has been chosen to make it mandatory to specify the type of cosmo fit in chi2.ini using

[cosmo-fit type]
cosmo fit func = ap_at

or 

[cosmo-fit type]
cosmo fit func = aiso_epsilon


mblomqvist:
Codacy is complaining that utils.cosmo_fit_func isn't callable. Any ideas for how to improve it?

ngbusca:
@mblomqvist I wonder if we actually need to initialize the `cosmo_fit_func attribute`. I.e., it might just work if we simply delete l.14 of utils.py

ngbusca:
by the way, could you post the results of a fit with master vs. your branch? In particular, do you get consistent results for ap/at and aiso/epsilon?

ngbusca:
Sorry, actually the test with master is already done in the unittests, so no need to do it here. Still, I'm curious as to whether the results ap/at and aiso/epsilon are consistent.

londumas:
@mblomqvist, I think the Codacy warning is because of this line `cosmo_fit_func = None`.
However I don't know how to make this work.

ngbusca:
@londumas indeed, as I said, I think it's enough to remove this line.

mblomqvist:
Thanks for your feedback. Removing cosmo_fit_func = None from utils.py still doesn't satisfy Codacy.

ngbusca:
After a bit of googling, I've found many packages for which codacy reports the same problem. I think we can ignore it for now.

mblomqvist:
OK, in that case it is ready to be merged. I modified the code according to @londumas' requests too.

ngbusca:
Actually, could you check that your ap,at values are exactly the same as for master?

mblomqvist:
Same values on the master branch
ap [ 1.03019664  0.03339192]
at [ 0.97928089  0.04146826]

PR #386, opened 2018-03-29
get the input nside in desi

londumas:
get the input nside in desi from reading the `*` in `in_dir+"spectra-*/")`

### DIFF
--- a/py/picca/io.py
+++ b/py/picca/io.py
@@ -496,23 +496,17 @@ def read_from_spplate(in_dir, thid, ra, dec, zqso, plate, mjd, fid, order, log=N
 
 def read_from_desi(nside,in_dir,thid,ra,dec,zqso,plate,mjd,fid,order):
 
-    tmp_in_dir = glob.glob(in_dir+"spectra-*/")
-    if len(tmp_in_dir)!=1:
-        sys.stderr.write("Can't find out input nside: {}\n".format(in_dir+"spectra-*/"))
-        sys.exit()
-    else:
-        tmp_in_dir = tmp_in_dir[0]
-        in_nside = int(tmp_in_dir[len(in_dir+"spectra-"):-1])
-    nest     = True
+    in_nside = int(in_dir.split('spectra-')[-1].replace('/',''))
+    nest = True
     data = {}
-    ndata=0
+    ndata = 0
 
     ztable = {t:z for t,z in zip(thid,zqso)}
     in_pixs = healpy.ang2pix(in_nside, sp.pi/2.-dec, ra,nest=nest)
     fi = sp.unique(in_pixs)
 
     for i,f in enumerate(fi):
-        path = in_dir + "spectra-"+str(in_nside)+"/"+str(int(f/100))+"/"+str(f)+"/spectra-"+str(in_nside)+"-"+str(f)+".fits"
+        path = in_dir+"/"+str(int(f/100))+"/"+str(f)+"/spectra-"+str(in_nside)+"-"+str(f)+".fits"
 
         sys.stderr.write("\rread {} of {}. ndata: {}".format(i,len(fi),ndata))
         try:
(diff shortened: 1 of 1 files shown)


ngbusca:
I don't understand why we need to know `nside`, shouldn't `in-dir` point to the full path including `[...]/spectra-64` ? Why not simply remove all references to `nside`?

ngbusca:
Note also that `read_from_desi` has a different signature as the other `read_from` functions due to the extra `nside` parameter, another reason to just remove `nside`

Cyeche:
Right!

ngbusca:
@londumas sorry for the confusion,I meant the `in_nside` variable. It can be removed if the user sets `in-dir` to the full path including the `spectra-64` part. The `nside` is harcoded to 8 for `read_from_desi` and I think it should be left as is for now (although we can discuss how to improve that as well).

londumas:
@ngbusca, sory for the confusion again. The parameter `nside` can be left has it is.
However the parameter `in_nside` has to read from either the `in-dir` file or from the `in-dir/spectra-*/` file. It is needed to get the path to each files

londumas:
The current version woks well.
If you prefer I change so that the user has to give the path to the `in-dir/spectra-*/` but the structure will still have to be `in-dir/spectra-*/` not something like `path_not_finishing_by_spectra-*/`, because we will read `in_nside` from the path.

ngbusca:
Ok, I understand, so the problem is that given a `TARGETID` and its `RA, DEC` it's impossible to know in which `spectra-` file the spectrum is, unless we know the `nside`, which is only available in the full path of the spectra files, which I don't think it's the desired solution. We should ask the desi team to address this. I agree with your solution, I do still have some comments that I will make directly on the code.

ngbusca:
@londumas actually, if the full path doesn't finish by `spectra-*`, the solution you propose also doesn't work. Your solution relies on `spectra-*` being the only directory of this form, but it will fail if `in-dir` contains two directories, say `spectra-64` and `spectra-128`. I would still require the user to set `in-dir` to the full path including the `spectra-*` part.

py/picca/io.py
```diff
@@ -496,7 +496,13 @@ def read_from_spplate(in_dir, thid, ra, dec, zqso, plate, mjd, fid, order, log=N
...
 def read_from_desi(nside,in_dir,thid,ra,dec,zqso,plate,mjd,fid,order):
 
-    in_nside = 64
+    tmp_in_dir = glob.glob(in_dir+"spectra-*/")
+    if len(tmp_in_dir)!=1:
+        sys.stderr.write("Can't find out input nside: {}\n".format(in_dir+"spectra-*/"))
+        sys.exit()
+    else:
+        tmp_in_dir = tmp_in_dir[0]
+        in_nside = int(tmp_in_dir[len(in_dir+"spectra-"):-1])
```
ngbusca: this is fragile since the user could input an arbitrary number of `/` at the end of `in-dir` which might be removed by `glob`. You could simply do:

    in_nside = int(tmp_in_dir.replace('/','')[-2:])

py/picca/io.py
```diff
@@ -496,7 +496,13 @@ def read_from_spplate(in_dir, thid, ra, dec, zqso, plate, mjd, fid, order, log=N
...
     nest     = True
     data = {}
```
ngbusca: please remove the extra spaces around `=` to leave just one on each side

py/picca/io.py
```diff
@@ -496,7 +496,13 @@ def read_from_spplate(in_dir, thid, ra, dec, zqso, plate, mjd, fid, order, log=N
...
     ndata=0
```
ngbusca: please add one space around the `=` on each side

py/picca/io.py
```diff
@@ -496,7 +496,13 @@ def read_from_spplate(in_dir, thid, ra, dec, zqso, plate, mjd, fid, order, log=N
...
 
-    in_nside = 64
+    tmp_in_dir = glob.glob(in_dir+"spectra-*/")
+    if len(tmp_in_dir)!=1:
+        sys.stderr.write("Can't find out input nside: {}\n".format(in_dir+"spectra-*/"))
+        sys.exit()
+    else:
+        tmp_in_dir = tmp_in_dir[0]
+        in_nside = int(tmp_in_dir[len(in_dir+"spectra-"):-1])
     nest     = True
```
ngbusca: same here

py/picca/io.py
ngbusca: actually, in all the lines below, please just use one space around the `=`

py/picca/io.py
```diff
@@ -496,7 +496,13 @@ def read_from_spplate(in_dir, thid, ra, dec, zqso, plate, mjd, fid, order, log=N
 
 def read_from_desi(nside,in_dir,thid,ra,dec,zqso,plate,mjd,fid,order):
 
-    in_nside = 64
+    tmp_in_dir = glob.glob(in_dir+"spectra-*/")
+    if len(tmp_in_dir)!=1:
```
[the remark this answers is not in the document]
ngbusca: actually, if `in-dir` already contains the `spectra-*`, you can ignore this comment

ngbusca:
>@ngbusca, desi simulated data have nside=64 and this will not move, the data will have the same nside. Or so I have been told, that is why I had hard-coded it.

This is already not the case with the current sims!

PR #387, opened 2018-03-29
move condition on pairs in xcf

londumas:
This PR moves the condition on redshift of pairs to another line.
The reason is that it is very time consuming where it used to be.
It is way faster now.


ngbusca:
what is the problem with `nside=16` ?

londumas:
@ngbusca, the issue was that some bin files had `nside = 8`. This takes a very long time when computing `fill_neighs`. I simply set `nside = 16` by default. Also I moved a condition in `fill_neighs` to be executed only on the selection of pairs, not on every pairs.

londumas:
@ngbusca, the issue was that some bin files had `nside = 8`. This takes a very long time when computing `fill_neighs`. I simply set `nside = 16` by default. Also I moved a condition in `fill_neighs` to be executed only on the selection of pairs, not on every pairs.

PR #389, opened 2018-03-30
Desi pk1 d

Cyeche:
Make Pk1D analysis compatible with DESI mocks


py/picca/io.py
```diff
@@ -545,21 +548,39 @@ def read_from_desi(nside,in_dir,thid,ra,dec,zqso,plate,mjd,fid,order):
...
+            d  = forest(b_ll,fl,iv,t,ra[wt][0],de[wt][0],ztable[t],exp[wt][0],night[wt][0],fib[wt][0],order,diff,reso_in_km_per_s)
             ### R
             iv = r_iv[wt]
             fl = (iv*r_fl[wt]).sum(axis=0)
             iv = iv.sum(axis=0)
             w = iv>0
             fl[w]/=iv[w]
-            d += forest(r_ll,fl,iv,t,ra[wt][0],de[wt][0],ztable[t],exp[wt][0],night[wt][0],fib[wt][0],order)
-            ### Z
+            reso_sum = r_reso[wt].sum(axis=0)
```
ngbusca: Is this doing what you want? Shouldn't you first reconstruct the 2D resolution matrix?

Cyeche: So far the 2D resolution matrix does not really exist in the simulations. What we have is a gaussian PSF. But you are right I'm going to change the code I will introduce a function 
spectral_resolution_desi( reso_sum,b_ll) 
This will be allow us to change the code once a "true" 2D resolution matrix will be available and anayway the code will be more elegant for sure!

Cyeche:
Nicolas,

The question that you raise is important.  With the FFT approach we cannot use the full 2D resolution matrix but we can use a PSF which is not gaussian. All the computation can be done in the function spectra_resolution_desi with the current implementation. 

If we want tu use the full resolution matrix we have to develop a likelihood approach that we have already developped for our firs paper but which is not implemented in the picca code yet. 

The two  following plots show that it works for the current version of the PSF in mocks up to k = 0.025 (km/s)^-1 . In BOSS it was to go to so small scales ;-).  We will see with a more realsitic resolution function


pk1d_desi
pk1d_desi_highk

py/picca/prep_Pk1D.py
```diff
@@ -57,3 +57,16 @@ def spectral_resolution(wdisp) :
     reso = wdisp*constants.speed_light/1000.*1.0e-4*sp.log(10.)
 
     return reso
+
+def spectral_resolution_desi(reso_matrix, ll) :
+
+    dll = (ll[-1]-ll[0])/float(len(ll)-1)
+    reso= sp.clip(reso_matrix,1.0e-6,1.0e6)
+    rms_in_pixel = (sp.sqrt(1.0/2.0/sp.log(reso[len(reso)/2][:]/reso[len(reso)/2-1][:]))
+                    + sp.sqrt(4.0/2.0/sp.log(reso[len(reso)/2][:]/reso[len(reso)/2-2][:])))/2.0
```
ngbusca: @Cyeche please mind the integer division in python3:

    len(reso)/2 => len(reso)//2

this will work in python 2 as well. Can you substitute `1.0/2.0 => 0.5` and `4.0/2.0 => 2`

ngbusca: by the way, is it easy to understand where the formula for `rms_in_pixel` comes from? What I would have done is to reconstruct the 2D resolution matrix from the reso array (which contains a few main diagonals) and calculate the rms of lambda for each row using the resolution as a weight. Is this what this formula does?

py/picca/io.py
```diff
@@ -545,21 +549,30 @@ def read_from_desi(nside,in_dir,thid,ra,dec,zqso,plate,mjd,fid,order):
...
+            ### Z            
             iv = z_iv[wt]
             fl = (iv*z_fl[wt]).sum(axis=0)
             iv = iv.sum(axis=0)
             w = iv>0
             fl[w]/=iv[w]
-            d += forest(z_ll,fl,iv,t,ra[wt][0],de[wt][0],ztable[t],exp[wt][0],night[wt][0],fib[wt][0],order)
+            reso_sum = z_reso[wt].sum(axis=0)
+            reso_in_km_per_s=spectral_resolution_desi(reso_sum,z_ll)
+            diff = z_ll*0.0
```
londumas: would a `sp.zeros_like(z_ll)` also work

py/picca/io.py
```diff
@@ -545,21 +549,30 @@ def read_from_desi(nside,in_dir,thid,ra,dec,zqso,plate,mjd,fid,order):
...
             iv = r_iv[wt]
             fl = (iv*r_fl[wt]).sum(axis=0)
             iv = iv.sum(axis=0)
             w = iv>0
             fl[w]/=iv[w]
-            d += forest(r_ll,fl,iv,t,ra[wt][0],de[wt][0],ztable[t],exp[wt][0],night[wt][0],fib[wt][0],order)
-            ### Z
+            reso_sum = r_reso[wt].sum(axis=0)
+            reso_in_km_per_s=spectral_resolution_desi(reso_sum,r_ll)
+            diff = r_ll*0.0
```
londumas: would a `sp.zeros_like(r_ll)` also work

py/picca/io.py
```diff
@@ -545,21 +549,30 @@ def read_from_desi(nside,in_dir,thid,ra,dec,zqso,plate,mjd,fid,order):
             iv = iv.sum(axis=0)
             w = iv>0
             fl[w]/=iv[w]
-            d = forest(b_ll,fl,iv,t,ra[wt][0],de[wt][0],ztable[t],exp[wt][0],night[wt][0],fib[wt][0],order)
+            reso_sum = b_reso[wt].sum(axis=0)           
+            reso_in_km_per_s=spectral_resolution_desi(reso_sum,b_ll)
+            diff = b_ll*0.0
```
londumas: would a `sp.zeros_like(b_ll)` also work

py/picca/data.py
```diff
@@ -368,14 +369,14 @@ def from_ascii(cls,line):
         m_z = float(a[6])
         m_SNR = float(a[7])
         m_reso = float(a[8])
+        dll = float(a[9])
+        
+        nbpixel = int(a[10])
+        de = sp.array([float(a[11+i]) for i in range(nbpixel)])
```
londumas: `de = sp.array([a[11+i] for i in range(nbpixel)]).astype(float)` is more elegant and should be faster

bin/do_deltas.py
```diff
@@ -365,9 +365,11 @@ def cont_fit(data):
             out_ascii = open(args.out_dir+"/delta-{}".format(p)+".txt",'w')
             for d in deltas[p]:
                 nbpixel = len(d.de)
+                dll = d.dll 
+                if (args.mode=='desi') : dll = (d.ll[-1]-d.ll[0])/float(len(d.ll)-1)
```
londumas: why do you have to do that? Can you explain to me this line?

Cyeche:
> by the way, is it easy to understand where the formula for rms_in_pixel comes from? What I would have done is to reconstruct the 2D resolution matrix from the reso array (which contains a few main diagonals) and calculate the rms of lambda for each row using the resolution as a weight. Is this what this formula does?


Actually, I just compute the sigma of a gaussian centered à N/2., N is an odd number. N is the dimension of the resolution function. This cannot work for the real resolution but so far the PSF is gaussian in the mocks.

Cyeche:
> 
> 
> @londumas commented on this pull request.
> 
> In bin/do_deltas.py <https://github.com/igmhub/picca/pull/389#discussion_r178307417>:
> 
> > @@ -395,7 +397,9 @@ def cont_fit(data):
>                      hd["MEANZ"]=d.mean_z
>                      hd["MEANRESO"]=d.mean_reso
>                      hd["MEANSNR"]=d.mean_SNR
> -                    hd["DLL"]=d.dll
> +                    dll = d.dll
> +                    if (args.mode=='desi') : dll = (d.ll[-1]-d.ll[0])/float(len(d.ll)-1)
> why do you have to do that? Can you explain to me this line?
> 
> 

Two reasons :
1) dll is hard-coded in do_deltas to 1.0e-4 and its is not the same in desi
2) desi scale is not logarithmic but linear, it’s a way to by-pass it for our FFT approach. We will have the logarithmic scale soon.

For the rest, it is a matter of taste. I have no preference.

londumas:
In do_deltas, dll is not hardcoded. It is given by `args.rebin*1e-4`. Currently rebin is an integer however.

py/picca/io.py
ngbusca: @londumas why would you call a function for such a simple calculation? But just for the sake of it, I've timed three options:

    1) diff = z_ll*0
    2) diff = sp.zeros_like(z_ll)
    3) diff = sp.zeros(z_ll.shape)

 3) is fastest, 1) is about twice slower and 2) is the worst, about 4 times slower than 1). I would leave it as is, but up to @Cyeche .

I guess `numpy` provides `zeros_like` as a useful tool when the array can't be multiplied by zero or whether or not it's an array can change at runtime (so `.shape` might not work), and it's slower since it has to figure out what the argument is.

py/picca/data.py
ngbusca: actually the more compact de = a[11:11+nbpixel].astype(float) would completely avoid the for loop and instantiating a list.

ngbusca:
@Cyeche thanks! can you take a look at the codacy recommendations?

londumas:
@Cyeche, I don't really like this `if (args.mode=='desi') : dll = (d.ll[-1]-d.ll[0])/float(len(d.ll)-1)` because as far as I understood this will change the definition of `dll` for the 3d studies also.

Cyeche:
Done

PR #390, opened 2018-04-02
Add Poisson noise for co

londumas:
Add the Poisson error as default when exporting the object auto-correlation


londumas:
The Poisson error for the Landy-Saley estimator is set to be:
image

ngbusca:
I don't think this is the right equation (although it might be an ok approximation). Why not just use equation (48) from Landy-Szalay's paper?

ngbusca:
Landy-Szalay's formula is:

    error = DD/RR^1.5

londumas:
In the paper, I find var = (DD/RR)^2(p). Where do you find: `error = DD/RR^1.5`?
Paper: http://adsabs.harvard.edu/abs/1993ApJ...412...64L

ngbusca:
@londumas from eqs. (48), (43) and (11) (note that eq. (11) is an equation for <RR> not <DD>)

londumas:
@ngbusca, I don' see it.

ngbusca:
Do you agree that from equation (48) you have:

    std(landy-szalay) = d/sqrt(p)

?

bin/export_co.py
```diff
@@ -110,7 +111,10 @@
         co = hh[1]['CO'][:]
         hh.close()
     else:
-        co = sp.eye(dd.size)
+        w = rr>0.
+        co = sp.zeros_like(dd)
```
ngbusca: `zero_like` is much less efficient than `zero(dd.shape)` and `dd*0`, please consider one of the latter options

bin/export_co.py
```diff
@@ -110,7 +111,10 @@
...
+        co[w] = dd[w]*coefDD/(coefDD*rr[w])**2
+        co = sp.diag(co)
 
     ### Save
     h = fitsio.FITS(args.out,'rw',clobber=True)
```
ngbusca: I know this is not part of this PR, but it might be useful to keep DD, DR and RR in the output file. Note that the fitter won't be bothered by the extra columns or extra HDUs...

bin/export_co.py
londumas: done

londumas:
@ngbusca, do you mean that in equation 48 (`<d>(p)`) the parenthesis have to be understood as "multiply" and not "function of"? Than it makes sense.

ngbusca:
@londumas yes...

londumas:
Corrected the error Poisson. The chi2 is still not very good for the mocks, this has to be better understood with the update: `chi2/(nbBin-nbPar) = 2677.86 / (1252-7)`
capture d ecran de 2018-04-03 11-12-09

Issue #396, opened 2018-04-05
Add check for module versions to unittest


ngbusca:
It would be great to add checks for the module versions to the unittest and issue a WARNING message (not an error) if they are different from those in the `requirements.txt` file. @londumas could you implement that?

londumas:
@ngbusca, what is `future` in `../../../requirements.txt`

PR #397, opened 2018-04-05
test required version of module against local version

londumas:
test required version of module against local version.
Fix issue https://github.com/igmhub/picca/issues/396.

@ngbusca, what should we do for:
- newer version? I think we should still keep the warning
- for `future`? What is the purpose of this in the `requirements.txt`
- if the module is not installed?

### DIFF
--- a/py/picca/test/test_cor.py
+++ b/py/picca/test/test_cor.py
@@ -45,6 +45,8 @@ class TestCor(unittest.TestCase):
 
     def test_cor(self):
 
+        self.test_requirements()
+
         numpy.random.seed(42)
 
         print("\n")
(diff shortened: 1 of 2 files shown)


ngbusca:
Yes, print a warning if the module is newer as well. Future should be 0.16.0 (it's the module that makes `from future import ...` work). If it's not installed then print a warning as well.

ngbusca:
please update the `requirements.txt` to set future=0.16.0

londumas:
@ngbusca, ready to be merged.
The codacy message is because I use `try/except`. You can ignore it.

ngbusca:
@londumas can you specify the exception in the try-except block?

londumas:
@ngbusca, ready to be merged.

PR #403, opened 2018-04-18
enables the user to use the continuum of the mock instead of the compu…

solenechabanier:
Modifications of 1D mock to use the continuum of the mock instead of the computed one


londumas:
@solenechabanier, I am a little bit confused.
In the mocks you have already the delta field so why do you want to get it from picca?
Could you give us more physical context?

ngbusca:
@solenechabanier do you really need to have the 0.8 hardcoded? Wouldn't it be better to add the mean absorption to your mock files in case it changes in the future? e.g., 0.8 doesn't include a redshift evolution.

solenechabanier:
@londumas No worries, sorry if I am not very clear. We have the psf in the mocks but we need to compute the resolution, and we also need to apply the masking functions for the dla or the sky lines and everything is already implemented in do_deltas.
@ngbusca Yes you're right, I'll see with JM if it's ok

solenechabanier:
@ngbusca I put  the mean absorbtion in the io file for now, but we plan to add it in the future mocks

ngbusca:
@Cyeche @solenechabanier thanks for fixing the tabs. Using tabs is quite fragile in python (see https://www.python.org/dev/peps/pep-0008/#tabs-or-spaces) so please configure your editors such that `tab` prints four spaces.

@londumas do you have any more questions? 

Is it ready to be merged?

solenechabanier:
@ngbusca it's ready to be merged, I updated with the previous commits from the master branch, I am just waiting for the checks now

py/picca/io.py
```diff
@@ -274,8 +276,10 @@ def read_from_mock_1D(in_dir,thid,ra,dec,zqso,plate,mjd,fid,order,mode,log=None)
     except IOError:
         log.write("error reading {}\n".format(fin))
 
+    ih = 0
     for t,r,d,z,p,m,f in zip(thid,ra,dec,zqso,plate,mjd,fid):
-        h = hdu[t]
+        h = hdu[index[0][ih]+1]
```
ngbusca: @solenechabanier why can't you use `t` as an index here? Did something change in the simulated data? The introduction of this `index` variable seems a bit strange...

solenechabanier: We introduced it because we needed two thingID, one for the position of the spectrum in the fits file and the other to match with the DLAs from the catalog

ngbusca: Why are they different?

Cyeche:
Basically there is no thid in 1D mocks and originally we used the hdu number as thid.  The plan in a future version is to add the true thid in the 1D mocks to match all the information thanks to the thid.

ngbusca:
@Cyeche I still don't understand, you still have a unique thid that identifies each spectrum, don't you? If not, it shouldn't be up to `picca` to handle data that lacks required info.

But if you absolutely need this then please move the matching between the thid and the hdu to the function `read_from_mock_1D` since it's a bit fragile to pass two indexes that refer to the same thing (`thing_id` and `index` in the current implementation). You can efficiently do this by, e.g. building a dictionary thind_id=>hdu:

    thid_to_hdu = {h.read_header()['THING_ID']:h for h in hdus}

solenechabanier:
@ngbusca But how could we do the matching inside the read_from_mock_1D function without a second information ? We could pass a dictionary or the w list as parameter but then it would do basically the same thing

ngbusca:
@solenechabanier you actually assume that `THING_ID` is the header number (minus one) no? Then why pass a second index?

Cyeche:
The point is that we don’t have the true thing_id, in the header of the spectra and we need this true thind_id to simulate the effect of the  DLA….

Cyeche:
We want to have the DLA in the same position in the forest, for the same z…. For sure we need to have the thing_id in the header, but it is not the case yet.

ngbusca:
Ok, I understand. I've just discussed about it with Jean-Marc, he says that it would take him 1hr to redo the sims adding the `THING_ID` to the header. Otherwise you can run a simple script to the current sim files to add it to the current files. I think it's much better to keep the code clean than to do accrobacies to make it up for a transitory state...

Cyeche:
Anyway it is not optimal to open all the hdu to find which one we want to read… With this list of hdu, the reading will be faster, we open the hdu that we want.

ngbusca:
@Cyeche I agree, that's why it's best to have the thing_ID. Then you directly open the hdu indexed by the thing_ID.

londumas:
@solenechabanier, you have merged with an outdated version of master. Could you update your master then merge again?

py/picca/io.py
```diff
@@ -7,6 +7,7 @@
 import time
 import os.path
 
+import constants
```
londumas: this is unused

londumas: and it should have beed `from picca import constants`

solenechabanier: I should have commented sorry, and I 'll modify that, thanks ! We won't actually need it once the psf from the mock_1D is in pixels (instead of angstrom), the modification has not been done yet so I am using the constant package at the line 304 that I commented for the push

ngbusca:
@solenechabanier let me know when it's ready for review!

ngbusca:
@solenechabanier is it ready for a review?

solenechabanier:
@ngbusca Not yet, we are waiting for the last production of mocks to test the branch with the last modifications, I'll let you know as soon as it is done (very soon!)

ngbusca:
@solenechabanier sorry for taking so long to review. It looks good to me, should I merge? I'm double checking since it's been a while.

solenechabanier:
Great !  Thank you !

Issue #287, opened 2018-01-18
z-evol doesn't exist in do_cf1d


londumas:
One of the goal of computing do_cf1d is to use it in Wick.
For this to be valid, the 1d correlation of forests has to have the same weights and data has the 3d correlation of forests.
We thus need to add `--z-ref`, `--z-evol` and `--z-evol2`

ngbusca:
@londumas did we ever address this issue?

londumas:
@ngbusca, no but it is very simple to fix. The only question is: do we plan to use `do_wick` any time soon?

PR #407, opened 2018-04-20
Add z evol cf1d

londumas:
- add --z-ref, --z-evol and --z-evol2 in do_cf1d.py
- read data from io.py
fix issue https://github.com/igmhub/picca/issues/287

### DIFF
--- a/py/picca/cf.py
+++ b/py/picca/cf.py
@@ -2,8 +2,8 @@ import scipy as sp
 import sys
 from healpy import query_disc
 from numba import jit
-from .data import forest
 from scipy import random
+
 from picca import constants
 
 np = None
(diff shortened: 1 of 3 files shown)


py/picca/cf.py
```diff
@@ -425,7 +425,7 @@ def x_forest_cf1d(pix):
             we2 = d2.we
             xi1d[bins] += wde1 * wde2[:,None]
             we1d[bins] += we1*we2[:,None]
-            nb1d[bins] += 1
+            nb1d[bins] += (we1*we2[:,None]>0.).astype(int)
```
ngbusca: We remove masked bins from the forests, don't we? There shouldn't be any pixels with zero weight.

Issue #409, opened 2018-04-23
bug in UV fluctuation model


mblomqvist:
There is a bug in the function pk_uv. The parameter bias_prim is replaced by bias1 when calculating the scale-dependent redshift-space distortion. I can fix it.

ngbusca:
Thanks for reporting this, please go ahead!

PR #412, opened 2018-04-23
UV fluctuation model for cross-correlation

mblomqvist:
Add the UV fluctuation model to object-forest cross-correlation fit in fitter2.


mblomqvist:
Also fixes the bug reported in #409

mblomqvist:
Codacy is complaining about the use of 'assert', but we have many such occurrences in the codes. I assume that we can ignore the complaint.

londumas:
@mblomqvist, yes ignore the complaint for `assert`. Some day we will dive into it.
Thank you for the PR.

mblomqvist:
ready to be merged

PR #416, opened 2018-04-24
correct velo dispersion if both tracers are discrete

londumas:
correct velo dispersion if both tracers are discrete. For example for the auto-correlation of quasars.

### DIFF
--- a/py/picca/fitter2/pk.py
+++ b/py/picca/fitter2/pk.py
@@ -170,19 +170,15 @@ def pk_hcd_cross(k, pk_lin, tracer1, tracer2, **kwargs):
 def pk_velo_gaus(k, pk_lin, tracer1, tracer2, **kwargs):
     assert 'discrete' in [tracer1['type'],tracer2['type']]
     kp = k*muk
-    res = 1.
-    if tracer1['type'] == 'discrete':
-        res *= sp.exp( -0.25*(kp*kwargs['sigma_velo_gauss'])**2)
-    if tracer2['type'] == 'discrete':
-        res *= sp.exp( -0.25*(kp*kwargs['sigma_velo_gauss'])**2)
-    return res
+    if tracer1['type'] == tracer2['type']:
+        return sp.exp( -0.5*(kp*kwargs['sigma_velo_gauss'])**2)
+    else:
+        return sp.exp( -0.25*(kp*kwargs['sigma_velo_gauss'])**2)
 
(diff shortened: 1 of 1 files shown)


py/picca/fitter2/pk.py
```diff
@@ -168,11 +168,21 @@ def pk_hcd_cross(k, pk_lin, tracer1, tracer2, **kwargs):
...
 def pk_velo_lorentz(k, pk_lin, tracer1, tracer2, **kwargs):
-    assert tracer1['type'] == 'discrete' or tracer2['type'] == 'discrete'
+    assert 'discrete' in [tracer1['type'],tracer2['type']]
     kp = k*muk
-    return 1/sp.sqrt(1.+(kp*kwargs['sigma_velo_lorentz'])**2)
+    res = 1.
+    if tracer1['type'] == 'discrete':
+        res *= sp.exp( -0.25*(kp*kwargs['sigma_velo_lorentz'])**2)
+    if tracer2['type'] == 'discrete':
+        res *= sp.exp( -0.25*(kp*kwargs['sigma_velo_lorentz'])**2)
```
ngbusca: it would be much more efficient to do res = res**2 than to recalculate the exponential. You could then remove the first `if` since you already know from the first assert that at least one tracer is `discrete`:

    res *= ...
    if tracer1['type'] == tracer2['type']:
        res *= res

londumas:
@ngbusca, Just applied your requested changes.

py/picca/fitter2/pk.py
```diff
@@ -168,11 +168,17 @@ def pk_hcd_cross(k, pk_lin, tracer1, tracer2, **kwargs):
...
     kp = k*muk
-    return sp.exp( -0.25*(kp*kwargs['sigma_velo_gauss'])**2)
+    if tracer1['type'] == tracer2['type']:
+        return sp.exp( -0.5*(kp*kwargs['sigma_velo_gauss'])**2)
+    else:
+        return sp.exp( -0.25*(kp*kwargs['sigma_velo_gauss'])**2)
 
 def pk_velo_lorentz(k, pk_lin, tracer1, tracer2, **kwargs):
-    assert tracer1['type'] == 'discrete' or tracer2['type'] == 'discrete'
+    assert 'discrete' in [tracer1['type'],tracer2['type']]
```
ngbusca: @londumas why are you changing the lorentzian model to gaussian?

londumas:
@ngbusca, thank you for noticing this. I corrected the bug from copy/paste.

PR #418, opened 2018-04-25
fix some issue with randoms in do_deltas and fitter2

londumas:
Add two initialization of random seeds:
- in `do_deltas` when we use a dla catalog
- in `fitter2` when we use fast mc. For this the user has to give a `seed` in the `chi2.ini`

### DIFF
--- a/py/picca/fitter2/chi2.py
+++ b/py/picca/fitter2/chi2.py
@@ -5,7 +5,6 @@ import iminuit
 import time
 import h5py
 from scipy.linalg import cholesky
-from scipy import random
 
 from . import utils, priors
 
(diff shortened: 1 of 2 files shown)


py/picca/fitter2/chi2.py
```diff
@@ -27,6 +26,7 @@ def __init__(self,dic_init):
             self.verbosity = dic_init['verbosity']
 
         if 'fast mc' in dic_init:
+            self.seedfast_mc = dic_init['fast mc']['seed']
```
ngbusca: I disagree with forcing the user to choose a seed for the `fast mc`, the code should handle it correctly internally

londumas:
@ngbusca, thank you for the review. Just made the corrections: the seed is `0` is not given.

londumas:
ok to merge?

PR #419, opened 2018-05-04
Pk1 d noise

nathalieGit:
New method for diff noise in Pk1D


londumas:
Could you use scipy instead of numpy? It would be nice to have only one of the two in picca. And scipy is the one we have settled on.

nathalieGit:
Hi Helion. I changed numpy to scipy.

londumas:
@nathalieGit, Thank you.

PR #420, opened 2018-05-14
Branch for mpi support

ngbusca:
This branch implements mpi parallelization:

* added mpi parallelization to do_cf.py and do_dmat.py
* tested on the saclay/v2.0 mocks, `do_cf` runs in ~1.5hrs and `do_dmat`  in ~3.5hrs on 10 Haswell nodes at NERSC, with the config below (but note this version contains about 500,000 l.o.s. compared to ~1,000,000 in the london mocks)
* it's not ready to be merged. In particular, the tests will fail since mpi4py is not (and won't be) installed on Travis. I will move the mpi versions to separate scripts
* mpi versions of the other do_X.py are still missing, but this is enough for the current mocks

     #!/bin/bash
    #SBATCH -N 10
    #SBATCH -C haswell
    #SBATCH -q regular
    #SBATCH -J do_cf
    #SBATCH --mail-user=nbusca@lpnhe.in2p3.fr
    #SBATCH --mail-type=ALL
    #SBATCH -t 02:00:00 
    #SBATCH -L project
    #SBATCH -A desi

    #OpenMP settings:
    export OMP_NUM_THREADS=1

    #run the application:
    srun -n 10 -c 64 do_cf.py --in-dir deltas --out cf.fits.gz --nproc 32

    #!/bin/bash
    #SBATCH -N 10
    #SBATCH -C haswell
    #SBATCH -q regular
    #SBATCH -J do_dmat
    #SBATCH --mail-user=nbusca@lpnhe.in2p3.fr
    #SBATCH --mail-type=ALL
    #SBATCH -t 04:00:00
    #SBATCH -L project
    #SBATCH -A desi

    #OpenMP settings:
    export OMP_NUM_THREADS=1

    #run the application:
    srun -n 10 -c 64 do_dmat.py --in-dir deltas --out dmat.fits.gz --nproc 32 --rej 0.95


ngbusca:
The io can also be improved. Right now, each node unnecessarily inputs the full data set

julienguy:
in the current implementation, each rank is reading the data
it should be read only by rank 0 and then broadcast.

julienguy:
I am working on a modified version i'll push in a new branch

julienguy:
I have modified do_cf.py in branch mpi4py2. You can merge back to this branch if ok.
The changes are:
 - only rank 0 reads the data
 - same script can run multiprocessing only with --no-mpi option
 - mpi version does not use multiprocessing.pool (known to be problematic when used together with mpi)
 - checked the following commands give the same result (1st is multiprocessing, second is mpi)
```
time do_cf.py --in-dir delta-test --out correlations/cf-test-multiproc.fits.gz --z-cut-max 2.2 --nproc 32 --no-mpi

time srun -N 1 -n 32 -c 2 do_cf.py --in-dir delta-test --out correlations/cf-test-mpi.fits.gz --z-cut-max 2.2
```
On this example, mpi is slower because there are overheads with srun, but the advantage is that you can run on multiple nodes. I am now trying a full run.

julienguy:
The progress message with a global counter and a lock is of course broken with this version but this is probably a detail.

julienguy:
The full MPI solution does not work for large dataset because of a memory issue: the data is duplicated for each process of a node. A good solution that requires some work is to distribute the data among the processes without duplication of memory. This was implemented by Ted Kisner for redrock.
On the short term, we can stick with the MPI + multiprocessing approach as implemented in this branch.
I am modifying the branch mpi4py2 accordingly ...

ngbusca:
@julienguy thanks, indeed it's much better to have only rank 0 read the data. I'll put that in.

> The full MPI solution does not work for large dataset because of a memory issue: the data is duplicated for each process of a node. A good solution that requires some work is to distribute the data among the processes without duplication of memory.

It works for full DESI no? Or do we need that already?  The complication is that you have to send each rank enough pixels so that it correctly gets all the neighbours. It's doable, but it needs quite a bit of code refactoring.

ngbusca:
@julienguy we shouldn't add mpi4py as a dependency so people can still run in their laptops. So, two solutions:

1) write separate do_mpi_X scripts
2) put if's everywhere so that we only import mpi4py when the --mpi option is used

Both solutions are a bit ugly, but I'm leaning towards 1)... what do you think?

ngbusca:
@julienguy I've included your suggestions to do_cf, do_dmat, do_xcf and do_xdmat:

* added an option `--mpi` to run the mpi version (instead of `--no-mpi` to avoid modifying the travis scripts)

Can you expand on:
> mpi version does not use multiprocessing.pool (known to be problematic when used together with mpi)

your branch also uses a pool. I've googled mpi and multiprocessing and couldn't find people complaining. What do you propose?

ngbusca:
Ok, I found the problem with the broadcast of large data samples. I'm working on a fix...

julienguy:
Crash with current version when running on a small data set because of some empty sample I presume.
(I let you fix this minor bug as you are already working on some other modifications)

```
multiprocessing.pool.RemoteTraceback: 
  File "/global/common/software/desi/cori/desiconda/20180130-1.2.4-spec/conda/lib/python3.6/multiprocessing/pool.py", line 119, in worker
    result = (True, func(*args, **kwds))
  File "/global/common/software/desi/cori/desiconda/20180130-1.2.4-spec/conda/lib/python3.6/multiprocessing/pool.py", line 44, in mapstar
    return list(map(*args))
  File "/global/common/software/desi/users/jguy/igmhub/picca/bin/do_cf.py", line 20, in corr_func
    cf.fill_neighs(p)
  File "/global/common/software/desi/users/jguy/igmhub/picca/py/picca/cf.py", line 46, in fill_neighs
    npix = query_disc(nside,[d1.xcart,d1.ycart,d1.zcart],angmax,inclusive = True)
  File "healpy/src/_query_disc.pyx", line 71, in healpy._query_disc.query_disc (healpy/src/_query_disc.cpp:2137)
TypeError: must be real number, not NoneType
```

ngbusca:
Thanks @julienguy. Can you post the submission command ?

julienguy:
On a cori interactive session:
```time srun -N 1 -n 2 -c 32 do_cf.py --in-dir delta-test --out correlations/cf-test-mpi-and-multiprocessing-bis.fits.gz --z-cut-max 2.2 --mpi```

(and delta-test has just 8 pixel files)

**Comment by julienguy** [comment]
I have a working mpi version of ```do_cf.py``` , pushed  in the branch mpi4py3 (branched from mpi4py, I let you merge if you like it),
 - each rank reads a fraction of the data files
 - then an "allgather" is performed, per chunk of data, in order to avoid a python pickle bug for large data sets

Example, on delta from Saclay full mock ```/project/projectdirs/desi/mocks/lya_forest/saclay/v2.1/quick-0.3```

```
cd /project/projectdirs/desi/users/jguy/lya/saclay-v2.1-0.3/delta
time srun -N 10 -n 10 -c 64 do_cf.py --in-dir delta --out correlations/cf.fits.gz --mpi --nproc 32
```

 - reading the 328 delta files took 127 sec
 - gathering took 47 sec
 - computation of the correlation function took 78 min

I am now trying on 50 nodes in a batch job ...

ngbusca:
@julienguy if it works I like it! I'm making a number of refactors, I'll include your modifications.

ngbusca:
@julienguy for reference, it seems that using `bcast` makes `pool.map` hang, but using `allgather` doesn't:

This runs ok:

    dels = comm.allgather(dels)
    dels = dels[0]

but this gets correctly (as far as I can tell) the objets to all the nodes, but `pool.map` hangs at all ranks but rank=0:

    dels = comm.bcast(dels, root=0)

I'll incorporate all this to all other do_X functions.

londumas:
@ngbusca, could you add also `do_co`? It is ok for O(1M) objects, but quite slow for O(10M), which is the number we need for randoms.
Also most of the Codacy issues are easy to fix: trailing spaces, unused imports, unused variables

ngbusca:
@julienguy I think we've superseded the need of `mpi` for now, do you agree? If yes, I'll close this PR

julienguy:
Well, we can close the PR but please do not delete the branch. I hope I/someone can find some time to work on this later.

ngbusca:
ok, I'll leave it open then

londumas:
Closing the PR for the moment but not the branch.

PR #428, opened 2018-05-21
Apply codacy 4

londumas:
Remove unused variables and correct implementation of checking type

### DIFF
--- a/py/picca/cf.py
+++ b/py/picca/cf.py
@@ -287,7 +287,6 @@ def metal_dmat(pix,abs_igm1="LYA",abs_igm2="SiIII(1207)"):
             z1_abs2 = 10**d1.ll/constants.absorber_IGM[abs_igm2]-1
             r1_abs2 = cosmo.r_comoving(z1_abs2)
             w1 = d1.we
-            l1 = d1.ll
             r = random.rand(len(d1.neighs))
             w=r>rej
             npairs += len(d1.neighs)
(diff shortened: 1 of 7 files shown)


ngbusca:
@londumas can you take a look at the conflicts?

londumas:
@ngbusca, ready to be merged

mblomqvist:
I'm viewing chi2.py. By replacing 'it' by '_' aren't you breaking the line 198
self.fast_mc_data[d.name+'_'+str(it)] = d.da

londumas:
@mblomqvist, thank you. I did not noticed that `it` was used now.
I just corrected this error.

ngbusca:
Ideally @londumas @mblomqvist this should have been picked up by the automatic tests, is it because we don't test the `fast mc` option? they are already pretty long, should we add a test for the `fast mc` ? what do you think?

londumas:
@ngbusca, The unittest are long simply because of the environment setup.
The rest is reasonable. We can add a unittest for `fast_mc` if you want.

mblomqvist:
I'm considering some more options for the fast MC, so maybe the unit test can wait for those to be implemented first

londumas:
ok, works for me

ngbusca:
> @ngbusca, The unittest are long simply because of the environment setup. The rest is reasonable. We can add a unittest for fast_mc if you want.

This is not true, setting up the environment takes ~5min, the rest takes ~15min.

ngbusca:
> I'm considering some more options for the fast MC, so maybe the unit test can wait for those to be implemented first

no problem, let's open an issue so we don't forget.

londumas:
@ngbusca, ready to be merged.

Issue #417, opened 2018-04-25
Random seed is the same across all processes


londumas:
When running functions that requires random numbers, for example `do_dmat.py` the same random seed (`=0`) is distributed across all processes (I don't know how this issue plays a role in `py/picca/Pk1D.py`).
The result is that the line `r = random.rand(len(d1.neighs))` in `dmat()` is going to return the same values across all processes.
This is not a good practice but I am not sure it will create a problem.
One case where it might be a problem is: since forests or objects are not in a random order in their catalogs (for example in DR12Q the qsos are ordered according to their RA and Dec) we could be always running on  the same kind of pairs.

The solution is to pass a random seed to each process, following such a logic.
Note that `pool.map` doesn't support multiple arguments, that is why I use here tuples.

In `do_dmat.py`
```
def calc_dmat(args):
    p = args[0]
    seed = args[1]
    if x_correlation:
        cf.fill_neighs_x_correlation(p)
    else:
        cf.fill_neighs(p)
    sp.random.seed(seed)
    tmp = cf.dmat(p)
    return tmp

sp.random.seed(0)
table_of_seed_for_processes = sp.unique((100000.*sp.random.rand(10*len(cpu_data))).astype(int))
values = sorted(list(cpu_data.values()))
to_send = [ (v,table_of_seed_for_processes[i]) for i, v in enumerate(values) ]
dm = pool.map(f,to_send)
```

What do you think @ngbusca? @Cyeche?

londumas:
fixed.

PR #429, opened 2018-05-21
One seed per proc

londumas:
This PR allows to have a different seed per processors. Currently, the same seed is used across all processors, and thus the same series of pairs are rejected.
This can cause some small issues.
Since all these branches are being refactored for MP, I didn't implement for do_dmat and do_metal_dmat. The goal is to have a trace of a way to correct this small issue and apply it to MP files once they are merged or even before. This code works for Python2 and Python3, but implies to recreate a lot of test/data/ files.
Correct issue https://github.com/igmhub/picca/issues/417


ngbusca:
I don't understand this phrase `Currently, the same seed is used across all processors, and thus the same series of pairs are rejected.` What do you mean?

londumas:
What I mean is simple: in the current master, for `do_xdmat.py` for example, we set the seed to `0` (`random.seed(0)`). Then we split the data set, and send the computation on different processors. Each of these will have the seed set to `0`. The result is that the same random number will be used to reject some of the pairs in the computation in `py/picca/xcf.py`. It is never good to use the same randoms. However, normally each pair forest-qso is unique so it shouldn't produce a big difference.
The only case where it could is if the pairs were sorted by the quasar index, which is the case.
Normally this issue shouldn't have a big impact, but "prevention is better than cure".

ngbusca:
I see, I thought that there was only one random state common to all multipocessing child processes, but it's actually not the case. All child process inherit the same random state, so you indeed get the same sequence. 

What I don't like in your solution is that we need to modify the arguments of functions to pass a seed which is annoying and syntactically ugly. I also don't like that the user would set a seed, it's useless  low level control.

Wouldn't it be much simpler if the random generator is initialized with the first healpixel id? e.g.:

    def calc_dmat(p):
        random.seed(p[0])

londumas:
agreed and agreed. Let me do a PR with your idea.

londumas:
re-implemented in https://github.com/igmhub/picca/pull/461.

PR #430, opened 2018-05-21
get only transmission files in zcat

londumas:
get only transmission files in zcat, send with something like this:
```
import picca.utils
zcat = '/project/projectdirs/desi/mocks/lya_forest/london/v1.1/quick-0.0/zcat.fits'
indir = '/project/projectdirs/desi/mocks/lya_forest/london/v1.1/*/*/transmission-16-*.fits'
outdir = '/global/cscratch1/sd/hdumasde/igmhub/picca/CoLoRe_mocks/v2.1.1/Test_convert_transmission_2/'
picca.utils.desi_convert_transmission_to_delta_files(zcat=zcat,indir=indir,outdir=outdir,nspec=10000)
```

### DIFF
--- a/py/picca/utils.py
+++ b/py/picca/utils.py
@@ -172,16 +172,30 @@ def desi_from_ztarget_to_drq(ztarget,drq,spectype="QSO"):
     out.close()
 
     return
-def desi_convert_transmission_to_delta_files(indir,outdir,lObs_min=3600.,lObs_max=5500.,lRF_min=1040.,lRF_max=1200.,dll=3.e-4,nspec=None):
-    '''
-    Convert desi transmission files to picca delta files
-    indir: path to transmission files directory
-    outdir: path to write delta files directory
-    lObs_min, lObs_max = 3600.,5500.: observed wavelength range in Angstrom
-    lRF_min, lRF_max = 1040.,1200.: Rest frame wavelength range in Angstrom
-    dll: size of the bins in log lambda
-    nspec: number of spectra
-    '''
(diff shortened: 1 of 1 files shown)


ngbusca:
Thanks @londumas ! does it work on both london and saclay mocks?

londumas:
It works only on London mocks (Saclay doesn't have EXTNAME), either we wait 1 or 2 days for the latest Saclay mocks or I change it so that it reads both now. What do you prefer?

ngbusca:
It would be great to have that now so the analysis will be ready for the meeting! But if the code gets too complicated, then don't worry about it.

py/picca/utils.py
```diff
@@ -252,6 +278,8 @@ def desi_convert_transmission_to_delta_files(indir,outdir,lObs_min=3600.,lObs_ma
             deltas[pixnum].append(delta(thid[i],ra[i],dec[i],z[i],thid[i],thid[i],thid[i],cll,civ,None,cfl,1,None,None,None,None,None,None))
         if not nspec is None and sp.sum([ len(deltas[p]) for p in list(deltas.keys())])>=nspec: break
 
+    print('\n')
```
ngbusca: this will print two empty lines, is that what you want?

londumas: I just want to separate the two `sys.stderr.write`. What is the best way?

ngbusca: It's ok, eventually we will use one single function for screen output, probably a flushed `print()`

londumas:
@ngbusca, ready to be merged.
It works for both London and Saclay mocks.
It is at the cost of not beautiful code. As soon as we have the two newest versions,
I will make a simple PR using EXTNAME.

Issue #436, opened 2018-05-23
Should `cos>=1.` be treated as an error?


londumas:
https://github.com/igmhub/picca/blob/beb820697f6160266a6538a49e99eed561dece1e/py/picca/data.py#L36

Currently when we get `cos>=1.`, we set the angle to `theta = 0`.
But this is not totally correct, should we treat `cos>=1.` as an error? Should we add more precision to
`xcart`, `ycart`, `zcart` ?

ngbusca:
@londumas I think we should use something more robust like `atan2` instead of `acos` to avoid this problem.

londumas:
also we can run a more precise calculation for pairs with `cos>=1.`.
At least we should send a `WARNING`.

ngbusca:
at small ddec et dra, the most reliable computation of the space angle is `ang = sqrt(ddec**2 + sin(dec)**2*dra**2)`. We could do that where `ddec` and `dra` are smaller than some threshold value like `1e-5`.

londumas:
Do you mean `ang = sqrt(ddec**2 + cos(dec)**2*dra**2)`? At least that is what you gave me in your email: `ang_small=sp.sqrt((q2.dec-q1.dec)**2+sin(pi/2-q1.dec)**2*(q1.ra-q2.ra)**2)`

ngbusca:
@londumas Yes, indeeed!! Sorry...

PR #437, opened 2018-05-23
Cast to float64

londumas:
This PR cast to float64 `RA`, `DEC` and `Z` when converting desi data or mocks.
It also add a warning when `cosinus<-1` or `cosinus>1`.
It also removes computing the angle between an object and itself.
I am currently testing it, I tell you @ngbusca when it is ready to be merged.

Fix partly issue https://github.com/igmhub/picca/issues/436


londumas:
Here is the not-projected cross-correlation variance after and before this PR for the transmission.
There is still some spikes, but way less important.
We removed the majority of the effect.
The remaining spikes could be linked to the computation of the angle when it is small.
We have to dig into this.
This covariance matrix is not positive definite for all before and now, as it can be understood by the big correlation matrix.

capture d ecran de 2018-05-23 18-27-45

capture d ecran de 2018-05-23 18-27-57

londumas:
The current PR don't fix all of the issue, see attached plot for London mocks. Using the random catalog don't produce these spikes. The random catalog was produced in double precision.
Still investigating, I plan do test with a random shuffle of the mock catalog to see if indeed the issue still comes from the precision.

covar

covar_zoom

londumas:
Looking only at r_perp<12 Mpc/h, we can see this issue using the catalog of quasars, but not when using the randoms. This effect is caused by close pairs of delta-quasar, where the quasar is behind
the quasar hosting the delta: r_perp<4 and r_paral<-50.
I have looked and it doesn't seem to be linked to our calculation of the angle.

mockcat_london_smallscales

randomcat_london_smallscales

londumas:
@ngbusca, ready to be merged

ngbusca:
@londumas I don't see the codacy checks. For some reason sometimes they disappear. Let's wait a bit to see if they come back.

londumas:
@ngbusca, Codacy doesn't tell what is wrong when I click on the link.

PR #442, opened 2018-05-29
Add subsample covar co

londumas:
Add computation of the co-variance matrix with sub-sampling for the auto-correlation and cross-correlation of objects.
In progress...


londumas:
The resulting matrix is very different from the Poisson results:
- covariance is high
- correlation matrix is positive definite but distant bins are correlated

in London mocks, fitting from 40 to 160:
Poisson covariance matrix:
- chi2 = 2022.77 / (1177-6),  p = 0
- ap = 0.970 +/- 0.018
- at = 0.965 +/- 0.015

Sub-sampling covariance matrix: 
- chi2 = 1173.35 / (1177-6),  p = 0.48
- ap = 0.967 +/- 0.026
- at = 0.980 +/- 0.024

variance:

 capture d ecran de 2018-05-29 18-08-18

correlation:

capture d ecran de 2018-05-29 18-08-38

londumas:
The structure of the correlation matrix is very different from what we are used to.
The following plot gives the un-smoothed correlation matrix.

capture d ecran de 2018-05-30 08-52-36

ngbusca:
Interesting. What are the blocks?

londumas:
The "bug" is fixed. It was coming from two aspects:
- we have to run on all pairs and not on half so that all sub-samples are filled with their pairs.
This implies a code twice as slow. I removed the term: `if o2.ra > o1.ra` in `py:picca:co.py:::fill_neighs()`
- when looking at the difference between sub-samples the terms DR and RD are different.
They are not however when summing over `axis-0`. I had to add calculating the therm RD.

The result is:
- a correlation-matrix that is as expected:
- a variance that is approximately the same but higher than the Poisson expectation.
However, please remember that in the variance there is contribution from the number
of random. Here I used only twice as many. More are running.

co_correlationmatrix_subsampling

co_correlation_subsampling

co_variance_poisson_subsampling

londumas:
@ngbusca, ready to be merged.

ngbusca:
@londumas I'm very confused by:

> we have to run on all pairs and not on half so that all sub-samples are filled with their pairs.
This implies a code twice as slow. I removed the term: if o2.ra > o1.ra in py:picca:co.py:::fill_neighs()

Why do you need to count pairs twice?

londumas:
@ngbusca, The thing is that we need each sub-sample to be correct. The issue is when on quasar is from healpix pixel `i` and another quasar is from healpix pixel `j`. The way we compute the correlation, the pair is going to go only in one, for example `i`, of the two healpix, not on both. This results in a incorrect partition of the pairs, that produces incorrect estimation of the correlation in each sub-sample, and then an incorrect estimation of the variance and correlation matrix. Is that clear?

ngbusca:
So the same pairs en up in two different subsamples? I'd say you want to avoid that, otherwise the subsamples will be very correlated, no ?

ngbusca:
@londumas any news? I really don't see why you want to put the same pairs on different subsamples... or maybe I misunderstood your explanation?

londumas:
@ngbusca, Here is the evolution of the variance compared to Poisson for:
- nb_random = 1 x nb_data
- nb_random = 2 x nb_data
- nb_random = 5 x nb_data

I don't have 10 times because for some reasons my Nersc jobs did't work.
You can see that the variance gets closer and closer to the Poisson (i.e. not correlated) case.
However, I would expect to get very close to the Poisson case with no contribution from the variance of the randoms. I want to send the calculation again but only up to `r = 20` and see how it goes.

> why you want to put the same pairs on different subsamples

I don't want to, I have tried to get a correct computation of the covariance matrix without, but it doesn't work. This is a pain because it produces a computation that is twice as long as before.
We can choose:
- get a faster computation but no evaluation of the true variance and of the correlation matrix
- computation twice as slow and all of the above.

 
capture d ecran de 2018-06-05 11-34-26

ngbusca:
@londumas before merging I want to understand why it's absolutely necessary to have a double computation of the pairs... I see that your tests suggest that that's the case, but do you understand why?

londumas:
@ngbusca, The way I understand it is: to compute the sub-sampling co-variance matrix you need to measure the correlation correctly in each sub-sample. When dealing with a pair involving two quasars from different healpix, it will go only in the computation of one of the two healpix. This will forbids
`(DD -DR -RD +RR)` to converge to `RR` in each sub-sample. The result for the global correlation function is however exactly the same.

londumas:
@ngbusca, What could I do to convince you that we need this PR?

ngbusca:
@londumas I'm not arguing against the PR, but there has to be a way of doing things without double counting pairs. Let me take a look at the code again, I haven't thought about it in a while...

ngbusca:
@londumas merge when ready

PR #443, opened 2018-05-30
Rescale covariance in fast MC

mblomqvist:
Adds the option to rescale the covariance matrix inside fast MC for fitter2. Useful for making simple parameter projections. Set the optional scale factor in chi2.ini using e.g. (add more values for multi-fits)

[fast mc]
covscaling = 0.5

The PR also saves the chi2 of every fast MC fit.

### DIFF
--- a/py/picca/fitter2/chi2.py
+++ b/py/picca/fitter2/chi2.py
@@ -31,6 +31,8 @@ class chi2:
             else:
                 self.seedfast_mc = 0
             self.nfast_mc = dic_init['fast mc']['niterations']
+            if 'covscaling' in dic_init['fast mc']:
+                self.scalefast_mc = dic_init['fast mc']['covscaling']
 
         if 'minos' in dic_init:
             self.minos_para = dic_init['minos']
@@ -184,8 +186,13 @@ class chi2:
         if not hasattr(self,"nfast_mc"): return
 
         sp.random.seed(self.seedfast_mc)
-
         nfast_mc = self.nfast_mc
+
+        if hasattr(self,"scalefast_mc"):
+            for d, s in zip(self.data, self.scalefast_mc):
+                d.co = s*d.co
+                d.ico = d.ico/s
+
         for d in self.data:
             d.cho = cholesky(d.co)
 
@@ -271,6 +278,8 @@ class chi2:
             g = f.create_group("fast mc")
             g.attrs['niterations'] = self.nfast_mc
             g.attrs['seed'] = self.seedfast_mc
+            if hasattr(self, "scalefast_mc"):
+                g.attrs['covscaling'] = self.scalefast_mc
             for p in self.fast_mc:
                 vals = sp.array(self.fast_mc[p])
                 d = g.create_dataset("{}/values".format(p), vals[:,0].shape, dtype="f")
--- a/py/picca/fitter2/chi2.py
+++ b/py/picca/fitter2/chi2.py
@@ -33,6 +33,8 @@ class chi2:
             self.nfast_mc = dic_init['fast mc']['niterations']
             if 'covscaling' in dic_init['fast mc']:
                 self.scalefast_mc = dic_init['fast mc']['covscaling']
+            else:
+                self.scalefast_mc = sp.ones(len(self.data))
 
         if 'minos' in dic_init:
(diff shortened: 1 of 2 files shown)


py/picca/fitter2/chi2.py
```diff
@@ -184,12 +186,18 @@ def fastMC(self):
...
+        if hasattr(self,"scalefast_mc"):
+            for d, s in zip(self.data, self.scalefast_mc):
+                d.co = s*d.co
+                d.ico = d.ico/s
+
         for d in self.data:
             d.cho = cholesky(d.co)
 
         self.fast_mc = {}
+        self.fast_mc['chi2'] = []
```
londumas: 'chi2' can be a parameter in fast_mc

py/picca/fitter2/chi2.py
```diff
@@ -31,6 +31,8 @@ def __init__(self,dic_init):
             else:
                 self.seedfast_mc = 0
             self.nfast_mc = dic_init['fast mc']['niterations']
+            if 'covscaling' in dic_init['fast mc']:
+                self.scalefast_mc = dic_init['fast mc']['covscaling']
```
londumas: set to 1 if not given

py/picca/fitter2/chi2.py
```diff
@@ -184,12 +186,18 @@ def fastMC(self):
         if not hasattr(self,"nfast_mc"): return
 
         sp.random.seed(self.seedfast_mc)
-
         nfast_mc = self.nfast_mc
+
+        if hasattr(self,"scalefast_mc"):
```
londumas: always present, but `1` if not given

py/picca/fitter2/chi2.py
```diff
@@ -203,6 +211,7 @@ def fastMC(self):
                 if not p in self.fast_mc:
                     self.fast_mc[p] = []
                 self.fast_mc[p].append([v, best_fit.errors[p]])
+            self.fast_mc['chi2'].append(best_fit.fval)
```
londumas: self.fast_mc[p].append([ sp.append(v,chi2), sp.append(best_fit.errors[p],0.)])

py/picca/fitter2/chi2.py
```diff
@@ -271,12 +280,18 @@ def export(self):
             g = f.create_group("fast mc")
             g.attrs['niterations'] = self.nfast_mc
             g.attrs['seed'] = self.seedfast_mc
+            if hasattr(self, "scalefast_mc"):
```
londumas: always present

py/picca/fitter2/chi2.py
mblomqvist: Is it more intuitive that way?

londumas: it is.

py/picca/fitter2/chi2.py
mblomqvist: I considered it but found it unnecessary to define and store it if it's unity (we don't store an empty list of priors when there are none). I can change it though if you want.

londumas: it is simply that with setting it to 1 by default you don't have to use if hasattr(self,"scalefast_mc").
As you want.

py/picca/fitter2/chi2.py
mblomqvist: I don't see how your suggestion fits in here

londumas: it doesn't anymore.

mblomqvist: How do you mean?

londumas: forget about my comment

mblomqvist:
Update to always define and save the covscaling. Codacy doesn't like 'assert', as we know.

londumas:
@mblomqvist, can you try what codacy is recommending in replacement of the assert?

ngbusca:
@mblomqvist @londumas I'll go through this PR today. Regarding the use of `assert`, it's standard python and we shouldn't be avoiding it. If coadacy complains, so be it...

londumas:
@ngbusca, I agree. However if we can do another way, why not.

mblomqvist:
@ngbusca ready to be merged?

PR #444, opened 2018-05-31
Metals boss reso correction

Cyeche:
Compute metal 1D power spectrum
Add correction of spectrograph resolution

### DIFF
--- a/bin/do_Pk1D.py
+++ b/bin/do_Pk1D.py
@@ -120,6 +120,8 @@ if __name__ == '__main__':
     parser.add_argument('--no-apply-filling', action='store_true', default = False, required=False,
                         help = ' No fill masked pixels')
 
+    parser.add_argument('--forest-type', type = str, default = 'Lya', required=False,
+                        help = ' , Forest used LY-a, SiIV or CIV')
 
     args = parser.parse_args()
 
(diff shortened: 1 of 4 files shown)


bin/do_Pk1D.py
```diff
@@ -131,10 +133,21 @@ def compute_mean_delta(ll,delta,iv,zqso):
...
+        # control histograms
+        if (args.forest_type=='Lya'):
+            forest_inf=1040.
+            forest_sup=1200.
+        elif (args.forest_type=='SiIV'):
+            forest_inf=1270.
+            forest_sup=1380.
+        elif (args.forest_type=='CIV'):
+            forest_inf=1410.
+            forest_sup=1520.
```
londumas: why not read forest_inf and forest_sup from parser.parse_args()?

py/picca/data.py
```diff
@@ -161,7 +161,7 @@ def __init__(self,ll,fl,iv,thid,ra,dec,zqso,plate,mjd,fid,order,diff=None,reso=N
 
         # compute means
         if reso is not None : self.mean_reso = sum(reso)/float(len(reso))
-
+
```
londumas: trailing spaces

bin/do_Pk1D.py
Cyeche: It is just for the histograms, by definition it is the maximal width of the forest, I want that the limits are fixed. By this way I can see if in do_deltas I have used a narrower forest

py/picca/data.py
Cyeche: Ouais I know

Issue #306, opened 2018-02-01
Remove empty subsample in do_cf.py, do_xcf.py, ...


londumas:
When saving the data in `do_cf.py`, `do_xcf.py`,... , we could look if all bins of a given sub-sample are empty, and if it is the case remove this sub-sample.

PR #445, opened 2018-06-04
remove empty sub-samples

londumas:
Fix issue https://github.com/igmhub/picca/issues/306.
Remove empty sub-samples.


py/picca/test/test_cor.py
```diff
@@ -62,7 +62,7 @@ def test_cor(self):
         self.send_cf1d()
         self.send_cf1d_cross()
 
-        self.send_cf_angl()
+        #self.send_cf_angl()
```
ngbusca: why is this test turned off?

londumas:
@ngbusca, It was for tests, I have to understand what is happening.

londumas:
@ngbusca, the issue is that in the master `py/picca/test/data/` some sub-samples are empty.
This PR removes them, then the unittest failes because there are not the same number of sub-samples. I would need to reproduce the /data/ files.
Or we can change nothing, we can live with some empty sub-samples.
What do you want?

ngbusca:
I don't mind the empty subsamples...

londumas:
ok, closing the PR and the ticket.

Issue #434, opened 2018-05-22
Better documentation


londumas:
This issue aims at better documentation:

- [x] Add a description into `argparse.ArgumentParser(description='')` to all scripts
- [x] Add a description of each parameters in fits file
- [ ] add a description for all methods and functions, like in `py/picca/utils.py:::desi_convert_transmission_to_delta_files`

londumas:
@Cyeche and @nathalieGit, could you add documentation to the fit files in `do_Pk1D.py`?
Simply have a look at how to add comments to header and units to data in fits file.
An example is https://github.com/igmhub/picca/blob/master/bin/do_xcf.py

PR #446, opened 2018-06-04
Better documentation

londumas:
Here is a commit at a proposition of better documentation.
My goal is for us to agree on one file and then go very fast on the rest of them.
Please @ngbusca and @mblomqvist, tell me what you think about name of variables
and comments.
Step toward fixing https://github.com/igmhub/picca/issues/434


bin/do_xcf.py
```diff
@@ -9,83 +9,97 @@
...
-    tmp = xcf.xcf(p)
-    return tmp
+def corr_func(pixels):
+    """Send correlation on one processor for a list of healpix
+    
+    Args:
+        pixels (list of int): list of healpix to compute
+            the correlation on.
+
+    Returs:
```
mblomqvist: Returns:

bin/do_xcf.py
```diff
@@ -9,83 +9,97 @@
...
-                    help = 'number of processors')
+    parser.add_argument('--z-cut-min', type = float, default=0., required=False,
+        help = 'Use only pairs of forest x object with the mean of the last absorber \
+        redshift and the object redshift higher than z-cut-min')
 
-    parser.add_argument('--z-min-obj', type = float, default = None, required=False,
-                    help = 'min redshift for object field')
+    parser.add_argument('--z-cut-max', type=float, default=10., required=False,
+        help = 'Use only pairs of forestx object with the mean of the last absorber \
+        redshift and the object redshift smaller than z-cut-max')
```
mblomqvist: forest x object

bin/do_xcf.py
```diff
@@ -9,83 +9,97 @@
...
-    parser.add_argument('--nside', type = int, default = 16, required=False,
-                    help = 'healpix nside')
+    parser.add_argument('--z-max-obj', type=float, default=None, required=False,
+        help = 'Max redshift for object field')
 
-    parser.add_argument('--nproc', type = int, default = None, required=False,
-                    help = 'number of processors')
+    parser.add_argument('--z-cut-min', type = float, default=0., required=False,
+        help = 'Use only pairs of forest x object with the mean of the last absorber \
+        redshift and the object redshift higher than z-cut-min')
```
mblomqvist: larger than

bin/do_xcf.py
```diff
@@ -9,83 +9,97 @@
...
 
-    parser.add_argument('--no-remove-mean-lambda-obs', action="store_true", required=False,
-                    help = 'Do not remove mean delta versus lambda_obs')
+    parser.add_argument('--nproc', type=int, default=None, required=False,
+        help = 'Number of processors')
 
-    parser.add_argument('--from-image', type = str, default = None, required=False,
-                    help = 'use image format to read deltas', nargs='*')
+    parser.add_argument('--nspec', type=int, default=None, required=False,
+        help = 'Maximum spectra to read')
```
mblomqvist: Maximum number of

bin/do_xcf.py
```diff
@@ -186,17 +200,22 @@ def corr_func(p):
...
-    head['NSIDE']=xcf.nside
-
-    out.write([rp,rt,z,nb],names=['RP','RT','Z','NB'],header=head)
-    head2 = [{'name':'HLPXSCHM','value':'RING','comment':'healpix scheme'}]
-    out.write([hep,wes,cfs],names=['HEALPID','WE','DA'],header=head2)
+    head = [ {'name':'RPMIN','value':xcf.rp_min,'comment':'Minimum r-parallel'},
+        {'name':'RPMAX','value':xcf.rp_max,'comment':'Maximum r-parallel'},
+        {'name':'RTMAX','value':xcf.rt_max,'comment':'Maximum r-transverse'},
+        {'name':'NP','value':xcf.np,'comment':'Number bin r-parallel'},
+        {'name':'NT','value':xcf.nt,'comment':'Number bin r-transverse'},
```
mblomqvist: Number of bins (if that many characters are allowed in a header)

bin/do_xcf.py
```diff
@@ -186,17 +200,22 @@ def corr_func(p):
...
+        {'name':'Z_CUT_MIN','value':xcf.z_cut_min,'comment':'Minimum redshift of pairs'},
```
mblomqvist: Minimum redshift of pairs is not strictly true, but I have no better suggestion.

bin/do_xcf.py
```diff
@@ -186,17 +200,22 @@ def corr_func(p):
...
+        {'name':'Z_CUT_MAX','value':xcf.z_cut_max,'comment':'Maximum redshift of pairs'},
+        {'name':'NSIDE','value':xcf.nside,'comment':'Healpix nside'}
+    ]
+    out.write([rp,rt,z,nb],names=['RP','RT','Z','NB'],
+        comment=['R-parallel','R-transverse','Redshift','Number of pairs'],
+        header=head,extname='ATTRIBUTES')
+
+    head2 = [{'name':'HLPXSCHM','value':'RING','comment':'Healpix scheme'}]
```
mblomqvist: Separate comments for HLPXSCHM and RING?

mblomqvist: Sorry, I read that wrong.

bin/do_xcf.py
londumas: indeed.

londumas:
Thank you for the review @mblomqvist.

bin/do_xcf.py
```diff
@@ -186,17 +200,22 @@ def corr_func(p):
...
-    out.write([rp,rt,z,nb],names=['RP','RT','Z','NB'],header=head)
-    head2 = [{'name':'HLPXSCHM','value':'RING','comment':'healpix scheme'}]
-    out.write([hep,wes,cfs],names=['HEALPID','WE','DA'],header=head2)
+    head = [ {'name':'RPMIN','value':xcf.rp_min,'comment':'Minimum r-parallel'},
+        {'name':'RPMAX','value':xcf.rp_max,'comment':'Maximum r-parallel'},
+        {'name':'RTMAX','value':xcf.rt_max,'comment':'Maximum r-transverse'},
+        {'name':'NP','value':xcf.np,'comment':'Number of bins in r-parallel'},
+        {'name':'NT','value':xcf.nt,'comment':'Number of bins in r-transverse'},
+        {'name':'Z_CUT_MIN','value':xcf.z_cut_min,'comment':'Minimum redshift of pairs'},
+        {'name':'Z_CUT_MAX','value':xcf.z_cut_max,'comment':'Maximum redshift of pairs'},
```
ngbusca: @londumas could you change `Z_CUT_MIN` and `Z_CUT_MAX` to `ZCUTMIN` and `ZCUTMAX` ? keyword names in .fits must be 8 characters or less...

londumas:
@ngbusca, @mblomqvist, @vserret, would you accept to split the work on adding documentation?
I could take care of all the files in `bin/` and also in `py/picca/test/`. We have a growing number of user and it would be nice to have the documentation written. You would just need to add a preamble like the following one to all functions and methods and add a comment to all variables written FITS file and add a `EXTNAME`.
```
    """Send correlation on one processor for a list of healpix
    
    Args:
        pixels (list of int): list of healpix to compute
            the correlation on.

    Returns:
        cor (list of scipy array): list of array with the
            computed correlation and other attributes.
    
    """

```

ngbusca:
I think this effort would be much more useful if directed towards filling readthedocs. It's ok to have good command help messages but I doubt someone is going to learn to run picca by reading them.

londumas:
I though that readthedocs was using automatically the documentation inside the code. Is it not?
I totally agree that documentation of functions is not that useful for end-user, however I do believe that:
- documentation, i.e. comments, inside produced FITS file is very important.
- documentation of our `do_*.py` scripts is very useful for end-users.
What do you think @ngbusca and @mblomqvist ?

mblomqvist:
@londumas I agree that comments in FITS files is useful and that the help messages in the do_*.py scripts may need some updating. For example, the 'mode' option of do_deltas.py doesn't mention that one can run on spPlates.

ngbusca:
@londumas tell me when it's ready

londumas:
@ngbusca, it is ready.
Unitest are very useful !

PR #448, opened 2018-06-07
Option to modify fast MC fiducial model

mblomqvist:
This PR adds the option to modify the fast MC fiducial model by specifying parameter values in chi2.ini, e.g.

[fast mc]
ap = 1.0
at = 1.0

Parameters not specified are taken from the best fit, as before. The specified fiducial parameter values are saved in the output file (if any; similar to priors). The fiducial model is also saved.

### DIFF
--- a/py/picca/fitter2/chi2.py
+++ b/py/picca/fitter2/chi2.py
@@ -35,6 +35,7 @@ class chi2:
                 self.scalefast_mc = dic_init['fast mc']['covscaling']
             else:
                 self.scalefast_mc = sp.ones(len(self.data))
+            self.fidfast_mc = dic_init['fast mc']['fiducial']
 
         if 'minos' in dic_init:
             self.minos_para = dic_init['minos']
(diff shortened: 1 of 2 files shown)


londumas:
@mblomqvist, thank you for this PR. It would be nice also to add the possibility to fix the parameter. It was useful in crossDR12 paper. With something like this:
[fast mc]
ap = 1.0 fixed
at = 1.0 free

mblomqvist:
@londumas Yes, I see what you mean. When running fast MC for the purpose of mapping between Delta-chi2 and confidence, one would need the chi2 value corresponding to the fiducial parameter. An option is to run the second instance of the fit with fixed ap=at=1 and specify all the other parameters to take the values of the first instance. But I agree that it is a bit more work for the user. I'll take a look.

mblomqvist:
Updated according to the suggestion by @londumas Use e.g.

[fast mc]
ap = 1.0 fixed
at = 1.0 free

where fixed/free tells the fit model how to treat the parameter in fast MC.

mblomqvist:
ready to be merged, pending review

ngbusca:
@londumas can you make a review?

py/picca/fitter2/parser.py
```diff
@@ -55,14 +55,23 @@ def parse_chi2(filename):
...
+            elif item=='covscaling':
                 value = value.split()
                 dic_init['fast mc'][item] = sp.array(value).astype(float)
                 if not len(dic_init['fast mc'][item])==len(dic_init['data sets']['data']):
                     raise AssertionError()
             else:
-                dic_init['fast mc'][item] = int(value)
+                value = value.split()
+                dic_init['fast mc']['fiducial']['values'][item] = float(value[0])
+                if not (value[1]=='fixed' or value[1]=='free'):
```
londumas: `if not value[1] in ['fixed','free']:`
it is simplier

py/picca/fitter2/chi2.py
```diff
@@ -195,13 +197,34 @@ def fastMC(self):
             d.ico = d.ico/s
             d.cho = cholesky(d.co)
 
+        self.fiducial_values = self.best_fit.values.copy()
+        if len(self.fidfast_mc) != 0:
```
londumas: would simply `for p in self.fidfast_mc:` work?
If empty then do not enter the loop

py/picca/fitter2/chi2.py
```diff
@@ -195,13 +197,34 @@ def fastMC(self):
...
+            d.fiducial_model = self.fiducial_values['bao_amp']*d.xi_model(self.k, self.pk_lin-self.pksb_lin, self.fiducial_values)
+
+            self.fiducial_values['SB'] = True
+            snl = self.fiducial_values['sigmaNL_per']
+            self.fiducial_values['sigmaNL_per'] = 0
+            d.fiducial_model += d.xi_model(self.k, self.pksb_lin, self.fiducial_values)
+            self.fiducial_values['SB'] = False
+            self.fiducial_values['sigmaNL_per'] = snl
+        del self.fiducial_values['SB']
+
```
londumas: @mblomqvist, why do you do this? Is it in case a parameter value has changed?

py/picca/fitter2/chi2.py
```diff
@@ -280,6 +303,17 @@ def export(self):
             g.attrs['niterations'] = self.nfast_mc
             g.attrs['seed'] = self.seedfast_mc
             g.attrs['covscaling'] = self.scalefast_mc
+            if len(self.fidfast_mc) != 0:
+                fid = []
+                for p in self.fidfast_mc:
+                    fix = "fixed"
+                    if self.fixfast_mc['fix_'+p] == False: fix = "free"
```
londumas: `if not self.fixfast_mc['fix_'+p]:`
works well also

py/picca/fitter2/chi2.py
mblomqvist: That's right

PR #450, opened 2018-06-12
Remove metals with redshift larger than zqso

londumas:
When calculating the metal matrix, remove metal absorption with redshift larger than the host quasar redshift. @ngbusca, this is similar to what you and @vserret corrected for auto-correlation.


londumas:
@ngbusca, ready to be merged. I tested on data and it works also.

PR #452, opened 2018-06-14
returning the unsmooth covriance when there are empty bins

ngbusca:
This PR include a minor change in the behavior of `export.py` when empty bins are present.

### DIFF
--- a/py/picca/utils.py
+++ b/py/picca/utils.py
@@ -31,8 +31,8 @@ def smooth_cov(da,we,rp,rt,drt=4,drp=4):
     nda = da.shape[1]
     var = sp.diagonal(co)
     if sp.any(var==0.):
-        print('ERROR: data has some empty bins, impossible to smooth')
-        print('ERROR: returning the unsmoothed covariance')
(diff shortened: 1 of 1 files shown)


py/picca/utils.py
```diff
@@ -32,7 +32,8 @@ def smooth_cov(da,we,rp,rt,drt=4,drp=4):
     var = sp.diagonal(co)
     if sp.any(var==0.):
         print('ERROR: data has some empty bins, impossible to smooth')
```
londumas: Small detail, it should be Warning

py/picca/utils.py
```diff
@@ -32,7 +32,8 @@ def smooth_cov(da,we,rp,rt,drt=4,drp=4):
...
-        sys.exit()
+        print('ERROR: returning the unsmoothed covariance')
```
londumas: Small detail, it should be Warning

ngbusca:
@londumas thanks for your comments. Can you take another look?

ngbusca:
Thanks @londumas !!

PR #454, opened 2018-06-14
Option to mask absorbers in spectra

mblomqvist:
This PR introduces to do_deltas the option to mask pixels around identified absorbers at the per-spectrum level. Reads an ascii catalog with columns "ThingID" and "lambda" (analogous to the DLA catalog). Input the number of spectral pixels to remove on each side of the central wavelength.

### DIFF
--- a/bin/do_deltas.py
+++ b/bin/do_deltas.py
@@ -86,7 +86,7 @@ if __name__ == '__main__':
         help='Absorber catalog file')
 
     parser.add_argument('--absorber-mask',type=float,default=2.5,required=False,
-        help='Number of spectral pixels to mask on each side of the absorber central observed wavelength')
+        help='Mask width on each side of the absorber central observed wavelength in units of 1e4*dlog10(lambda)')
 
     parser.add_argument('--mask-file',type=str,default=None,required=False,
         help='Path to file to mask regions in lambda_OBS and lambda_RF. In file each line is: region_name region_min region_max (OBS or RF) [Angstrom]')
--- a/bin/do_deltas.py
+++ b/bin/do_deltas.py
@@ -86,7 +86,7 @@ if __name__ == '__main__':
         help='Absorber catalog file')
 
     parser.add_argument('--absorber-mask',type=float,default=2.5,required=False,
-        help='Number of spectral pixels to mask on either side of the absorber central observed wavelength')
+        help='Number of spectral pixels to mask on each side of the absorber central observed wavelength')
 
     parser.add_argument('--mask-file',type=str,default=None,required=False,
         help='Path to file to mask regions in lambda_OBS and lambda_RF. In file each line is: region_name region_min region_max (OBS or RF) [Angstrom]')
@@ -239,7 +239,7 @@ if __name__ == '__main__':
 
     ### Veto absorbers
     if not args.absorber_vac is None:
-        print("removing absorbers")
+        print("adding absorbers")
         absorbers = io.read_absorbers(args.absorber_vac)
         nb_absorbers_in_forest = 0
         for p in data:
--- a/bin/do_deltas.py
+++ b/bin/do_deltas.py
@@ -82,6 +82,12 @@ if __name__ == '__main__':
     parser.add_argument('--dla-mask',type=float,default=0.8,required=False,
         help='Lower limit on the DLA transmission. Transmissions below this number are masked')
 
+    parser.add_argument('--absorber-vac',type=str,default=None,required=False,
+        help='Absorber catalog file')
+
+    parser.add_argument('--absorber-mask',type=float,default=2.5,required=False,
+        help='Number of spectral pixels to mask on either side of the absorber central observed wavelength')
+
     parser.add_argument('--mask-file',type=str,default=None,required=False,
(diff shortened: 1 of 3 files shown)


mblomqvist:
Using this feature to mask identified strong Lya absorbers removes the cross-covariance between the Lya auto and SLA-Lya cross.

mblomqvist:
ready to be merged, pending review

py/picca/io.py
```diff
@@ -43,6 +43,34 @@ def read_dlas(fdla):
 
     return dlas
 
+def read_absorbers(file_absorbers):
```
londumas: why not give it a fits file? It is so much better than ascii...

py/picca/data.py
```diff
@@ -232,6 +235,19 @@ def add_dla(self,zabs,nhi,mask=None):
             self.diff = self.diff[w]
             self.reso = self.reso[w]
 
+    def add_absorber(self,lambda_absorber):
+        if not hasattr(self,'ll'):
+            return
+
+        w = sp.ones(self.ll.size).astype(bool)
+        w = w & (sp.fabs(1e4*(self.ll-sp.log10(lambda_absorber)))>forest.absorber_mask)
```
londumas: `w &=` is more beautiful.
I don't understand the `1e4*`, is it because of the size of the pixel? It is not very robust.

bin/do_deltas.py
```diff
@@ -82,6 +82,12 @@ def cont_fit(data):
     parser.add_argument('--dla-mask',type=float,default=0.8,required=False,
         help='Lower limit on the DLA transmission. Transmissions below this number are masked')
 
+    parser.add_argument('--absorber-vac',type=str,default=None,required=False,
+        help='Absorber catalog file')
+
+    parser.add_argument('--absorber-mask',type=float,default=2.5,required=False,
+        help='Number of spectral pixels to mask on each side of the absorber central observed wavelength')
```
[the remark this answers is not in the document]
mblomqvist: I chose float to give a bit more flexibility

py/picca/io.py
mblomqvist: Analogous to the DLA catalog and line mask.

py/picca/data.py
mblomqvist: The pixels size is hardcoded to SDSS, which isn't great, but note that the factor 1e-4 is already hardcoded elsewhere in do_deltas so I decided not to change that in this PR

londumas: but you can simply use `forest.dll` no?

mblomqvist: forest.dll is rescaled by args.rebin though. I want spectral pixel (i.e. rebin=1) always.

londumas: why?

mblomqvist: Because the absorber mask is then always defined in this way, independent of other choices. People talk about masking pixels at the absorber wavelength and I want to avoid any confusion what "pixel" means.

ngbusca:
Shouldn't the size of the mask be defined in the vac? In the DLA catalog, this is given by nhi, I'd expect the same kind of input here, no?

mblomqvist:
@ngbusca No, there is no physical parameter to define this, just a central wavelength. The user decides how conservative to be (not unlike masking Galactic absorption or strong sky lines)

ngbusca:
@mblomqvist right, my point is that the user should include that in the vac as another column, then he can chose the width on an object-by-object basis and not as a one-size-fits-all argument to `do_deltas.py`

mblomqvist:
@ngbusca Well, it's a catalog of absorbers, not a "mask" per se (e.g. the DLA catalog doesn't specify where the transmission limit should be set for masking them). Since the correct width to mask isn't really known, the catalog maker would need to generate different versions with different widths. I can enforce this principle if you really want to, though.

bin/do_deltas.py
ngbusca: Regarding the discussion of `1e4`, I would change the help to something like `mask width in units of dlambda/lambda*1e-4`

ngbusca: I agree that it would be a bit constraining to put it in the catalog, I'm fine with adding a command line option.

mblomqvist:
ready to be merged

ngbusca:
@londumas can you do a review?

ngbusca:
@londumas could you review this branch ?

py/picca/data.py
```diff
@@ -232,6 +235,19 @@ def add_dla(self,zabs,nhi,mask=None):
             self.diff = self.diff[w]
             self.reso = self.reso[w]
 
+    def add_absorber(self,lambda_absorber):
+        if not hasattr(self,'ll'):
+            return
+
+        w = sp.ones(self.ll.size).astype(bool)
```
londumas: sp.ones(self.ll.size, dtype=bool)

py/picca/data.py
```diff
@@ -232,6 +235,19 @@ def add_dla(self,zabs,nhi,mask=None):
...
+        w &= (sp.fabs(1e4*(self.ll-sp.log10(lambda_absorber)))>forest.absorber_mask)
```
londumas: w &= sp.fabs(1.e4*(self.ll-sp.log10(lambda_absorber)))>forest.absorber_mask

PR #465, opened 2018-07-03
use input correlation-matrix

londumas:
Add the possibility to give an input correlation matrix but get the variance from sub-samples.


londumas:
@ngbusca I made the changes. Ready to be merged, Codacy fails because of the use of `assert`

ngbusca:
@londumas thanks for the changes, but it would be nice to tell the user why the code would fail. By the way, do you think we should issue an `ERROR` and refuse to produce output, or a `WARNING` and use whatever the user gave as correlation anyway?

londumas:
@ngbusca, I just did the corrections.

PR #466, opened 2018-07-03
add Info message to track chi2scan status

londumas:
add Info message to track chi2scan status


mblomqvist:
@londumas Can you give the current step count too? Right now the screen output only says that an iteration finished

londumas:
@mblomqvist, does this work for you?

PR #467, opened 2018-07-06
Fit for relativistic effects

mblomqvist:
This PR adds the option to fit for the amplitudes of the dipole and octupole contribution from relativistic effects in the object-forest cross-correlation. It is not yet ready for a full review. @ngbusca could you take a look at whether the FFTlog implementation seems OK?

### DIFF
--- a/py/picca/fitter2/utils.py
+++ b/py/picca/fitter2/utils.py
@@ -40,6 +40,9 @@ def Pk2Mp(ar,k,pk,ell_vals,tform=None):
         if tform=="rel":
             pk_ell=pk
             n=1.
+        elif tform=="asy":
+            pk_ell=pk
+            n=2.
         else:
             pk_ell=sp.sum(dmuk*L(muk,ell)*pk,axis=0)*(2*ell+1)*(-1)**(ell//2)/2/sp.pi**2
             n=2.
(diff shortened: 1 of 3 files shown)


mblomqvist:
A preliminary fit to QSO-Lya yields a small improvement in the goodness-of-fit and a one-sigma significant dipole signal.

mblomqvist:
The plot shows the best-fit dipole and octupole, which look reasonable and have amplitudes in line with expectations.

dipole_octupole_rperp2_log

mblomqvist:
This PR is now open for review. Codacy doesn't like the use of "assert".

ngbusca:
@mblomqvist thanks a lot for implementing this. In order to reduce code duplication, would it be possible to modify the current `Pk2Mp` and `Pk2Xi` functions to accommodate odd `ell` ? This might imply handling the `mu` integral over -1 to 1 instead of the current 0 to 1, but I think it would make more sense. 

Also, don't worry about the `assert`, it's valid python code.

ngbusca:
`the effects are conceptually different`

I agree.

`The relativistic effects are not the odd multipoles for the Lya transmission power spectrum, but a FT of the isotropic P(k) with a different k weight factor` 

The FT machinery should be independent of the physics that goes into the `P(k)`, no?
E.g., you can write `P_full(k) = P_lya(k,mu) + P_rel_1(k)*L_1(mu) + P_rel_3(k)*L_3(mu)` where `P_rel_1,3(k)` are the isotropic `P(k)` with the appropriate `k` factor. This would allow you to use the current `Pk2Mp` function (with minor modifications) without code duplication (e.g., one day we might want to use a more efficient log FFT, we don't want to have to change that in several functions).

What do you think?

mblomqvist:
unique to Lya FFT: extra integral over mu, go to arbitrary ell_max, n = 2, different integrand.
unique to relativistic FFT: restricted to ell = 1 and 3, n = 1, different integrand.
One should probably pass a list of desired ell values to Pk2Mp, and then rely on a couple of if statements. There is superficially a slight confusion here that e.g. ell = 1 will refer to the relativistic dipole and not the Lya dipole, but since we're not using odd Lya multipoles, I guess nobody will complain about it.

ngbusca:
I guess what I'm arguing for is to generalize `Pk2Mp` and `Pk2Xi` such that it will handle odd `ell`. Then, to add a relativistic model, you would add a `pk_rel_irsic` function to pk.py that will have even and odd mu powers. For some models that would be odd powers up to 3.

I would like to avoid, to the extent that it's possible!, writing an FT dedicated for a particular relativistic model since it goes against the philosophy of `fitter2`. If it's impossible then so be it, but I don't see why it would (but I'm not the one implementing it so...)

mblomqvist:
Sure, I see your point and can modify the code along the lines of what I mentioned above to only have one FFT function. I tried to be non-invasive to the pre-existing code when the update was in the test phase.

mblomqvist:
@ngbusca the latest commit modifies the transformation function to address your request. Now Pk2Mp is the only function. It gets passed a list of desired ell values and a string to specify the type of transformation. Note that a third transformation type might get added in a future update if we decide to introduce also the "standard" dipole.

I simply commented out the old code to keep it for reference during review. I'll clean it up in the next commit if you agree that this new modification is the correct approach.

ngbusca:
@mblomqvist I still think this could be further simplified to make the `Pk2Mp` totally agnostic of the Pk physics. Can I make the changes directly in your branch to show you what I mean? we can always roll back thanks to the magic of github.

mblomqvist:
@ngbusca yes, please go ahead

ngbusca:
@mblomqvist sorry I'm delaying this... I won't be able to work on it before next week. But it's a matter of implementation and it shouldn't change the results, so I guess you can safely work with the results from this branch as if it was master.

mblomqvist:
@ngbusca OK. I also want to test to add the "standard" asymmetry terms in the fit, but that could in principle be a separate PR later if I'm unable to get started on it today.

londumas:
@mblomqvist, when do you think we need to review this PR and it will be ready to be merged?

mblomqvist:
@londumas I will run some new fits and make a final check that everything seems OK. I'll let you know. @ngbusca wanted to restructure the transformation code, but that could be done later in a separate PR when he wants.

ngbusca:
(moving my comment to the right PR) @mblomqvist @londumas go ahead whenever the branch is ready to be merged. I'll restructure things on a later branch as soon as I have time...

londumas:
@mblomqvist, could you add docstrings to every new functions and methods?
I don't ask to write the doc for existing ones, but we must start wrtiting doc inside the functions
so that other users, or us in 6 months can understand the code.
See an example here: `desi_from_ztarget_to_drq` in `py/picca/utils.py`.

mblomqvist:
@londumas docstrings added for new functions in fitter2/utils.py

py/picca/fitter2/xi.py
```diff
@@ -114,6 +114,36 @@ def xi_qso_radiation(r, mu, tracer1, tracer2, **pars):
 
     return xi_rad
 
+### Relativistic correction
+def xi_relativistic(r, mu, k, pk_lin, tracer1, tracer2, **pars):
```
londumas: Add docstring

py/picca/fitter2/xi.py
```diff
@@ -114,6 +114,36 @@ def xi_qso_radiation(r, mu, tracer1, tracer2, **pars):
...
+    arp = ap*rp
+    art = at*rt
+    ar = sp.sqrt(arp**2+art**2)
+    amu = arp/ar
+
+    xi_rel = utils.Pk2XiRel(ar, amu, k, pk_lin, pars)
+    return xi_rel
+
+### Standard asymmetry
+def xi_asymmetry(r, mu, k, pk_lin, tracer1, tracer2, **pars):
```
londumas: Add docstring

py/picca/fitter2/utils.py
```diff
@@ -56,11 +63,48 @@ def Pk2Mp(ar,k,pk,ell_max=None):
     return xi
 
 def Pk2Xi(ar,mur,k,pk,ell_max=None):
-    xi=Pk2Mp(ar,k,pk,ell_max)
-    for ell in range(0,ell_max+1,2):
+    ell_vals=[ell for ell in range(0,ell_max+1,2)]
```
londumas: `range(ell_max+1,2)`

mblomqvist: range needs three arguments representing start, end, step size.

londumas: I don't think so. If `start` is not given, it is assumed to be `0`
`range(10)` gives `[0, 1, 2, 3, 4, 5, 6, 7, 8, 9]`.
In the same way, if `step` is not given it is assumed to be `1`.
But you can keep this code, it was only intended as a remark.

londumas: yes you are right, sorry for the noise.

londumas:
@mblomqvist, thank you for this PR.

PR #470, opened 2018-09-05
allow to read path variables

londumas:
allow to read path variables


londumas:
@ngbusca, ready to be merged

PR #472, opened 2018-09-06
force print fastMC

londumas:
This PR adds forcing fastMC to print the steps of the calculation to better track it

### DIFF
--- a/py/picca/fitter2/chi2.py
+++ b/py/picca/fitter2/chi2.py
@@ -159,7 +159,7 @@ class chi2:
                     if par in d.pars_init.keys():
                         d.pars_init[par] = step
                 result += [send_one_fit()]
-                print("INFO: finished chi2scan iteration {} of {}".format(it+1,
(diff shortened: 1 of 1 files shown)


ngbusca:
what's the problem with the current `print`?

londumas:
The issue with `print` is that when running `fitter2 chi2.ini > send.txt &`, it really prints only when the printing memory is full. Then it is not easy to keep track at what step the calculation is at.

mblomqvist:
@londumas Does the same issue affect the parameter scan print output too? If so, could you change it too?

londumas:
@mblomqvist, done.

ngbusca:
all this `sys/print` conundrum comes about because we want to keep python2 compatibility... eventually (in python >= 3.6) we should use systematically `print(...,flush=True)`. Let's go with Hélion's solution for now... @londumas anything else you want to change before merging?

londumas:
@ngbusca, makes sense. `print(...,flush=True)` sounds better.
Ready to be merged.

ngbusca:
@londumas can you check if you can merge yourself? Don't forget to erase the branch

Issue #475, opened 2018-10-02
read_from_spplate assumes that all observation have same fiberid


londumas:
https://github.com/igmhub/picca/blob/550dda84f2702578e783ebf24b3b20acd859de84/py/picca/io.py#L514

when reading `read_from_spplate` using all available observation, there are two errors:
- each observation don't have the same fiberid, when the code assumes it does.
The solution is to read photoPosPlate and get the proper fiberid
- we then use all observation, even if it is marked as bad.
One solution would be to add reading the platequality.txt and read it only if plate is good.

@mblomqvist, @vserret, @solenechabanier, do you use all observation or only the best ones?

Here is all observation before:

![before_bug_fiberid](https://user-images.githubusercontent.com/3166436/46366938-9597f700-c639-11e8-95c3-e6f4195f5c82.png)

Here are all observation after the bug is fixed:

![after_bug_fiberid](https://user-images.githubusercontent.com/3166436/46366941-97fa5100-c639-11e8-87ae-29241c809825.png)

mblomqvist:
@londumas I run analysis on pix files, not internal coadding in picca for spPlates.

londumas:
@mblomqvist or @slosar, could you have a look if ForetFusion is not also doing this mistake

ngbusca:
@londumas you are right, it's a bug! Thanks for reporting. Is there a reason for using spPhotoPlate instead of spAll?

ngbusca:
When I wrote the code I assumed that a plate would have the same plugin for different mjds which is obviously false...

londumas:
They have nothing in common.
spPhotoPlate is similar to spPlate: one per observation. It gives the mapping THING_ID -> Fiber_ID
spAll is a catalog of only "good" observation

@ngbusca, how would you like us to proceed not to use bad plates? Is reading the "platequality.txt" file sound good?

londumas:
@slosar, it seems to be correctly done here.
https://github.com/igmhub/ForetFusion/blob/20cff1851ff424a7d1106c1128c583887213dbb3/ffusion/automatic.py#L69
Do you confirm?

ngbusca:
@londumas @mblomqvist ForetFusion doesn't run into this problem because it uses the reobservation plate,mjd,fid info from DR14Q.

londumas:
Fixed.

PR #476, opened 2018-10-02
fix reading proper fiberid in spplate

londumas:
Fix issue https://github.com/igmhub/picca/issues/475:
Reading the proper fiberid when using multiple observation of spplates

### DIFF
--- a/bin/do_deltas.py
+++ b/bin/do_deltas.py
@@ -200,7 +200,7 @@ if __name__ == '__main__':
     data,ndata,healpy_nside,healpy_pix_ordering = io.read_data(os.path.expandvars(args.in_dir), args.drq, args.mode,\
         zmin=args.zqso_min, zmax=args.zqso_max, nspec=args.nspec, log=log,\
         keep_bal=args.keep_bal, bi_max=args.bi_max, order=args.order,\
-        best_obs=args.best_obs, single_exp=args.single_exp, pk1d=args.delta_format)
+        best_obs=args.best_obs, single_exp=args.single_exp, pk1d=args.delta_format )
 
     ### Get the lines to veto
     usr_mask_obs    = None
--- a/bin/do_deltas.py
+++ b/bin/do_deltas.py
@@ -47,6 +47,9 @@ if __name__ == '__main__':
     parser.add_argument('--single-exp',action='store_true', required=False,
         help='If mode == spcframe, then use only one of the available exposures. If best-obs then choose it among those contributing to the best obs')
 
+    parser.add_argument('--platequality',type=str,default=None,required=False,
+        help="Remove 'bad' observations from platequality file, works for mode in [spcframe,spplate]")
+
     parser.add_argument('--zqso-min',type=float,default=None,required=False,
         help='Lower limit on quasar redshift from drq')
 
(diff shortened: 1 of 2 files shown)


ngbusca:
@londumas I see that photoPosPlate seems better than spAll. But your solution assumes that the code is run on an environment where all those files are available. Could you add a check such that if photoPosPlate is not found the code tells the user to run with the `--bestobs` option and exits?

londumas:
@ngbusca, yes this is a good idea. Doing this now.

londumas:
@ngbusca, This fixes reading the proper fiberid in spplate or spcframe.
I had to change the way `str_iexp` was used in spcframe because it is a zfill(3) now in `$BOSS_SPECTRO_REDUX/v5_11_0/`

ngbusca:
@londumas let me know when it's ready for review

londumas:
@ngbusca, ok. I am adding the possibility to use the platequality.txt file.

londumas:
@ngbusca, ready to be reviewd.
This PR also adds a `--platequality` option to `do_deltas.py` that takes a file with a similar
format as platequality.txt of the SDSS pipeline, i.e. at least
```
 PLATE   MJD   QUALITY 
  3520 55071  bad
```
This is useful to remove bad observation when using all observations.

londumas:
@ngbusca, ok. Let me do this.

londumas:
@ngbusca, I corrected your last request.
The last thing we can do is that dome objects are observed on different plate.
Currently we are missing them, a simple solution is to change `wfib` to something like
`sp.in1d(thid,h_photoPlate[1]['THING_ID'][:])`. Does this work for you?

londumas:
@ngbusca, Ready to be attentively reviewed.

ngbusca:
@londumas I have a doubt now, aren't we also losing reobservations that are in plates different from the bestobs?

ngbusca:
I think we should take the time to get this right: if the option `--bestobs` is used then do what we do now. Otherwise: open spAll (discover it from the given plates directory) and get all the plate-mjd-fid for all the given thing_id from it. I think spAll contains only good plates (?). This avoids the files platequality.txt and spPhotoPlate files from being opened. What do you think?

londumas:
@ngbusca, I think that what we currently have corrects the bug for DR14 papers. Let's merge this and work on a refactor for DR16. It is only a vers small fraction of objects that will have a higher SNR.

As far as I have seen of it, spAll is not the file we need, we need a file that I produce for redrock that map thingid to all  observations. I don't think spAll does this. Does it?
It is in `$DESI_WORK/rrboss/SDSS_redrock_archetypes_newDeltaChi2/spplates/map_thingid_targetids_photoPosPlate.fits` for v5_10_10.

By the way, don't use `photoPlate`, use `photoPosPlate`. `photoPlate` is based on luminosity, and doesn't work. `photoPosPlate` is based on the position, and this works.

ngbusca:
@londumas I think this is the best time to fix this once and for all. spAll contains a list of all observations in good plates. So if you select a THING_ID you can get all plate-mjd-fid that correspond to that THING_ID. So I think yes, it has the info that we need. I'm find using your file if it's produced automatically by the pipeline.

ngbusca:
spAll contains a column for THING_ID, which is repeated for all the observations. There's a column SPECPRIMARY which is 1 for the best. I think spAll only contains good plates, so there's no need for platequality.txt (but it wold be good to check that a few bad plates are not in spAll).

londumas:
@ngbusca, why not simply use PLATE_DUPLICATE, MJD_DUPLICATE, FIBERID_DUPLICATE from `$EBOSS_ROOT/qso/DR14Q/DR14Q_v3_1.fits`. This seems so much simpler.

ngbusca:
I don't think that format will be kept for DR16...

ngbusca:
but I would be ok with that being a temporary solution...

ngbusca:
This issue will be fixed in PR #479

PR #479, opened 2018-10-04
fix issue #475

ngbusca:
This branch addresses issue #475 

The changes affect only the --mode spplate and spcframe options

 * When run with the --best-obs option the code behaves as before
 * otherwise, it uses spAll (by automagically discovering where it is) to match THING_ID with all the available observations, and uses that to get all the fiberid right.
 * I think the logic of the code is much simpler now


ngbusca:
@londumas it's ready for review

Issue #480, opened 2018-10-05
Add option to plot absolute value of mu for wedgize


londumas:
https://github.com/igmhub/picca/blob/5c3f8b2386d1e1017a17280d500bfdadd14e82fb/py/picca/wedgize.py#L13

PR #481, opened 2018-10-05
add absoluteMu option to wedge

londumas:
add absoluteMu option to wedge: fix https://github.com/igmhub/picca/issues/480
This is useful to plot xcf in wedgies.


londumas:
@ngbusca, yes, I will take care of this today.

PR #488, opened 2018-10-14
add 1d cross-correlation

londumas:
Add the 1d cross-correlation. For example DLA from a LOS with pixels from the same LOS.


londumas:
@ngbusca, ok for me to merge?

ngbusca:
sure!

PR #489, opened 2018-10-19
Faster cf1d

londumas:
Faster cf1d by only looping on same forests


ngbusca:
why is the proposed code is supposed to be faster?

londumas:
`if (d1.thid != d2.thid): continue` is long inside a loop.
It is better to loop only on useful forests

ngbusca:
both are actually pretty inefficient ways of doing this, since `i in data2.keys()` runs in linear time. For a sizeable performance improvement I would use `in1d`.

ngbusca:
@londumas any news on this PR? The solution you propose indeed improves the running time of that particular piece of code, but it still runs in quadratic time. Using `sp.in1d` would simplify the code and make that piece run in linear time.

londumas:
@ngbusca, I found the way to use `sp.in1d` by converting the list into arrays.
Ready to be reviewed, apparently @vserret can't with her browser.

londumas:
@ngbusca, I fixed all the travis errors. Ready to be reviewed.

ngbusca:
@londumas thanks for the changes, go ahead and merge when ready!

PR #490, opened 2018-10-23
Add cut on objectxobject pair redshift

londumas:
Add cut on objectxobject pair redshift.
This will allow to run the auto correlations of quasars on desi mocks in different redshift bin and see if the bias, beta and growth rate is correct.


londumas:
@ngbusca, ready to be reviewed.

PR #493, opened 2018-10-29
including additive and/or multiplicative broadbands

ngbusca:
This PR adds the possibility of including additive and/or multiplicative broadbands to the xi-model, both pre and post distortion and in r,mu or rp,rt

Example usage (add the following section to the individual .ini of the datasets):

    [broadband]
    bb1 = add pos r,mu 0:2:1 0:6:2

adds an `add`itive `pos`t-distortion broadband where monomials are proportional to `r^i mu^j` with `i=0,1,2` and `j=0,2,4,6`. The name `bb1` is arbitrary and ignored by the code (but required for the .ini format).

Other examples:

    ## multiplicative, pre-distortion, r^i*mu^j for i=0,1,2 and j = 0,1,2
    ## plus an additive post-distortion in rp^i*rt^j for i=0,1,2 and j=0,1,2
    ## the multiplicative broadband is implemented as xi_bb = (1+bb)*xi
    ## so adding constants to a multiplicative bb leads to degeneracies 
    ## (the user is expected to know what she is doing)
    bb2 = mul pre r,mu 0:2:1 0:2:1
    bb3 = add pos rp,rt 0:2:1 0:2:1

Broadbands are stored in the output file as the group `data-set-name/broadband/broadband-name` in the output `h5` file where  `broadband-name` is constructed as: `BB-data-set-name add|mul pre|pos `

There's no limit in the number of additive and multiplicative broadbands, the user is responsible for degeneracies. @londumas @vserret @mblomqvist can you test?

### DIFF
--- a/py/picca/fitter2/chi2.py
+++ b/py/picca/fitter2/chi2.py
@@ -313,7 +313,8 @@ class chi2:
                 gbb = g.create_group("broadband")
                 for k,bbs in d.bb.items():
                     for bb in bbs:
-                        bband = gbb.create_dataset(bb.name, d.da.shape, dtype = "f")
+                        bband = gbb.create_dataset(bb.name, 
+                                d.da.shape, dtype = "f")
                         bband[...] = bb(d.r, d.mu, **self.best_fit.values)
         del self.best_fit.values['SB']
 
--- a/py/picca/fitter2/chi2.py
+++ b/py/picca/fitter2/chi2.py
@@ -309,6 +309,12 @@ class chi2:
             g.attrs['chi2'] = d.chi2(self.k, self.pk_lin, self.pksb_lin, self.best_fit.values)
             fit = g.create_dataset("fit", d.da.shape, dtype = "f")
             fit[...] = d.best_fit_model
+            if d.bb != None:
+                gbb = g.create_group("broadband")
+                for k,bbs in d.bb.items():
+                    for bb in bbs:
+                        bband = gbb.create_dataset(bb.name, d.da.shape, dtype = "f")
+                        bband[...] = bb(d.r, d.mu, **self.best_fit.values)
         del self.best_fit.values['SB']
 
         if hasattr(self, "fast_mc"):
--- a/py/picca/fitter2/chi2.py
+++ b/py/picca/fitter2/chi2.py
@@ -311,9 +311,9 @@ class chi2:
             fit[...] = d.best_fit_model
             if d.bb != None:
                 gbb = g.create_group("broadband")
-                for k,bbs in d.bb.items():
+                for bbs in d.bb.values():
                     for bb in bbs:
-                        bband = gbb.create_dataset(bb.name, 
+                        bband = gbb.create_dataset(bb.name,
                                 d.da.shape, dtype = "f")
                         bband[...] = bb(d.r, d.mu, **self.best_fit.values)
         del self.best_fit.values['SB']
--- a/py/picca/fitter2/data.py
+++ b/py/picca/fitter2/data.py
@@ -143,8 +143,8 @@ class data:
                 deg_mu_max = dic_bb['deg_mu_max']
                 ddeg_mu = dic_bb['ddeg_mu']
     
-                name = 'BB-{}-{} {} {}'.format(self.name,
-                        ibb,dic_bb['type'],dic_bb['pre'])
+                name = 'BB-{}-{} {} {} {}'.format(self.name,
+                        ibb,dic_bb['type'],dic_bb['pre'],dic_bb['rp_rt'])
 
                 bb_pars = {'{} ({},{})'.format(name,i,j):0\
                     for i in range(deg_r_min,deg_r_max+1,ddeg_r)\
@@ -157,7 +157,7 @@ class data:
                 bb = partial(xi.broadband, deg_r_min=deg_r_min,
                     deg_r_max=deg_r_max, ddeg_r=ddeg_r,
                     deg_mu_min=deg_mu_min, deg_mu_max=deg_mu_max,
-                    ddeg_mu=ddeg_mu, 
+                    ddeg_mu=ddeg_mu,rp_rt = dic_bb['rp_rt']=='rp,rt',
                     name=name)
                 bb.name = name
 
--- a/py/picca/fitter2/data.py
+++ b/py/picca/fitter2/data.py
@@ -94,11 +94,6 @@ class data:
         self.r = r
         self.mu = mu
 
-        self.par_names = dic_init['parameters']['values'].keys()
-        self.pars_init = dic_init['parameters']['values']
-        self.par_error = dic_init['parameters']['errors']
-        self.par_limit = dic_init['parameters']['limits']
-        self.par_fixed = dic_init['parameters']['fix']
 
         self.pk = pk.pk(getattr(pk, dic_init['model']['model-pk']))
         self.pk *= partial(getattr(pk,'G2'), dataset_name=self.name)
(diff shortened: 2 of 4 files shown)


py/picca/fitter2/chi2.py
```diff
@@ -309,6 +309,13 @@ def export(self):
             g.attrs['chi2'] = d.chi2(self.k, self.pk_lin, self.pksb_lin, self.best_fit.values)
             fit = g.create_dataset("fit", d.da.shape, dtype = "f")
             fit[...] = d.best_fit_model
+            if d.bb != None:
+                gbb = g.create_group("broadband")
+                for k,bbs in d.bb.items():
+                    for bb in bbs:
+                        bband = gbb.create_dataset(bb.name, 
+                                d.da.shape, dtype = "f")
+                        bband[...] = bb(d.r, d.mu, **self.best_fit.values)
```
londumas: @ngbusca, why do you need to use pointers here (`**`) when few lines above we don't.
Simply to understand,

py/picca/fitter2/data.py
```diff
@@ -133,6 +128,48 @@ def __init__(self,dic_init):
...
+                name = 'BB-{}-{} {} {} {}'.format(self.name,
+                        ibb,dic_bb['type'],dic_bb['pre'],dic_bb['rp_rt'])
+
+                bb_pars = {'{} ({},{})'.format(name,i,j):0\
+                    for i in range(deg_r_min,deg_r_max+1,ddeg_r)\
+                        for j in range(deg_mu_min, deg_mu_max+1, ddeg_mu)}
+        
+                for k,v in bb_pars.items():
+                    dic_init['parameters']['values'][k]=v
+                    dic_init['parameters']['errors']['error_'+k]=0.01
```
londumas: `dic_init['parameters']['errors'] = {'error_'+k:0.01 for k in bb_pars.keys()}` is more beautiful, but as you want.

py/picca/fitter2/data.py
```diff
@@ -133,6 +128,48 @@ def __init__(self,dic_init):
...
+    
+                name = 'BB-{}-{} {} {} {}'.format(self.name,
+                        ibb,dic_bb['type'],dic_bb['pre'],dic_bb['rp_rt'])
+
+                bb_pars = {'{} ({},{})'.format(name,i,j):0\
+                    for i in range(deg_r_min,deg_r_max+1,ddeg_r)\
+                        for j in range(deg_mu_min, deg_mu_max+1, ddeg_mu)}
+        
+                for k,v in bb_pars.items():
+                    dic_init['parameters']['values'][k]=v
```
londumas: `dic_init['parameters']['errors'] = {k:v for v in bb_pars.values()}` is more beautiful, but as you want.

py/picca/fitter2/data.py
```diff
@@ -254,9 +291,25 @@ def xi_model(self, k, pk_lin, pars):
 
         if self.xi_asy_model is not None:
             xi += self.xi_asy_model(self.r, self.mu, k, pk_lin, self.tracer1, self.tracer2, **pars)
+ 
+        ## pre-distortion broadband 
+        for bb in self.bb['pre-mul']:
+            xi *= (1+ bb(self.r, self.mu,**pars))
```
londumas: no need for extra parenthesis

py/picca/fitter2/data.py
```diff
@@ -254,9 +291,25 @@ def xi_model(self, k, pk_lin, pars):
...
+
+        ## pre-distortion additive
+        for bb in self.bb['pre-add']:
+            xi += bb(self.r, self.mu, **pars)
 
         xi = self.dm.dot(xi)
 
+        ## pos-distortion multiplicative
+        for bb in self.bb['pos-mul']:
+            xi *= (1+bb(self.r, self.mu, **pars))
```
londumas: no need for extra parenthesis

londumas:
@ngbusca, Looks good. I am going to test now and tell you if I manage to make it work.

py/picca/fitter2/chi2.py
ngbusca: @londumas these are not pointers, what this does is it passes the items of the dictionary as keywords to the `bb` function. 

See https://stackoverflow.com/questions/36901/what-does-double-star-asterisk-and-star-asterisk-do-for-parameters

py/picca/fitter2/data.py
ngbusca: agreed, I'll make the change

py/picca/fitter2/data.py
ngbusca: agreed. Also, comprehensions are somewhat more efficient than for loops.

py/picca/fitter2/data.py
ngbusca: indeed...

py/picca/fitter2/data.py
ngbusca: indeed...

ngbusca:
@londumas ok, let me know if it works. I'll make the other changes tomorrow.

londumas:
@ngbusca, it works.
However, I though I remembered that we didn't have to fit some of them but simply to solve a matrix problem. Do I remember incorrectly?

ngbusca:
@londumas the additive broadband could be also implemented as a linear problem as it was for fitter1. That would imply a significant gain in performance. I'll think about whether I can find a simple way to do it without changing the logic of the code too much.

ngbusca:
@mblomqvist can you check as well?

mblomqvist:
I tried to run a QSO-Lya fit with the broadband

[broadband]
bb1 = add pre r,mu 0:2:1 0:6:2

but get a RuntimeError

  File "/Users/mblomqvist/Source/picca/bin/fitter2", line 19, in <module>
    chi.minimize()
  File "/py/picca/fitter2/chi2.py", line 102, in minimize
    self.best_fit = self._minimize()
  File "/py/picca/fitter2/chi2.py", line 88, in _minimize
    mig_init = iminuit.Minuit(self,forced_parameters=self.par_names,errordef=1,**kwargs_init)
  File "iminuit/_libiminuit.pyx", line 292, in iminuit._libiminuit.Minuit.__init__ (iminuit/_libiminuit.cpp:3044)
  File "iminuit/_libiminuit.pyx", line 1322, in iminuit._libiminuit.Minuit._check_extra_args (iminuit/_libiminuit.cpp:20562)
RuntimeError: Cannot understand keyword fix_1+f. May be a typo?
The parameters are array(['BB-QSO LYA-0 add pre r,mu (0,0)',
       'BB-QSO LYA-0 add pre r,mu (0,2)',
       'BB-QSO LYA-0 add pre r,mu (0,4)',
       'BB-QSO LYA-0 add pre r,mu (0,6)',
       'BB-QSO LYA-0 add pre r,mu (1,0)',
       'BB-QSO LYA-0 add pre r,mu (1,2)',
       'BB-QSO LYA-0 add pre r,mu (1,4)',
       'BB-QSO LYA-0 add pre r,mu (1,6)',
       'BB-QSO LYA-0 add pre r,mu (2,0)',
       'BB-QSO LYA-0 add pre r,mu (2,2)',
       'BB-QSO LYA-0 add pre r,mu (2,4)',
       'BB-QSO LYA-0 add pre r,mu (2,6)'], dtype='<U31')

mblomqvist:
The error changes every time I retry, e.g.
RuntimeError: Cannot understand keyword fix_sigmaNL_per. May be a typo?
or
RuntimeError: Cannot understand keyword fix_sigma_velo_lorentz. May be a typo?

ngbusca:
@mblomqvist are you retrying without changing the .ini and you get different results each time?

Can you verify whether this branch runs without the `[broadband]`?

mblomqvist:
@ngbusca It runs normally without the [broadband]

mblomqvist:
and, yes, I retry to run it without changing the ini file and get different reporter errors.

ngbusca:
@mblomqvist I think I found the problem, can you retry? It should be fixed now.

py/picca/fitter2/data.py
ngbusca: actually what you propose doesn't work. Your solution would override the `dic_init['parameters']['errors']` dictionary. What I'm doing here is adding new `k,v` pairs.

mblomqvist:
It runs nicely and yields results in agreement with expectations.

PR #494, opened 2018-10-29
correct option name

londumas:
correct option name for `-best-obs`.

### DIFF
--- a/py/picca/io.py
+++ b/py/picca/io.py
@@ -554,9 +554,13 @@ def read_from_spplate(in_dir, thid, ra, dec, zqso, plate, mjd, fid, order, log=N
 
     for pm in platemjd:
         p,m = pm
-        spplate = in_dir + "/{0}/spPlate-{0}-{1}.fits".format(p.zfill(4),m)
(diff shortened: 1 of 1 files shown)


ngbusca:
What is the purpose of `zfill`?

londumas:
`zfill` adds zeros at the beginning if the size of the string is less than a giving number. ex:
```
a = '123'
a.zfill(4)
-> '0123'
```

londumas:
https://www.tutorialspoint.com/python/string_zfill.htm

ngbusca:
Thanks @londumas, I know what `zfill` does and I know how to use google, I am asking why you added it since it seems totally unrelated to "correct option name for -best-obs."

londumas:
Lol, sorry.
I did two corrections in one. I can do another PR if you want.
In order to read the DR7 data, I need to use `--best-obs` because there are no spAll.fits,
and I need to read plates that are less than 4 digits long.

ngbusca:
Ok, I see. No need for another PR, in the future please spend some time explaining what PR is for!

londumas:
@ngbusca, Yes sorry.

ngbusca:
merge when ready @londumas

PR #496, opened 2018-10-30
Read delta using io.read_deltas

londumas:
Read delta using io.read_deltas and avoid code duplication.


ngbusca:
@londumas thanks for this PR, it's quite a cleanup of the code! It would be great if @mblomqvist and @vserret could run a quick test on this branch before merging.

PR #502, opened 2018-11-20
Modified codd_zint to use id matrix as dmat if requested

jfarr03:
As discussed in the picca-users group email, this merge adds a `no-dmat` option to `bin/coadd_zint`, setting the distortion matrix to the identity instead of loading from file.

I am not sure wether the multiplication of the identity matrix by `wet_aux` in line 77 - `dm += sp.identity(nb.shape[0])*wet_aux[:,None]` - is necessary, but including it seemed closer to the standard procedure (line 74).

### DIFF
--- a/bin/coadd_zint
+++ b/bin/coadd_zint
@@ -44,7 +44,7 @@ if not args.no_dmat:
     dm = h[1]['DM'][:]*0
     h.close()
 else:
-    dm = sp.zeros((nb.shape[0],nb.shape[0]))
+    dm = sp.eye(nb.shape[0])
 
 for f in args.data:
     if not (os.path.isfile(f.replace('cf','dmat')) or args.no_dmat):
@@ -73,8 +73,7 @@ for f in args.data:
         h = fitsio.FITS(f.replace('cf','dmat'))
         dm += h[1]['DM'][:]*wet_aux[:,None]
         h.close()
-    else:
-        dm += sp.identity(nb.shape[0])*wet_aux[:,None]
+
 for p in da:
     w=we[p]>0
     da[p][w]/=we[p][w]
--- a/bin/coadd_zint
+++ b/bin/coadd_zint
@@ -17,6 +17,8 @@ parser.add_argument("--data",type=str,nargs="*",required=True,
 parser.add_argument("--out",type=str,required=True,
         help="output file")
 
+parser.add_argument("--no-dmat",action='store_true',default=False,required=False,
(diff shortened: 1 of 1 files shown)


bin/coadd_zint
```diff
@@ -37,12 +39,15 @@ h.close()
...
 
-h = fitsio.FITS(args.data[0].replace('cf','dmat'))
-dm = h[1]['DM'][:]*0
-h.close()
+if not args.no_dmat:
+    h = fitsio.FITS(args.data[0].replace('cf','dmat'))
+    dm = h[1]['DM'][:]*0
+    h.close()
+else:
+    dm = sp.zeros((nb.shape[0],nb.shape[0]))
```
londumas: dm = sp.eye(nb.shape[0])

bin/coadd_zint
```diff
@@ -64,9 +69,12 @@ for f in args.data:
...
 
-    h = fitsio.FITS(f.replace('cf','dmat'))
-    dm += h[1]['DM'][:]*wet_aux[:,None]
-
+    if not args.no_dmat:
+        h = fitsio.FITS(f.replace('cf','dmat'))
+        dm += h[1]['DM'][:]*wet_aux[:,None]
+        h.close()
+    else:
+        dm += sp.identity(nb.shape[0])*wet_aux[:,None]
```
londumas: remove these two lines.

londumas:
@jfarr03, Thank you for this PR. In the future, could you always add at least one reviewer?

bin/coadd_zint
jfarr03: Line 45 sets dm to an array of zeros if you do have a dmat, so surely line 47 should also set it to an array of zeros?

bin/coadd_zint
jfarr03: Sorry exactly which two lines are you referring to?

jfarr03:
@londumas sorry yes will do in future!

bin/coadd_zint
londumas: But if you remove lines 76 and 77, it would work perfectly.

bin/coadd_zint
londumas: remove lines 76 and 77

ngbusca:
@londumas has James made the requested modifications? Can we merge?

londumas:
@jfarr03, Thanks for the work.

PR #504, opened 2018-11-21
add gaussian-smoothed p(k) to study mocks

julienguy:
* add model function `pk.pk_kaiser_gauss_smoothing` that performs a simple gaussian smoothing; with 2 new parameters to fit `sigma_smooth_per` and `sigma_smooth_par`.
* it's the same functional form as `pk.pk_NL` but it's applied to the whole P(k), not the BAO part only.
 another approach would be to add an option to have pk.pk_NL applied to the whole P(k), I let the main developers evaluate this (and maybe discard this PR).

### DIFF
--- a/bin/coadd_zint
+++ b/bin/coadd_zint
@@ -1,13 +1,13 @@
 #!/usr/bin/env python
 
+from __future__ import print_function
 import os
 import fitsio
 import argparse
 from scipy.linalg import inv
 import scipy as sp
 
-from picca.utils import smooth_cov
-from picca.utils import print
+from picca.utils import smooth_cov, print
 
 parser = argparse.ArgumentParser()
 
--- a/bin/do_cf.py
+++ b/bin/do_cf.py
@@ -77,8 +77,8 @@ if __name__ == '__main__':
     parser.add_argument('--no-project', action='store_true', required=False,
         help='Do not project out continuum fitting modes')
 
-    parser.add_argument('--no-same-wavelength-pairs', action='store_true', required=False,
-        help='Reject pairs with same wavelength')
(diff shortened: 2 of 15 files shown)


ngbusca:
@julienguy thanks for implementing this. I think it's better to keep the NL model specific for the BAO smoothing, it's less likely that people will get confused. @londumas what do you think?

londumas:
@julienguy and @ngbusca, sorry for taking so long to look at this PR.
It works great on the mocks, but does not fix 100% the issue.
I changed a little bit the way it was implemented to be applied easily to other pk.

Fitting the stack of 10 mocks of `v4.0.0`, I get:
- `chi2 = 3105.30, nbins = 1590, npar = 6, proba=0` for the auto-correlation, when using the previous model where we leave free `per binsize` and `par binsize` 
- `chi2 = 2020.02, nbins = 1590, npar = 6, proba=0` for the auto-correlation, when using the new model, where we fix `per binsize` and `par binsize` and free `par_sigma_smooth` and `per_sigma_smooth`.
- `chi2 = 1699.50, nbins = 1590, npar = 9, proba=0.019`, for the auto-correlation, adding HCD, even though the DLA are not there.

wedge_0

wedge_1

wedge_2

wedge_3

ngbusca:
By the way, it might be useful to look at the DR11 mocks to help understanding what might come from picca and what from the mocks. A bunch of DR11 mocks in spPlate format are available at:

/project/projectdirs/desi/users/nbusca/dr11_mocks/M3_0_[0-1]/00[0-9]

Dans chaque répertoire M3_0_[0-1] il y a un mock_drq.fits avec les Z vrais 
et en suite dans chacun des 00[0-9] il y a 

spAll-*.fits et ????/spPlate-MJD-PLATE.fits

PR #505, opened 2018-12-03
Io for transmission files

Waelthus:
This is for reading in continuum mocks and fixing the continuum to be one everywhere. Called the mode desi_transmission and changed some parameters to check if desi is part of the mode (such that every mode containing desi is treated like desi data). The read-in is mostly copied over from the original desi readin, but with different file and field names and without coaddition of spectral arms.

### DIFF
--- a/bin/do_deltas.py
+++ b/bin/do_deltas.py
@@ -309,10 +309,10 @@ if __name__ == '__main__':
             data_fit_cont = pool.map(cont_fit, sp.array(list(data.values()))[sort] )
         else:
             data_fit_cont = sp.array(list(data.values()))[sort]
-            for i,p in enumerate(sorted(list(data.keys()))):
-                data_fit_cont[p].co=sp.ones(data_fit_cont[p].iv.shape)
-                data_fit_cont[p].p0=1
-                data_fit_cont[p].p1=0
+            for d in data_fit_cont:
+                d.co=sp.ones(data_fit_cont[p].iv.shape)
+                d.p0=1
+                d.p1=0
         for i, p in enumerate(sorted(list(data.keys()))):
             data[p] = data_fit_cont[i]
 
--- a/bin/do_deltas.py
+++ b/bin/do_deltas.py
@@ -310,7 +310,7 @@ if __name__ == '__main__':
         else:
             data_fit_cont = sp.array(list(data.values()))[sort]
             for d in data_fit_cont:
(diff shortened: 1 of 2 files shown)


londumas:
@Waelthus, I don't understand why you need to run `do_deltas.py` if you already know the
delta_transmission. Couldn't you convert your mocks into the input for `do_pk1d.py`, and compute that directly? That is how we do it for London and Saclay mocks: `py/picca/utils.py::desi_convert_transmission_to_delta_files`.

Waelthus:
Isn't the transmission still something different than the delta-field? Normally I'd still need to divide by the mean transmission and cut everything to the right wavelength ranges which I guess |py/picca/utils.py::desi_convert_transmission_to_delta_files does|.
I did not know about this routine before today, but I think it  might still be useful to have the option of using the exact same workflow independent of the details of input data.
I also started working on mocks (and the corresponding truths that include a quasar continuum) from the data challenge, where it might help to be able to do the full analysis consistently in both cases.
But in the end it's your call, I can also add the code needed for this to utils if that's better.

Am 03.12.2018 um 16:54 schrieb Hélion du Mas des Bourboux:

ngbusca:
I agree with @londumas. `do_deltas.py` is not meant to be a generic entry point to picca, but rather the script that drives continuum fitting. You might want to input a pure transmission field, but the expected behaviour of `do_deltas` is that it will fit a continuum just like for real data.

The functionality that you propose seems to be covered by `desi_convert_transmission_to_delta_files`. If it's not the case, please open an issue describing what you would need and we can take it from there.

Waelthus:
Fair enogh, I wouldn't want to duplicate functionality. I guess I expected do_deltas to be a generic entry point to get the delta field in some specified format for further processing from whatever input.

ngbusca:
@Waelthus if `desi_convert_transmission_to_delta_files` is not what you need, can you open an issue describing why and what you would need?

Waelthus:
Oh, the desi_convert_transmission_to_delta_files was exactly what I needed at the time I opened this pull request. I just didn't know of its existance. This PR basically added io routines such that in principle the job of desi_convert_transmission_to_delta_files could be done with do_deltas.

I do need to read in the spectra in truth files from the data challenge as well, those files do contain continua, and therefore I think I'd need the continuum fitting. I could add an issue for this, but I already have a preliminary version doing this based on the changes in this PR in another branch.

PR #514, opened 2018-12-12
added function to model HCD following Rogers et al. 2018

andreufont:
I was able to run fitter2 with this model, and I get L = 10 Mpc/h, but I get a worse chi2 than with the sinc function. I haven't been able to make plots for both models, but it would be interesting to compare where they disagree.

Is there anything extra I should do before merging?

### DIFF
--- a/py/picca/fitter2/pk.py
+++ b/py/picca/fitter2/pk.py
@@ -53,6 +53,17 @@ def pk_hcd(k, pk_lin, tracer1, tracer2, **kwargs):
     return pk
 
 def pk_hcd_Rogers2018(k, pk_lin, tracer1, tracer2, **kwargs):
+    """Models the effect of HCD systems with the Fourier transform
+       of a Lorentzian profile. Motivated by Rogers et al. (2018).
+
+    Args:
+		Same than pk_hcd
+
+    Returns:
+        Same than pk_hcd
+
+    """
+
     # alternative to pk_hcd with a physically motivated model (Rogers++ 2018)
 
     bias1, beta1, bias2, beta2 = bias_beta(kwargs, tracer1, tracer2)
--- a/py/picca/fitter2/pk.py
+++ b/py/picca/fitter2/pk.py
@@ -53,7 +53,7 @@ def pk_hcd(k, pk_lin, tracer1, tracer2, **kwargs):
     return pk
 
 def pk_hcd_Rogers2018(k, pk_lin, tracer1, tracer2, **kwargs):
-    """Models the effect of HCD systems with the Fourier transform
+    """Model the effect of HCD systems with the Fourier transform
        of a Lorentzian profile. Motivated by Rogers et al. (2018).
 
     Args:
@@ -64,8 +64,6 @@ def pk_hcd_Rogers2018(k, pk_lin, tracer1, tracer2, **kwargs):
 
     """
 
-    # alternative to pk_hcd with a physically motivated model (Rogers++ 2018)
-
     bias1, beta1, bias2, beta2 = bias_beta(kwargs, tracer1, tracer2)
 
     bias_hcd = kwargs["bias_hcd"]
@@ -73,11 +71,7 @@ def pk_hcd_Rogers2018(k, pk_lin, tracer1, tracer2, **kwargs):
     L0 = kwargs["L0_hcd"]
 
     kp = k*muk
-    # Lorentzian part
-    F_hcd = utils.sinc(kp*L0)
-    # Gaussian part (1 Mpc/h ~ 100 km/s)
-    b = 1.0
-    F_hcd = sp.exp(-L0*kp) * sp.exp(-0.5*kp*kp*b*b)
+    F_hcd = sp.exp(-L0*kp)
 
     bias_eff1 = (bias1 + bias_hcd*F_hcd)
     beta_eff1 = (bias1 * beta1 + bias_hcd*beta_hcd*F_hcd)/(bias1 + bias_hcd*F_hcd)
--- a/py/picca/fitter2/pk.py
+++ b/py/picca/fitter2/pk.py
@@ -52,6 +52,32 @@ def pk_hcd(k, pk_lin, tracer1, tracer2, **kwargs):
 
     return pk
 
+def pk_hcd_Rogers2018(k, pk_lin, tracer1, tracer2, **kwargs):
+    # alternative to pk_hcd with a physically motivated model (Rogers++ 2018)
+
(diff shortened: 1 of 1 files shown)


ngbusca:
@andreufont I thought `b` was a free parameter in Rogers' model? Wouldn't we want to leave it free even if it isn't?

py/picca/fitter2/pk.py
```diff
@@ -52,6 +52,32 @@ def pk_hcd(k, pk_lin, tracer1, tracer2, **kwargs):
 
     return pk
 
+def pk_hcd_Rogers2018(k, pk_lin, tracer1, tracer2, **kwargs):
+    # alternative to pk_hcd with a physically motivated model (Rogers++ 2018)
```
londumas: Please add docstrings like in https://github.com/igmhub/picca/blob/5c44d4ecfcae7c25e79b7e5ed84f6172e2cd287a/py/picca/utils.py#L141.
Add the equation number, the page and the arxiv id of the paper.

py/picca/fitter2/pk.py
```diff
@@ -52,6 +52,32 @@ def pk_hcd(k, pk_lin, tracer1, tracer2, **kwargs):
...
+
+    bias1, beta1, bias2, beta2 = bias_beta(kwargs, tracer1, tracer2)
+
+    bias_hcd = kwargs["bias_hcd"]
+    beta_hcd = kwargs["beta_hcd"]
+    L0 = kwargs["L0_hcd"]
+
+    kp = k*muk
+    # Lorentzian part
```
londumas: no need for comment

py/picca/fitter2/pk.py
```diff
@@ -52,6 +52,32 @@ def pk_hcd(k, pk_lin, tracer1, tracer2, **kwargs):
...
+    F_hcd = utils.sinc(kp*L0)
+    # Gaussian part (1 Mpc/h ~ 100 km/s)
```
londumas: no need for comment

py/picca/fitter2/pk.py
```diff
@@ -52,6 +52,32 @@ def pk_hcd(k, pk_lin, tracer1, tracer2, **kwargs):
...
+    b = 1.0
```
londumas: why isn't that free?

py/picca/fitter2/pk.py
```diff
@@ -52,6 +52,32 @@ def pk_hcd(k, pk_lin, tracer1, tracer2, **kwargs):
...
+    F_hcd = sp.exp(-L0*kp) * sp.exp(-0.5*kp*kp*b*b)
```
londumas: `(kp*b)**2`

py/picca/fitter2/pk.py
```diff
@@ -52,6 +52,32 @@ def pk_hcd(k, pk_lin, tracer1, tracer2, **kwargs):
...
+
+    bias_eff1 = (bias1 + bias_hcd*F_hcd)
```
londumas: useless parenthesis

py/picca/fitter2/pk.py
```diff
@@ -52,6 +52,32 @@ def pk_hcd(k, pk_lin, tracer1, tracer2, **kwargs):
...
+    beta_eff1 = (bias1 * beta1 + bias_hcd*beta_hcd*F_hcd)/(bias1 + bias_hcd*F_hcd)
+
+    bias_eff2 = (bias2 + bias_hcd*F_hcd)
```
londumas: useless parenthesis

py/picca/fitter2/pk.py
```diff
@@ -52,6 +52,32 @@ def pk_hcd(k, pk_lin, tracer1, tracer2, **kwargs):
...
+    beta_eff2 = (bias2 * beta2 + bias_hcd*beta_hcd*F_hcd)/(bias2 + bias_hcd*F_hcd)
+
+    pk = pk_lin*bias_eff1*bias_eff2*(1 + beta_eff1*muk**2)*(1 + beta_eff2*muk**2)
```
londumas: `1.+`

londumas:
@andreufont, it would be nice to have a plots of the two best fit models and data, on the four wedges.

py/picca/fitter2/pk.py
andreufont: None of the functions in this file have these docstrings...

py/picca/fitter2/pk.py
andreufont: no comment

py/picca/fitter2/pk.py
andreufont: (see reply to Nico below)

py/picca/fitter2/pk.py
andreufont: I added a docstring with what I know (I don't know the input/output format of pk_hcd).

ngbusca:
@andreufont I would leave `b` as a free parameter and fix it in the `. ini`

for the docstring, it's your function that makes assumptions on what `pk` or other args are and the docstring should reflext that. E. g. I guess in your case it should be a numpy array...

londumas:
@ngbusca, the new PR is perfect. Could you plot the effect on the fit?

andreufont:
@londumas - I'm trying to figure out how to use the h5 files, I think I'll figure it out before 2019 :-)

ngbusca:
@andreufont you can take a look at the `tutorials/QA_plots.ipynb`. If you will make the plots on a notebook, don't hesitate PRing them in!

andreufont:
> @andreufont you can take a look at the `tutorials/QA_plots.ipynb`. If you will make the plots on a notebook, don't hesitate PRing them in!

@ngbusca - yes, Helion pointed me to that, but it was very difficult to identify the relevant lines. He sent me a different shorter script, and after a while I manage to get the plots :-)

I'm running now a proper fitting with metals, and will post the results.

andreufont:
The degeneracies between L_HCD and beta_F make it very difficult to do these comparisons. I ended up generating plots with L_HCD fixed at different values based on the fits done by Keir Rogers on the simulated DLAs. For instance, L6 in the mocks means a Lorentzian model with L=6 Mpc/h, a reasonable value if the contaminants were dominated by small (or sub-) DLAs. 

wedge_hcd_0 0_0 5
wedge_hcd_0 5_0 8
wedge_hcd_0 8_0 95
wedge_hcd_0 95_1 0
wedge_hcd_zoom

andreufont:
In any case I think we are ready to merge.

londumas:
@andreufont, Thanks for the work.

Issue #517, opened 2018-12-17
redshift split in cross correlation


ngbusca:
The current implementation of the redshift split in the cross correlation, based on the redshift of qso-forest pairs, does not ensure that the different redshift bins do not share forests. This is potentially a problem since we know that significant cross-covariance is induced by Wick diagrams with two quasars and a single forest, and might introduce undesired correlations among redshift bins.

@mblomqvist has shown using subsamples that there's very little cross covariance in the high and low z samples in his DR14 analysis, so this is not an issue for the DR14 paper.

However, with increasing statistics for DR16 and DESI it would be better to get that fixed. This means that one would simply control the redshift splits by cutting on the redshift of the deltas.

@londumas @julienguy : this way of splitting solves the problem of subtracting the average delta if the cut is done at the `do_deltas.py` level.

It amounts to changing l.46 in xcf.py from:

```python
d.neighs = sp.array([q for q in neighs if (10**(d.ll[-1]- sp.log10(lambda_abs))-1 + q.zqso)/2. >= z_cut_min and (10**(d.ll[-1]- sp.log10(lambda_abs))-1 + q.zqso)/2. < z_cut_max])
```

to:

```python
d.neighs = neighs
```

and removing the option `z-cut...` from `do_xcf.py`

ngbusca:
Actually, there's no need to do the redshift split at the `do_deltas.py` level, since `do_xcf.py` subtracts the mean of deltas for the selected sample.

londumas:
@ngbusca, Then we would have different ways to do redshift bins for the cross and the auto. I am fine with this.
Then the selection in redshift could simply consist in removing from forests the deltas with a redshift outside of the given range.

Using a shuffle of the distribution of forests (branch  add_shuffle_distribution), also does the trick, but it might be long since we want randoms to be at lest 10 times more precision than data.

![xi_bin_z_seed_100](https://user-images.githubusercontent.com/3166436/50221764-81d97280-0396-11e9-93cd-a928add008d0.png)

londumas:
@ngbusca, We can't use `--z-min-obj`/`--z-max-obj` because we get large correlations between different redshift bins: ~+/-10%. Indeed the same forest can contribute to multiple redshift bins.
We really want a delta or even a forest to be in one of the redshift bins.

ngbusca:
[...]

PR #521, opened 2018-12-20
Add shuffle distribution

londumas:
Add shuffle distribution of objects.


londumas:
This is a way to solve https://github.com/igmhub/picca/issues/517 and to study the covariance matrix.

PR #533, opened 2019-01-17
We can now do combining fits with different values of bias_hcd

vserret:
The ini files are the same than before except if you want to have different bias_hcd for a combining fit. 
Let's say we want to fit LYA(LYA)-LYA(LYA) and LYA(LYA)-LYA(LYB). 
Instead of bias_hcd, you write bias_hcd_LYA(LYA)-LYA(LYA) in the ini LYA(LYA)-LYA(LYA) file and bias_hcd_LYA(LYA)-LYA(LYB) in the bias_hcd_LYA(LYA)-LYA(LYB) file.


londumas:
@vserret, looks good. Does it work on a fit to the data?

vserret:
Yes, it works perfectly

vserret:
@londumas , ready to be merge !

PR #535, opened 2019-01-17
compute unfold cf 

vserret:
Add modifications in the computation of symmetric cfs such as lya(lya)xlya(lyb). 
Standardly, we compute this cf with rp = abs(rp) 
But we can also have rp positive or negative (depending on the relative position between abs1 and abs2) by adding the "unfold-cf" option


londumas:
@vserret, looks good. Did you have a look if it worked on data?

vserret:
Yes, it seems to work well

vserret:
@londumas , ready to be merged !

londumas:
@vserret, thank you very much.

PR #540, opened 2019-01-25
Checking on diff reso coaddition

### LINKED ISSUE #325: do_deltas breaks when coadding observations
The new fields `diff` and `reso` are not updated when coadding observations via the function `__add__`.  For coadditions, the resulting delta field might have a different number of pixels than the initial delta field, which generates an inconsistency between the number of pixels in `diff/reso` and the rest `ll` etc.

@Cyeche we need your input to solve this problem: how should we coadd the diff and reso fields for an object with multiple spectra?

Waelthus:
closes #325
I changed the coaddition of forests such that reso and diff are coadded as well just doing inverse variance weighting. This will prevent code from crashing, but might not be the optimal solution, so we might want to change the way this is done at a later time
Also restructured the whole __add__ routine responsable for the coadding in the following way:

- lambda and ivar are treated as before
- all other fields to be coadded (atm flux, mmef, diff and reso) are fed into a dictionary which is then looped over, each is coadded with inverse variance weighting. 
- So for flux nothing should change as it was done this way before
- for mmef the coaddition was done with iv*mmef, but there was no renormalization by the new iv, which I think was erroneous, but please double check
- for diff and reso the same inverse variance weighted coaddition is performed now
- if any other quantity should be added later and coaddition should be performed the same way it just needs to be added to the dict

### DIFF
--- a/py/picca/data.py
+++ b/py/picca/data.py
@@ -224,7 +224,6 @@ class forest(qso):
 
         return self
 
-
     def mask(self,mask_obs,mask_RF):
         if not hasattr(self,'ll'):
             return
--- a/py/picca/data.py
+++ b/py/picca/data.py
@@ -120,8 +120,9 @@ class forest(qso):
         ## mmef is the mean expected flux fraction using the mock continuum
         if mmef is not None:
             mmef = mmef[w]
-        if diff is not None :
+        if diff is not None:
             diff=diff[w]
+        if reso is not None:
             reso=reso[w]
 
         ## rebin
@@ -134,8 +135,9 @@ class forest(qso):
         cciv = sp.bincount(bins,weights=iv)
         if mmef is not None:
             ccmmef = sp.bincount(bins, weights=iv*mmef)
-        if diff is not None :
+        if diff is not None:
             cdiff = sp.bincount(bins,weights=iv*diff)
+        if reso is not None:
             creso = sp.bincount(bins,weights=iv*reso)
 
         cfl[:len(ccfl)] += ccfl
@@ -150,8 +152,9 @@ class forest(qso):
         iv = civ[w]
         if mmef is not None:
             mmef = cmmef[w]/civ[w]
-        if diff is not None :
+        if diff is not None:
             diff = cdiff[w]/civ[w]
+        if reso is not None:
             reso = creso[w]/civ[w]
 
         ## Flux calibration correction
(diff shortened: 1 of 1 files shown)


py/picca/data.py
```diff
@@ -186,42 +189,42 @@ def __init__(self,ll,fl,iv,thid,ra,dec,zqso,plate,mjd,fid,order, diff=None,reso=
...
+        iv = sp.append(self.iv, d.iv)
 
+        if self.mmef is not None:
+            dic['mmef'] = sp.append(self.mmef, d.mmef)
+        if self.diff is not None:
+            dic['diff'] = sp.append(self.diff, d.diff)
+        if self.reso is not None:
+            dic['reso'] = sp.append(self.reso, d.reso)
+
+        bins = sp.floor((ll - forest.lmin) / forest.dll + 0.5).astype(int)
```
londumas: Let's not modify this line only for spaces

py/picca/data.py
```diff
@@ -186,42 +189,42 @@ def __init__(self,ll,fl,iv,thid,ra,dec,zqso,plate,mjd,fid,order, diff=None,reso=
...
+        cll = forest.lmin + sp.arange(bins.max() + 1) * forest.dll
```
londumas: Let's not modify this line only for spaces

py/picca/data.py
```diff
@@ -186,42 +189,42 @@ def __init__(self,ll,fl,iv,thid,ra,dec,zqso,plate,mjd,fid,order, diff=None,reso=
...
+        civ = sp.zeros(bins.max() + 1)
```
londumas: Let's not modify this line only for spaces

py/picca/data.py
```diff
@@ -186,42 +189,42 @@ def __init__(self,ll,fl,iv,thid,ra,dec,zqso,plate,mjd,fid,order, diff=None,reso=
...
+        cciv = sp.bincount(bins, weights=iv)
```
londumas: Let's not modify this line only for spaces

py/picca/data.py
```diff
@@ -186,42 +189,42 @@ def __init__(self,ll,fl,iv,thid,ra,dec,zqso,plate,mjd,fid,order, diff=None,reso=
...
+        civ[:len(cciv)] += cciv
```
londumas: Let's not modify this line only for spaces

py/picca/data.py
```diff
@@ -186,42 +189,42 @@ def __init__(self,ll,fl,iv,thid,ra,dec,zqso,plate,mjd,fid,order, diff=None,reso=
...
+        w = (civ > 0.)
```
londumas: Let's not modify this line only for spaces

py/picca/data.py
```diff
@@ -186,42 +189,42 @@ def __init__(self,ll,fl,iv,thid,ra,dec,zqso,plate,mjd,fid,order, diff=None,reso=
...
+
+        for k, v in dic.items():
+            cnew = sp.zeros(bins.max() + 1)
+            ccnew = sp.bincount(bins, weights=iv * v)
+            cnew[:len(ccnew)] += ccnew
+            setattr(self, k, cnew[w] / civ[w])
 
         return self
 
+
```
londumas: Let's not modify this line only for spaces

py/picca/data.py
Waelthus: I don't see where this comment belongs to anymore, I guess it's resolved now

py/picca/data.py
```diff
@@ -186,39 +189,38 @@ def __init__(self,ll,fl,iv,thid,ra,dec,zqso,plate,mjd,fid,order, diff=None,reso=
         self.mean_z = (sp.power(10.,ll[len(ll)-1])+sp.power(10.,ll[0]))/2./lam_lya -1.0
 
 
-    def __add__(self,d):
+    def __add__(self, d):
```
londumas: here

py/picca/data.py
```diff
@@ -186,39 +189,38 @@ def __init__(self,ll,fl,iv,thid,ra,dec,zqso,plate,mjd,fid,order, diff=None,reso=
...
 
-        if not hasattr(self,'ll') or not hasattr(d,'ll'):
+        if not hasattr(self, 'll') or not hasattr(d, 'll'):
```
londumas: here

py/picca/data.py
```diff
@@ -186,39 +189,38 @@ def __init__(self,ll,fl,iv,thid,ra,dec,zqso,plate,mjd,fid,order, diff=None,reso=
...
             return self
 
-        ll = sp.append(self.ll,d.ll)
-        fl = sp.append(self.fl,d.fl)
-        iv = sp.append(self.iv,d.iv)
-        mmef = None
+        dic = {}  # this should contain all quantities that are to be coadded with ivar weighting
+
+        ll = sp.append(self.ll, d.ll)
```
londumas: here

py/picca/data.py
```diff
@@ -186,39 +189,38 @@ def __init__(self,ll,fl,iv,thid,ra,dec,zqso,plate,mjd,fid,order, diff=None,reso=
...
+        dic['fl'] = sp.append(self.fl, d.fl)
+        iv = sp.append(self.iv, d.iv)
```
londumas: here

Waelthus:
Okay, I'll try to not add whitespaces to lines anymore (or to not let my editor do it semi-automatically). Is there a reason for not doing so? Given that you can change the diff views to ignore whitespace changes

londumas:
@Waelthus, thanks for all the work. It look perfect and it is nice to fix a long standing issue.
I did not know that it was possible to hide whitespace changes. The reason why we don't like it is that it adds a lot of changes to the code that are not necessary.
Thanks!

londumas:
@Waelthus, thank you. Merging.

PR #543, opened 2019-01-25
Update read_from_spec to use multiple observations

philippedavid:
Allow the reading of multiple observation when using spec files.

### DIFF
--- a/py/picca/data.py
+++ b/py/picca/data.py
@@ -187,10 +187,10 @@ class forest(qso):
 
 
     def __add__(self,d):
-        
+
         if not hasattr(self,'ll') or not hasattr(d,'ll'):
             return self
-        
+
         ll = sp.append(self.ll,d.ll)
         fl = sp.append(self.fl,d.fl)
         diff = sp.append(self.diff,d.diff)
@@ -199,7 +199,7 @@ class forest(qso):
         mmef = None
         if self.mmef is not None:
             mmef = sp.append(self.mmef,d.mmef)
-            
+
         bins = sp.floor((ll-forest.lmin)/forest.dll+0.5).astype(int)
         cll = forest.lmin + sp.arange(bins.max()+1)*forest.dll
         cfl = sp.zeros(bins.max()+1)
@@ -221,7 +221,7 @@ class forest(qso):
         if mmef is not None:
             cmmef[:len(ccmmef)] += ccmmef
         w = (civ>0.)
-        
+
         self.ll = cll[w]
         self.fl = cfl[w]/civ[w]
         self.iv = civ[w]
@@ -233,7 +233,7 @@ class forest(qso):
         err = 1./sp.sqrt(self.iv)
         SNR = self.fl/err
         self.mean_SNR = sp.mean(SNR)
-        
+
         return self
 
     def mask(self,mask_obs,mask_RF):
--- a/py/picca/data.py
+++ b/py/picca/data.py
@@ -224,6 +224,7 @@ class forest(qso):
         err = 1./sp.sqrt(self.iv)
         SNR = self.fl/err
         self.mean_SNR = sp.mean(SNR)
+        lam_lya = constants.absorber_IGM["LYA"]
         self.mean_z = (sp.power(10.,ll[len(ll)-1])+sp.power(10.,ll[0]))/2./lam_lya -1.0
 
         return self
--- a/py/picca/data.py
+++ b/py/picca/data.py
@@ -120,8 +120,9 @@ class forest(qso):
         ## mmef is the mean expected flux fraction using the mock continuum
         if mmef is not None:
             mmef = mmef[w]
-        if diff is not None :
+        if diff is not None:
             diff=diff[w]
+        if reso is not None:
             reso=reso[w]
 
         ## rebin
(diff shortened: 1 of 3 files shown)


philippedavid:
@Waelthus, Hello Michael,
I (working with Helion) have updated picca to use multiple observation by default when reading spec files. This should result in better 1D power spectra. Would you have a look and do a review?
Thanks,
Philippe

londumas:
@philippedavid, can you merge with master, fix the merging conflicts and then push?

py/picca/io.py
```diff
@@ -261,42 +260,111 @@ def read_data(in_dir,drq,mode,zmin = 2.1,zmax = 3.5,nspec=None,log=None,keep_bal
...
-    return pix_data
+    for meta in allmeta:
+        t = meta.thid
+        if not t in thids:
+            thids[t] = []
+        thids[t].append(meta)
+
+    print("reading {} thids".format(len(thids)))
 
+    for t in t_list:
```
Waelthus: this could be a `for t in thids.keys()` or even a `for t,tmeta in thids.items():` followed by `for meta in tmeta:`. Just to be independent of the t_list array which is elsewise not needed anymore

Waelthus: also: I think t_list should contain the same tid multiple times, whereas the data in thids is unique for each tid, so I guess this is actually an issue that might lead to double counting spectra.

philippedavid:
@Waelthus Hi Michael, thank you very much for your comments. I missed that double counting issue. I think I have resolved that specific issue with my latest update. The reason I use a list rather than just traversing through the thids.keys() is that maintaining the original order is necessary to pass the current unit test. From what I can tell, dictionary order is only guaranteed to match insertion order in Python 3.7 and later.

py/picca/io.py
```diff
@@ -340,7 +340,7 @@ def read_from_spec(in_dir,thid,ra,dec,zqso,plate,mjd,fid,order,mode,log=None,pk1
         for meta in thids[t]:
             r,d,z,p,m,f = meta.ra,meta.dec,meta.zqso,meta.plate,meta.mjd,meta.fid
             try:
-                fin = in_dir + "/{}/{}-{}-{}-{:04d}.fits".format(p,mode,p,m,fid)
```
Waelthus: nope, `str(fid)` shouldn't work as then it can't do the  int->str conversion from the format specifier

py/picca/io.py
```diff
@@ -261,42 +260,114 @@ def read_data(in_dir,drq,mode,zmin = 2.1,zmax = 3.5,nspec=None,log=None,keep_bal
...
+        plate = plate_spall[w]
+        mjd = mjd_spall[w]
+        fid = fid_spall[w]
+        spAll.close()
+
+    ## to simplify, use a list of all metadata
+    allmeta = []
+    ## Used to preserve original order and pass unit tests.
+    t_list = []
+    t_set = set()
```
Waelthus: this should work in principle, but down there you still iterate over t_list instead of t_set.
I think it's probably more readable if you just make `thids` an `collections.OrderedDict` and use `items()` or `keys()`

py/picca/data.py
```diff
@@ -221,6 +221,11 @@ def __add__(self,d):
             ccnew = sp.bincount(bins, weights=iv * v)
             cnew[:len(ccnew)] += ccnew
             setattr(self, k, cnew[w] / civ[w])
+        err = 1./sp.sqrt(self.iv)
+        SNR = self.fl/err
+        self.mean_SNR = sp.mean(SNR)
+        lam_lya = constants.absorber_IGM["LYA"]
+        self.mean_z = (sp.power(10.,ll[len(ll)-1])+sp.power(10.,ll[0]))/2./lam_lya -1.0
```
Waelthus: in between you added and removed the mean_reso, was there a reason for the removal?

py/picca/io.py
Waelthus: this is also the reason why travis failed below

py/picca/io.py
philippedavid: Sorry about the delay. I was thinking about that other data structure but I feel like it's better to not import libraries and just stick with the default Python data structures. I can use it if you would prefer. The point of using the set was to keep track of what thids were added to the list and prevent duplicates from being added. Sets have constant time lookup so I felt like the performance overhead would be minimal.

py/picca/io.py
philippedavid: Ok, I just fixed that. Thanks.

py/picca/io.py
Waelthus: The set is fine with me, but you still need to iterate over the set in the for loop of line 338.

py/picca/io.py
Waelthus: Note that for the set approach you need to change the iterator here to t_set

py/picca/io.py
philippedavid: I use the list to preserve order which the set doesn't. The set is just temporarily used to prevent duplicates thids from being added to the list.

py/picca/data.py
philippedavid: I was wrong to remove the mean_reso update. It's fixed now, thanks.

Issue #382, opened 2018-03-28
bias_<tracer> is a wrong name in fitter2


londumas:
In fitter2, `bias_i` is actually defined as the following:
![image](https://user-images.githubusercontent.com/3166436/38054875-69584c22-3295-11e8-8fec-ab4a3654ed68.png)
Either we should change the name of this parameter or change the definition in `py/picca/py/fitter2/utils.py:::bias_beta()`.
I have no strong opinions on what is better.
What do you think @ngbusca, @vserret, @mblomqvist?

vserret:
bias_eff_i ?

andreufont:
The problem with "velocity bias" is that it is used in other contexts in sister communities (galaxy clusetring). 

I would call it bias_eta_i , the name that has been used in the past in most Lyman alpha papers, where eta is the (dimensionless) line of sight velocity gradient.

On a related note, I would change the parameter (and not just the name) of the second bias in Lya fits. I would use bias_delta_i as the other parameter, instead of beta_i, unless someone figures out the combination of parameters that are less degenerate.

andreufont:
Another thing that we could do is to have fitter2 to write down also "derived parameters", similar to what CAMB / CLASS do. We can distinguish the parameters that actually go in the fits (and those should be whatever combination is less degenerate) to the parameters that we want to look at and have physical meaning, and in this case we can write 5 different combinations if we wanted.

ngbusca:
The problem is that this should be a parameter name that works for both discrete and absorption tracers. Is bias_eta also used in the galaxy clustering comunity? Same thing for bias_delta, we would have a `bias_delta_QSO` that we would typically fix to 1.

andreufont:
bias_eta and bias_delta are both valid parameters for any tracer. They represent the response of a large scale fluctuation in the tracer with respect to a large scale fluctuation in the density field (bias_delta) or in the velocity gradient (bias_eta). 

For galaxies (or quasars) we have:
 - bias_delta = bias_galaxy
 - bias_eta = 1 (although Hirata 2009 points out that there might be % deviations)

ngbusca:
Thanks @andreufont, it makes sense to me. If there are no objections from @londumas @vserret or @mblomqvist I'll PR the changes (within a week or two, no rush)

mblomqvist:
[...]

PR #545, opened 2019-01-28
change from bias to bias_eta

londumas:
change from `bias` to `bias_eta`. Address issue https://github.com/igmhub/picca/issues/382.


ngbusca:
@londumas you should check with @andreufont , bu I think `bias` should go to `bias_delta` not `bias_eta`.

andreufont:
No, I think @londumas is right. Picca was calling "bias" what we would usually call bias_eta, i.e., the response of delta_F to a velocity gradient. 

I believe bias_delta, the response to a large scale overdensity, is already called this way in picca?

londumas:
@ngbusca, I agree word for word. @andreufont is that good for you?

py/picca/test/data/config_xcf.ini
```diff
@@ -36,7 +36,7 @@ in tracer2 = SiIII(1207)
 
 [parameters]
 
-bias_QSO  = 109.8 1. None None free
+bias_eta_QSO  = 109.8 1. None None free
```
andreufont: why do you set bias_eta_QSO to such a large number? I guess this is a test that has no impact anywhere else, but I was a bit surprised to see it.

londumas:
@ngbusca, can you also do a review?

ngbusca:
@londumas you should go ahead, don't wait for me

londumas:
Thanks @andreufont and @ngbusca.

Issue #531, opened 2019-01-14
change names of executable


londumas:
Change names of executable in `/bin/*` to `/bin/picca_*`. This will allow not to be mixed with other executable from other projects.

londumas:
Fixed.

PR #547, opened 2019-01-28
Change names of executable

londumas:
Change names of executable, fix issue https://github.com/igmhub/picca/issues/531.
The names are now: `picca_*.py`. I removed the `do` when used and added `*.py` in order for Codacy tu see the file. @ngbusca could you tell me if that is what you had in mind?
I rand `picca_nersc_submit.py` and everything work.


londumas:
@Waelthus and @Cyeche, are the new names of the executable you use ok?

Waelthus:
I'm fine, just need to change the name in some scripts/notebooks, but that should be easy enough. So it's really just `do_` -> `picca_` and everything in bin that didn't have a `do_` before now has been prepended?

londumas:
@Waelthus, yes and also I added a *.py everywhere to allow Codacy to check all files.

londumas:
@vserret, I hve little time to do that now. What we can do is merge this and add a issue ticket with this idea. Anyway, it has very little effect since it acts only on the decimals. Is that ok for you?

vserret:
> @vserret, I hve little time to do that now. What we can do is merge this and add a issue ticket with this idea. Anyway, it has very little effect since it acts only on the decimals. Is that ok for you?

@londumas I think you did not reply on the good PR. Anyway, you are right. The effect would be very very small. We can keep that for latter.

PR #549, opened 2019-02-02
Correct small bug introduced by PR:543

londumas:
Correct small bug introduced by PR:543 in updating `mean_reso`. @philippedavid could you have a look also?


londumas:
@philippedavid and @Waelthus, thanks for the review.

Issue #510, opened 2018-12-04
Add missing absorption lines to `py/picca/constants.py` 


londumas:
Some absorption lines are missing from the py/picca/constants.py file.
For example FeIIλ1142.36, is present in the auto-correlation of Lya, in the form of an auto-correlation
FeIIλ1142.36 x FeIIλ1142.36 and a cross-correlation FeIIλ1142.36 x SIIIλ1190.21 for example.

londumas:
Here is a comparison of the stack of absorption at the DLA rest-frame in  Mas-Ribas et al 2017, and the one from picca. This allows to see where are the missing absorption lines.

londumas:
Other reference useful to get the rest-frame of the lines:
arXiv:1309.6768, astro-ph/0601279, arXiv:1610.02711, http://adsabs.harvard.edu/abs/1991ApJS...77..119M

londumas:
Setting this in the proper PR:
> @vserret, I hve little time to do that now. What we can do is merge this and add a issue ticket with this idea. Anyway, it has very little effect since it acts only on the decimals. Is that ok for you?

>@londumas I think you did not reply on the good PR. Anyway, you are right. The effect would be very very small. We can keep that for latter.

londumas:
Fixed

PR #551, opened 2019-02-04
Add missing absorption lines

londumas:
Add missing absorption lines. Fixes issue https://github.com/igmhub/picca/issues/510/.
In the following plots, I give the stack of the delta in the different forests at the rest-frame of the DLAs.
In red you have the definition of the lower bound of the Lya and Lyb forests, in green the Lyman serries and in black the absorption lines. I added missing lines up to the CIV forest.

stack_0

stack_1

stack_2


vserret:
@londumas Where did you fint the values for the atomic transitions ? 
I've found some different values on NIST

londumas:
@vserret, Thanks for the review. I used http://adsabs.harvard.edu/abs/1991ApJS...77..119M, the same paper I used for the other transitions and I checked it worked on the DLA stack.

vserret:
@londumas I think we have already had this conversation but would you mind to use NIST values ? I think these are the reference values

londumas:
Setting this in the proper PR:

> @vserret, I hve little time to do that now. What we can do is merge this and add a issue ticket with this idea. Anyway, it has very little effect since it acts only on the decimals. Is that ok for you?

> @londumas I think you did not reply on the good PR. Anyway, you are right. The effect would be very very small. We can keep that for latter.

londumas:
@vserret, then could you review the PR? Thanks a lot for the work.

PR #556, opened 2019-02-05
add dust correction

philippedavid:
Allows users to correct for dust with Schlegel map. Based on the "add_dust_map" branch by @londumas


londumas:
@philippedavid, thanks for the work.

PR #564, opened 2019-02-11
bin/picca_cf.py

jfarr03:
When reading deltas, the "from_image" option was not being included, where previously the options were included. This simply adds the options back in again.


londumas:
@jfarr03, Thanks. Waiting for checks and then I will merge.

londumas:
@jfarr03, thanks for the work. Merged.

Issue #571, opened 2019-03-13
read DLA catalog from a fits file


londumas:
In order to read DLA catalogs from DESI, let's only read this catalog from fits file, and add a conversion function in `py/picca/utils.py` to convert the ASCII files to a fits file. I think we should aim for the following header. Would that sound reasonable @alxogm and @solenechabanier?

for data
```
extname = DLACAT
data = RA, DEC, Z, ZQSO, NHI, THING_ID, DLAID, PLATE, MJD, FIBERID 
```

for mocks
```
extname = DLACAT
data = RA, DEC, Z_QSO_NO_RSD, Z_QSO_RSD, Z_DLA_NO_RSD, Z_DLA_RSD,
N_HI_DLA, MOCKID, DLAID
```

alxogm:
Hi @londumas will you use explicitly Z_QSO_NO_RSD, Z_QSO_RSD, Z_DLA_NO_RSD, Z_DLA_RSD at the picca level? The question  is if you need to distinguish between mock and data... I would've think this distinction is only at previous steps, before running picca on mocks, so there is  one DLACAT header that doesn't  know about being mock or data...

solenechabanier:
Yes that's ok for me !

londumas:
fixed

PR #572, opened 2019-03-13
Read DLA from fits file

londumas:
Read DLA from fits file, fix issue https://github.com/igmhub/picca/issues/571.
The idea is to read DLAs from a fits file, to be able to use DESI mocks.
To convert the BOSS and eBOSS catalogs, I added the function `utils.eBOSS_convert_DLA()` and
to convert the DESI mocks, I added the function `utils.desi_convert_DLA()`.
@alxogm, would that be good for you? Could you try running `utils.desi_convert_DLA()`? It is very similar to the already existing function `desi_from_ztarget_to_drq()` we use fro quasars.

### DIFF
--- a/py/picca/io.py
+++ b/py/picca/io.py
@@ -17,30 +17,35 @@ class metadata:
     pass
 
 def read_dlas(fdla):
-    f=open(os.path.expandvars(fdla))
-    dlas={}
-    nb_dla = 0
-    col_names=None
-    for l in f:
-        l = l.split()
-        if len(l)==0:continue
-        if l[0][0]=="#":continue
-        if l[0]=="ThingID":
-            col_names = l
-            continue
-        if l[0][0]=="-":continue
-        thid = int(l[col_names.index("ThingID")])
-        if thid not in dlas:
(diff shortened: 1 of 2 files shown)


alxogm:
> Read DLA from fits file, fix issue #571.
> The idea is to read DLAs from a fits file, to be able to use DESI mocks.
> To convert the BOSS and eBOSS catalogs, I added the function `utils.eBOSS_convert_DLA()` and
> to convert the DESI mocks, I added the function `utils.desi_convert_DLA()`.
> @alxogm, would that be good for you? Could you try running `utils.desi_convert_DLA()`? It is very similar to the already existing function `desi_from_ztarget_to_drq()` we use fro quasars.

Thanks Helion, I understand now, yes it seems it will work as I was thinking...  I'll test it as soon as I have access to cori.

londumas:
@alxogm, perfect.

alxogm:
OK.  I made a small test to run picca_deltas with the DLA catalog and everything went smoothly. I'll make a larger run, to see the result, but I think this is done. Thanks!

PR #573, opened 2019-03-14
Correct name xD2R1 to xR1D2, correct bug in reading xfiles

londumas:
Correct a small bug in `picca_export_co.py`, when running on the DLAxQSO cross-covariance, the bug was linked to the name of the files.


londumas:
@TEtourneau, could you do a simple review?

Issue #574, opened 2019-03-19
picca is cryptic is can not find any delta field


londumas:
picca is cryptic is can not find any delta field

londumas:
fix in PR https://github.com/igmhub/picca/pull/578

PR #578, opened 2019-03-19
Better verbose if no data

londumas:
Better verbose if no data.
Fix issue https://github.com/igmhub/picca/issues/574


londumas:
@Andrea-MG, thank you.

PR #580, opened 2019-03-21
Add outdir

londumas:
Add out directory to nersc submit plots. This allows to send a production from outside of the directory; useful when sending n>>1 jobs.


londumas:
fixes also issue https://github.com/igmhub/picca/issues/581

PR #583, opened 2019-03-23
import speed of light from scipy

londumas:
import speed of light from scipy, this allows, to have no hard codded speed of light.


londumas:
related to this ticket https://github.com/desihub/desisim/issues/477

PR #587, opened 2019-03-25
Add dr14 data

vserret:


londumas:
@vserret, could you add a description and also tell me why you add more than 1 file, I would rather have only one. At some point we would hit the repository size limit and also people will not understand if there is more than one file.

vserret:
I have written a readme following what as been done with dr12. 
I have added more than one file to do the same thing as before but I can remove the best fit results since they can be found in the papers.

londumas:
@vserret, Let's simply store the best fit parameters and the chi2 scan of the combined fit. This should fit in one h5py file. Why are there 16 files?

vserret:
I have reorganized the directory but there is only 4 new files for each paper, ie 8 new files. 
I think that would be better to use a text file for the chi2scan. It is easier for people to use it.

londumas:
@vserret, let's keep the directory structure by paper as it were before, and I would really prefere leaving a h5py files, with a simple example of reading it in the reame for example.

vserret:
Ok I will restore the old directory but I prefer the text format

londumas:
ok, and we agree on only giving the combined chi2scan?

vserret:
I agree that would be better but if we want to refer to one directory in each paper, it is better to have two files

londumas:
@vserret, Thank you very much.
Can you keep the same structure as in previous results?
See `https://github.com/igmhub/picca/tree/add_dr14_data/data/duMasdesBourbouxetal2017/fits`.
picca/data/duMasdesBourbouxetal2017/fits
with something like combined_stdFit and cross_alone_stdFit

vserret:
I don't think this is very usefull to a directory in an other directory. Except if we want to add somothing else, which I think it is not the case, I think this is better to have just one directory.

londumas:
I think it is nice to keep the same structure as before and to splitt
between auto alone and auto+cross combined

vserret:
ok. Done

londumas:
@vserret, Thank you very much for the work.
And thank you for the patience. :)

Issue #598, opened 2019-04-04
Add option to fit (ap,at) on the fulll shape of the correlation


londumas:
Add option to fit (ap,at) on the fulll shape of the correlation

londumas:
Done in the following branch: https://github.com/igmhub/picca/tree/full-shape
I don't think we want to merge this, however it is useful to test how biased the BAO parameters can be in a full shape measurement, how correlated with other parameters and how better the error could be.

```
noMet+noHCD model
 ||                      || \alpha_{\parallel}   || \alpha_{\perp}        || 
 || cf-z-0-10            || 1.015 +/- 0.023      || 0.981 +/- 0.035       || 1838.41 / (1590-6),  p = 8.2e-0600 || 
 || cf-z-0-10+fullShape  || 0.788 +/- 0.024      || 0.750 +/- 0.022       || 1765.75 / (1590-6),  p = 0.00089 ||
 || xcf-z-0-10           || 1.067 +/- 0.035      || 0.927 +/- 0.041       || 3208.73 / (3180-6),  p = 0.33 || 
 || xcf-z-0-10+fullShape || 1.016 +/- 0.021      || 1.018 +/- 0.026       || 3212.05 / (3180-6),  p = 0.31 ||


Full model
||                      || \alpha_{\parallel}   || \alpha_{\perp}       ||  ||
|| cf-z-0-10            || 1.024 +/- 0.027      || 0.972 +/- 0.036      || 1604.02 / (1590-13),  p = 0.31 || 
|| cf-z-0-10+fullShape  || 0.993 +/- 0.016      || 1.003 +/- 0.021      || 1604.68 / (1590-13),  p = 0.31 || 
|| xcf-z-0-10           || 1.070 +/- 0.035      || 0.924 +/- 0.041      || 3189.16 / (3180-10),  p = 0.40 || 
|| xcf-z-0-10+fullShape || 1.014 +/- 0.021      || 1.032 +/- 0.027      || 3192.10 / (3180-10),  p = 0.39 ||
```

![cf_z_0_10_w0_0 95_w1_1 0](https://user-images.githubusercontent.com/3166436/55643098-8aac5380-578f-11e9-971c-21f60c4b098a.png)
![cf_z_0_10_w0_0 8_w1_0 95](https://user-images.githubusercontent.com/3166436/55643099-8aac5380-578f-11e9-8102-1db0769f1ba3.png)
[...]

PR #602, opened 2019-04-09
Full shape

londumas:
Add option to fit the full shape correlation function. Fix issue https://github.com/igmhub/picca/issues/598


londumas:
@vserret, can you review the changes?

Issue #599, opened 2019-04-05
add requirements-extras.txt with extra non-essential packages


londumas:
add requirements-extras.txt with extra non-essential packages.
For example pk1d uses `from ROOT import TH1D, TFile, TTree, TProfile2D, TProfile`

londumas:
@Waelthus, what are the different packages that are needed to run the Pk1d analysis that are not already in `requirements.txt`?

londumas:
Adding this in branch https://github.com/igmhub/picca/tree/correlation_physical_model

Waelthus:
I don't think we need any additional packages for the 1d measurement right now, that are not included.
The code contains imports of the `ROOT` python bindings, but only if you decide to use root as an output format.
As `ROOT` cannot be installed with pip/conda anyway (at least to my knowledge), I don't think it should be in `requirements.txt` (not sure if that makes it suitable for the `requirements_extra.txt`). I could put a test into the `picca_Pk1d` that checks if `ROOT` is importable and if not spits out an error message (and either breaks or sets the output to `fits`)

londumas:
Fixed in branch https://github.com/igmhub/picca/compare/correlation_physical_model

Issue #524, opened 2018-12-21
peak / no_peak decomposition


londumas:
Add a script to decompose a give P(k) into a peak / no_peak.
This script could simply be integrated into pk2fits and produce the necessary file for fitter2 from the output of CAMB.
The explanation on how to do it is done in Kirkby et al 2013 (https://arxiv.org/pdf/1301.3456.pdf). Another way to do it is in https://nbodykit.readthedocs.io/en/latest/
```
from nbodykit.cosmology import *

# tu choisis une cosmo

cosmodep=Planck15

# tu peux la cloner et changer des paramètres au passage

cosmocl=cosmodep.clone(h=.676,Omega0_b=.048,Omega0_cdm=.26,n_s=.97)

# et normaliser sigma8 si tu veux…

cosmo=cosmocl.match(sigma8=.8)
redshift = 1.5

# ensuite tu calcules le spectre de puissance :    

P_linear = LinearPower(cosmo,redshift,transfer='CLASS') # ça linke CLASS

nowig = LinearPower(cosmo,redshift,transfer='NoWiggleEisensteinHu')
# ca fait le spectre de puissance sans baryons
# attention la parametrisation transfer='EisensteinHu' (avec wiggle)
# est trop ancienne, à ne pas utiliser
```

londumas:
Working on that in https://github.com/igmhub/picca/tree/correlation_physical_model

londumas:
Fixed in https://github.com/igmhub/picca/pull/603

PR #603, opened 2019-04-09
Correlation physical model

londumas:
This PR implement ticket https://github.com/igmhub/picca/issues/524, by producing the power-spectrum given the CAMB input config file, then split it into the wiggles and no-wiggles part.
Simply run `picca_compute_pk_pksb.py -i PlanckDR16.ini -o PlanckDR16.fits`.
I had to change the input config file since CAMB deals with neutrinos differently than it used to. I get exactly the same power-spectrum, which is good.
There are still some things to understand:
 - CAMB output `omega_lambda = 0.685351`, does not add up to 1 with `omega_matter = 0.31457`, even though `omega_k = 0`. The difference is `7.899999999994023e-05`, is that precision of single float? @andreufont and @ngbusca, do you know anything about that?
 - in their paper, @dkirkby uses the range `[50,86]` for their first side band, I think this is cosmology dependent because in Planck2015 cosmology, the BAO peak starts at `~85`, so I used `[50,82]` in the code.
 - I don't have the plateau of the wiggles P(k), it seems to be linked to the FFT pk->xi, then xiSB -> pkSB. Is there a way to conserve the large scales?
 - I don't exactly have the same PWiggles(k) at high k, I get a slope and not oscillations about zero.

I tested on DR14 auto and cross-correlations and I get exactly the same values for the BAO parameters at the `Delta alpha = 0.001` level but I have a bit of changes in the chi2.
```
 ||                      || \alpha_{\parallel}   || \alpha_{\perp}       ||                      || 
 ||                      ||                      ||                      ||                      || 
 || cf-z-0-10DR12        || 1.021 +/- 0.029      || 0.975 +/- 0.040      || 1841.67 / (1590-6),  p = 6.4e-0600 || 
 || cf-z-0-10DR16        || 1.021 +/- 0.029      || 0.974 +/- 0.040      || 1840.74 / (1590-6),  p = 6.9e-0600 || 
 || xcf-z-0-10DR12       || 1.079 +/- 0.040      || 0.921 +/- 0.044      || 3210.71 / (3180-6),  p = 0.32 || 
 || xcf-z-0-10DR16       || 1.080 +/- 0.040      || 0.922 +/- 0.044      || 3210.55 / (3180-6),  p = 0.32 || 
```

The plots show:
 - the correlation and the correlation-Side_band (no BAO)
 - its zoom at `r~85`
 - the P(k)wiggles
 - its zoom at high k

xi
xi_zoom
DR12_vs_DR16
DR12_vs_DR16_zoom

[edit] fix also issue https://github.com/igmhub/picca/issues/599


andreufont:
Quick comment about Omega_k. You should take into account radiation. If matter - radiation equality was at z=3000, and now we are looking at z=2, there should be a contribution of Omega_rad / Omega_m = 1/1000.

Neutrinos makes things a bit more complicated, but you get my point.

londumas:
@andreufont, very useful info. `Omega_m`, `Omega_k` and `Omega_l` are give at `z=0`. Neutrino are included inside `Omega_m` is given in CAMB output `Om_m (inc Om_u)      =  0.314570`. So do you think that the `7.899999999994023e-05` difference are `omega_r(photon)`?

andreufont:
Yes, at z=0 Omega_m should include *non-relativistic* neutrinos, and if you are using a "degenerate" model for massive neutrinos it means that all neutrinos contribute to Omega_m if you use sum m_nu = 0.06 eV. If you are using normal or inverted hierarchy, at least one or two neutrinos might still be relativistic at z=0 and therefore should not be included in Omega_m.

I'm not sure if you can find Omega_rad in CAMB, but what you can do is read the value of z_eq in your model. That is when omega_{CDM+baryons} = omega_{photons+neutrinos}. You know that the density of CDM + baryons evolve from z_eq to 0 following (1+z)^3, and photons like (1+z)^4. Neutrinos are more complicated, but you know the value of omega_nu at z=0 and you know that at very high redshift their density is relalted to the radiation density by a constant factor.

One you know the radiation density at z_eq it should be trivial to compute the radiation density now.  

Or you can take the value of the CMB temperature at z=0 and use the fact that it is a black body to compute the density of radiation, good luck with the units :-p

PR #609, opened 2019-04-26
Fix new fast_xcf

londumas:
fast_xcf has changed in PR https://github.com/igmhub/picca/pull/603, to input Dm(z) and Dc(z), this change was forgotten in `xcf1d`


londumas:
@TEtourneau, thanks for the review

Issue #618, opened 2019-05-13
dmrp not defined in picca_export_coadd_zint.py


TEtourneau:
```
picca_export_coadd_zint.py --data cf_z_0_2.2.fits cf_z_2.2_2.4.fits cf_z_2.4_2.6.fits cf_z_2.6_2.8.fits cf_z_2.8_3.0.fits cf_z_3.0_3.2.fits cf_z_3.2_10.fits --out cf_z_0_10-exp.fits

coadding healpix 1799 in file cf_z_3.2_10.fitsComputing cov...
smoothing 2499

Traceback (most recent call last):
  File "/global/homes/t/tetourne/.conda/envs/py27/bin/picca_export_coadd_zint.py", line 4, in <module>
    __import__('pkg_resources').run_script('picca==4.0', 'picca_export_coadd_zint.py')
  File "/global/homes/t/tetourne/.conda/envs/py27/lib/python2.7/site-packages/pkg_resources/__init__.py", line 744, in run_script
    self.require(requires)[0].run_script(script_name, ns)
  File "/global/homes/t/tetourne/.conda/envs/py27/lib/python2.7/site-packages/pkg_resources/__init__.py", line 1506, in run_script
    exec(script_code, namespace, namespace)
  File "/global/homes/t/tetourne/.conda/envs/py27/lib/python2.7/site-packages/picca-4.0-py2.7.egg/EGG-INFO/scripts/picca_export_coadd_zint.py", line 112, in <module>
    
NameError: name 'dmrp' is not defined
```
I tried to coadd all the auto-correlation functions in:
`/global/cscratch1/sd/tetourne/Out/debug_2560_2/picca-0.1`

The distortion matrices are also present in the directory. They have the right name, but the fits files only have one hdu.
It looks like the line 112:
`elif dmrp.size==rp.size:`
should not try to access the dmrp variable.

londumas:
Can you try the stack function in tutorials instead, and tell me if it
works ?

londumas:
@TEtourneau, what version of picca did you use? Are you up-to-date with master?
As you can see there: https://github.com/igmhub/picca/blob/849e55d0e633d622ab79861e83795327a8254c56/bin/picca_dmat.py#L220,
you should have two HDUs. 
```
hdumasde@cori07:~> desi_header /global/cscratch1/sd/tetourne/Out/debug_2560_2/picca-0.1/dmat_z_0_2.2.fits --info
no key, will dump header
[...]

PR #620, opened 2019-05-14
read data if no dmrp

londumas:
Fix issue https://github.com/igmhub/picca/issues/618, for backward compatibility.
@TEtourneau, can you test?

### DIFF
--- a/bin/picca_export_coadd_zint.py
+++ b/bin/picca_export_coadd_zint.py
@@ -109,7 +109,7 @@ if 'dmrp' in locals():
     dmrp /= nbdm
     dmrt /= nbdm
     dmz /= nbdm
-elif dmrp.size==rp.size:
+if ('dmrp' not in locals()) or (dmrp.size==rp.size):
     dmrp = rp.copy()
     dmrt = rt.copy()
     dmz = z.copy()


TEtourneau:
I tested the branch, it works. I also checked that the 2 coadd cf are equals (the one from the stack function in tutorial, and the one from the branch backward-comatibility-for-dmrt): it is, up to ~ 1e-18
Thanks for the fix

bin/picca_export_coadd_zint.py
```diff
@@ -109,7 +109,7 @@
     dmrp /= nbdm
     dmrt /= nbdm
     dmz /= nbdm
-elif dmrp.size==rp.size:
+if ('dmrp' not in locals()) or (dmrp.size==rp.size):
```
TEtourneau: Shouldn't that be an elif instead of if ?
If 'dmrp' is in locals(), and dmrp.size == rp.size, this will overwrite dmrp, dmrt, dmz

londumas:
@TEtourneau, thanks for the review and the check.

bin/picca_export_coadd_zint.py
londumas: We want to ovewrite dmrp if they have the same size has rp, since rp is computed on all the pairs in picca_cf, when dmrp is only computed with a fraction in picca_dmat.

PR #623, opened 2019-05-22
verbose ZWARNING

londumas:
be more verbose about the selection of spectra


londumas:
@TEtourneau, thanks for the review.

PR #625, opened 2019-05-23
Get cov from shuffle

londumas:
Get covariance and stack from shuffled realizations or from mocks


londumas:
@vserret, could you do a review when you have the time?

PR #626, opened 2019-05-24
Correct DeprecationWarning for pkg_resources

londumas:
Correct the following `DeprecationWarning` from `pkg_resources`
```
<HOME>/.local/lib/python3.6/site-packages/pkg_resources/__init__.py:1145:
DeprecationWarning: Use of .. or absolute path in a resource path is not allowed
and will raise exceptions in a future release.
```


londumas:
@TEtourneau, thanks for the review

PR #627, opened 2019-05-25
Adapt to new way iminuit=1.3.6 is verbose

londumas:
Adapt to new way iminuit=1.3.6 is verbose


londumas:
@TEtourneau, thanks for the review.

Issue #636, opened 2019-06-06
log10 in comment


londumas:
https://github.com/igmhub/picca/blob/8c65fb237a55b25ee63d9ab01a7164821947f874/bin/picca_deltas.py#L132

The fit is in log10 and not in log, correct the comment.

londumas:
fixed in PR https://github.com/igmhub/picca/pull/637

PR #637, opened 2019-06-07
Improve doc prints

londumas:
Improve doc prints, this will allow to grep easilly rejected forests.
Also fix ticket https://github.com/igmhub/picca/issues/636.

### DIFF
--- a/bin/picca_deltas.py
+++ b/bin/picca_deltas.py
@@ -296,6 +296,7 @@ if __name__ == '__main__':
         log.write("Found {} DLAs in forests\n".format(nb_dla_in_forest))
 
     ## cuts
+    log.write("INFO: Sample has {} forests\n".format(sp.sum([len(p) for p in data.values()])))
     for p in data.keys():
         l = []
         for d in data[p]:
--- a/bin/picca_deltas.py
+++ b/bin/picca_deltas.py
@@ -296,7 +296,7 @@ if __name__ == '__main__':
         log.write("Found {} DLAs in forests\n".format(nb_dla_in_forest))
 
     ## cuts
-    for p in list(data.keys()):
+    for p in data.keys():
         l = []
         for d in data[p]:
             if not hasattr(d,'ll') or len(d.ll) < args.npix_min:
@@ -317,6 +317,7 @@ if __name__ == '__main__':
         data[p][:] = l
         if len(data[p])==0:
             del data[p]
+    log.write("INFO: Sample has {} forests\n".format(sp.sum([len(p) for p in data.values()])))
 
     for p in data:
         for d in data[p]:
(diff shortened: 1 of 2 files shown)


bin/picca_deltas.py
```diff
@@ -296,27 +296,29 @@ def cont_fit(data):
...
                 continue
 
             if isnan((d.fl*d.iv).sum()):
-                log.write("{} nan found\n".format(d.thid))
+                log.write("INFO: Rejected {} due to nan found\n".format(d.thid))
                 continue
 
             if(args.use_constant_weight and (d.fl.mean()<=0.0 or d.mean_SNR<=1.0 )):
-                log.write("{} negative mean of too low SNR found\n".format(d.thid))
+                log.write("INFO: Rejected {} due to negative mean of too low SNR found\n".format(d.thid))
```
Waelthus: change this to "... or too low SNR ..."

bin/picca_deltas.py
```diff
@@ -296,27 +296,29 @@ def cont_fit(data):
         log.write("Found {} DLAs in forests\n".format(nb_dla_in_forest))
 
     ## cuts
-    for p in list(data.keys()):
+    log.write("INFO: Sample has {} forests\n".format(sp.sum([len(p) for p in data.values()])))
```
Waelthus: Maybe something like "Input sample has..."

bin/picca_deltas.py
```diff
@@ -296,27 +296,29 @@ def cont_fit(data):
...
                 continue
 
             l.append(d)
             log.write("{} {}-{}-{} accepted\n".format(d.thid,
-                d.plate,d.mjd,d.thid))
+                d.plate,d.mjd,d.fid))
         data[p][:] = l
         if len(data[p])==0:
             del data[p]
+    log.write("INFO: Sample has {} forests\n".format(sp.sum([len(p) for p in data.values()])))
```
Waelthus: Maybe something like "Accepted sample has..." (just to clarify the difference to the earlier line)

bin/picca_deltas.py
Waelthus: Maybe save the number to a variable so that subsequent similar lines can say "... {} remaining forests (of {} initial)" or something similar

bin/picca_deltas.py
```diff
@@ -400,17 +402,19 @@ def cont_fit(data):
...
-    for p in sorted(list(data.keys())):
+    for p in sorted(data.keys()):
         deltas[p] = [delta.from_forest(d,st,forest.var_lss,forest.eta,forest.fudge, args.use_mock_continuum) for d in data[p] if d.bad_cont is None]
         data_bad_cont = data_bad_cont + [d for d in data[p] if d.bad_cont is not None]
 
     for d in data_bad_cont:
-        log.write("rejected {} due to {}\n".format(d.thid,d.bad_cont))
+        log.write("INFO: Rejected {} due to {}\n".format(d.thid,d.bad_cont))
+
+    log.write("INFO: Sample has {} forests\n".format(sp.sum([len(p) for p in deltas.values()])))
```
Waelthus: see above

bin/picca_deltas.py
londumas: Done

bin/picca_deltas.py
londumas: Good idea, let's keep that in mind for when we have the time.

bin/picca_deltas.py
londumas: Good idea, let's keep that in mind for when we have the time.

londumas:
@Waelthus, thanks for the review. I applied everything except quoting the number of lost forests. I am a bit lazy, but let's keep this in mind for the future.

PR #639, opened 2019-06-10
fix computing angle max if not flat LCDM

londumas:
fix computing angle max if not flat LCDM, i.e. rt is then given by Dm, not by Dc.


londumas:
@TEtourneau, thanks.

Issue #640, opened 2019-06-10
iminuit.values changed structure


echaussidon:
The code raises an alerte : 
```
File "<HOME>/software/picca/bin/picca_fitter2.py", line 22, in <module>
    chi.fastMC()
  File "<HOME>/software/picca/py/picca/fitter2/chi2.py", line 217, in fastMC

    self.fiducial_values = self.best_fit.values.copy()

AttributeError: 'iminuit._libiminuit.ValueView' object has no attribute 'copy'
```

when I run picca_fitter2.py and I use Fast MC (activate in chi2.ini)

londumas:
Fixed in https://github.com/igmhub/picca/pull/641

PR #641, opened 2019-06-10
Correct the structure iminuit.value

londumas:
Correct the structure iminuit.value. Fix ticket https://github.com/igmhub/picca/issues/640


londumas:
@echaussidon, can you try the branch and review the PR?

PR #642, opened 2019-06-13
also set sigmaNL_par=0 for sideband in fiducial model of fast MC

julienguy:
Bug fix in fiducial model of fast MC due to the introduction of the new parameter sigmaNL_par
(need to set it to zero in the "side band" xi)


londumas:
@julienguy, thanks for catching that.

PR #643, opened 2019-06-13
remove entries in dic outside the loop

londumas:
remove entries in dic outside the loop, removing inside was causing the following exception:
`RuntimeError: dictionary changed size during iteration`


londumas:
@TEtourneau, thanks.

Issue #651, opened 2019-06-22
shuffle_distrib_forests not rearranging pixel objects


joselotl:
The function shuffle_distrib_forests in utils.py, shuffles the angles of the quasars but doesn't move them from their current healpix pixel assignation. As the functions to compute the correlations only search for objects in the corresponding neighboring pixels, this results in the correlations computed using this function having less pairs than they should have.

Here I include  a possible fix, although I suppose there are more elegant solutions.

Josue De Santiago

```

def shuffle_distrib_forests(obj,seed):
    '''Shuffle the distribution of forests by assiging the angular
        positions from another forest

    Args:
        obj (dic): Catalog of forests
        seed (int): seed for the given realization of the shuffle

    Returns:
        obj (dic): Catalog of forest
    '''

    print('INFO: Shuffling the forests angular position with seed {}'.format(seed))

    dic = {}
    lst_p = ['ra','dec','xcart','ycart','zcart','cosdec','thid']
    dic['pix'] = []
    for p in lst_p:
        dic[p] = []

    for pix, oss in obj.items():
        for o in oss:
            dic['pix'].append(pix)
            for p in lst_p:
                dic[p].append(getattr(o, p))

    sp.random.seed(seed)
    idx = sp.arange(len(dic['ra']))
    sp.random.shuffle(idx)

    i = 0
    data_shuffled = {}
    for oss in obj.values():
        for o in oss:
            for p in lst_p:
                setattr(o,p,dic[p][idx[i]])
            if not dic['pix'][idx[i]] in data_shuffled:
                data_shuffled[dic['pix'][idx[i]]]=[]
            data_shuffled[dic['pix'][idx[i]]].append(o)
            i += 1

    return data_shuffled
```

londumas:
@joselotl, nice catch! Sorry I made that bug, let me correct it.

londumas:
@joselotl, I think the simplest way is to reassign delta, weight, ... instead of the angular distribution. Would you agree?

londumas:
@joselotl, fixed in https://github.com/igmhub/picca/pull/652. Could you test it and tell me if it is ok?

joselotl:
It now works for me. Thank you! I didn't consider reassigning the deltas, I agree that it would have been easier

londumas:
Let's close it once the PR is merged. THanks for the work.

joselotl:
Ok, sorry. Should I comment in the fix?

londumas:
[...]

PR #652, opened 2019-06-24
Fix missing pairs from wrong assigned healpix

londumas:
Fix missing pairs from wrong assigned healpix, in shuffled.
Fix ticket https://github.com/igmhub/picca/issues/651

### DIFF
--- a/bin/picca_xcf.py
+++ b/bin/picca_xcf.py
@@ -178,7 +178,7 @@ if __name__ == '__main__':
                                 args.z_evol_obj, args.z_ref,cosmo)
 
     if not args.shuffle_distrib_obj_seed is None:
-        objs = utils.shuffle_distrib_obj(objs,args.shuffle_distrib_obj_seed)
+        objs = utils.shuffle_distrib_forests(objs,args.shuffle_distrib_obj_seed)
     if not args.shuffle_distrib_forest_seed is None:
         xcf.dels = utils.shuffle_distrib_forests(xcf.dels,
             args.shuffle_distrib_forest_seed)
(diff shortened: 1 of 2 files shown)


joselotl:
The fix works for me. It assigns correctly the healpix pixels.

londumas:
@joselotl, perfect. Can you do the review?

londumas:
Here is the ratio of number of pairs, for the cross-correlation and its equivalent shuffled: 
`--shuffle-distrib-obj-seed`
cross

```
import matplotlib.pyplot as plt
import fitsio
h1 = fitsio.FITS('../../../Correlations/xcf_z_0_10.fits.gz')
h2 = fitsio.FITS('xcf_z_0_10.fits.gz')
plt.plot(h2[1]['NB'][:]/h1[1]['NB'][:])
plt.xlabel('bin index')
plt.ylabel('nb ration')
plt.title('cross')
plt.grid()
plt.show()
```

londumas:
@joselotl, can you post a similar plot for your 3pcf? I am computing right now for the auto.

londumas:
@joselotl, so now, thanks to you, the mean ratio of number of pairs is indeed correct and equal to 1, however because of the 3d clustering of quasars, it is not exactly 1 for all bins. Hence, what I said by email still stands: to get the estimated variance, in each bin, you have to correct by the ratio of number of pairs.

joselotl:
Here is the ratio for the 3pcf. In this case corresponding to the number of triangles per bin. It is also close to one but not excactly

image

londumas:
Simply to keep track, here is the ratio for the cross, when using `--shuffle-distrib-forest-seed`.

cross_different

londumas:
@joselotl, thanks for catching this issue.

PR #654, opened 2019-06-25
Select a random forest and then compute all

londumas:
Select a random forest and then compute all, instead of selecting a random sample of the pairs.
Will break travis tests, because change the *dmat.


londumas:
Won't fix.

PR #660, opened 2019-07-10
Allow util.desi_convert_transmission_to_delta_files to read both tran…

TEtourneau:
…smission*.fits and transmission*.fits.gz files

Warning: transmission files need to be either all *.fits or *.fits.gz files.


londumas:
@TEtourneau, thank you for catching that. Merging as soon as the tests finished running.

londumas:
@TEtourneau, thank you very much.

PR #661, opened 2019-07-12
Added logfile and echo command in the nersc picca submit scripts

TEtourneau:
This PR adds logfile to each .batch file produce by picca_nersc_submit.
It also added an print of the command run in each .batch file, in order to know which command was run (in case the batch scripts were modified).
It also allows to run the submit.sh script without necessarily being in the submit.sh directory location.


londumas:
@TEtourneau, the log are already inside the slurm. Won't fix

PR #666, opened 2019-08-02
nHI weighted dla

julienguy:
Add option to compute the nHI weighted DLA LYA cross-correlation.
This is for a specific analysis so I am not sure we really need it in master, but on the other hand it's not very invasive.


londumas:
@julienguy, are you sue we need to merge this PR? This look like a complexity for a single test cast. As you wish. If yes, please fix the branch so it runs on travis

julianbautista:
@julienguy, I know this is quite old. Can we close this PR and come back at this later? 
One needs to update the branch with the new master version.

PR #669, opened 2019-09-19
update the documentation

echaussidon:
just add one line in my documentation


londumas:
@echaussidon, thank you for the PR

PR #674, opened 2019-12-05
Remove non-ASCII characters in headers of sky masks

andreufont:


andreufont:
Hi @ngbusca , @londumas , @julienguy , @vserret  - do you know who has access to merge pull requests?

This is a very small and trivial change that would make our life easier if it was approved, otherwise python complains about non-ASCII characters in headers.

Thanks!

londumas:
@andreufont, done

andreufont:
Thanks @londumas!

Should we add new people in the list of authorised users? I would be happy to be added.

londumas:
@andreufont, this is done. https://github.com/orgs/igmhub/people. You are now a owner.
However use that with caution, e.g. each PR should be precedeed by a bug report to explain what is the issue, and there should always have at least a few senteces explaining the PR. Also each PR should be reviviwed before.

andreufont:
Great. Thank you very much, I'll be cautious with PRs.

PR #675, opened 2019-12-31
Redo dr12

andreufont:
I've done a couple of minor changes to be able to redo DR12 and DR14 analyses:
 - add option to keep objects with non-positive THING_ID, since this was done in du Mas des Bourboux et al. (2017).
 - ignore DLAs in DLA catalog if there is no quasar with same THING_ID in quasar catalog (19 objects in one of the DR14 DLA catalogs are not in the DR14 quasar catalog).

### DIFF
--- a/bin/picca_deltas.py
+++ b/bin/picca_deltas.py
@@ -57,7 +57,7 @@ if __name__ == '__main__':
         help='Upper limit on quasar redshift from drq')
 
     parser.add_argument('--keep-zero-thid',action='store_true',required=False,
-        help='Do not quasars with THING_ID < 1')
+        help='Do not reject quasars with THING_ID < 1')
 
     parser.add_argument('--keep-bal',action='store_true',required=False,
         help='Do not reject BALs in drq')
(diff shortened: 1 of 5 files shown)


andreufont:
I'm not passing the codacy test, because with the addition of one extra argument the function is now considered to be "too complex". 
I agree that shorter argument lists are better, but it is not the fault of this very simple commit, so I'd say we should go ahead and accept the pull request as is.

londumas:
@andreufont, I am in favor of not merging this issue.
It is adding too many if and conditions to a already complicated io.
Even Codacy said enough with too much complication.
You can keep this branch unmerged for record.
But you decide.

andreufont:
I think it would nice to have the option to reproduce the DR12 results. I noticed that there is an old branch named keep_zero_thid hta was trying to do the same, but it was never merged either.

andreufont:
Ok, I've been convinced by Helion. We'll keep this branch alive but unmerged, in case others want to reproduce DR12 results with Picca.

PR #678, opened 2020-01-30
Forecast

andreufont:
This short PR adds a feature in the fitter to compute something very similar to the Fisher forecasts. It runs through the FastMC code, but it does not add randomness. This was used to generate Fig 14 of the eBOSS DR16 paper, and I think it will be very useful in DESI as well, to test the impact of different survey strategies BAO uncertainties, without needing to generate multiple FastMC to compute an average error.

### DIFF
--- a/py/picca/fitter2/chi2.py
+++ b/py/picca/fitter2/chi2.py
@@ -224,7 +224,7 @@ class chi2:
             d.co = s*d.co
             d.ico = d.ico/s
             # no need to compute Cholesky when computing forecast
-            if not self.forecast:
+            if not self.forecast_mc:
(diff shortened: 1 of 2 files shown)


alxogm:
@andreufont  changes look good. My only suggestion would be that we add a config file, maybe to the tutorials, as reference of how to use it? The options to use the FastMC where confusing (I think), so we can use this to add some documentation... Thanks.

andreufont:
Hi @alxogm - I opened an issue about documentation of FastMC, but I think it makes sense to separate it from this one.

@andreicuceu will soon submit a large PR about the MCMC in fitter2 and I'd like to update thr documentation after these changes

Issue #680, opened 2020-02-19
DeprecationWarning for scipy


londumas:
Looking at the latest Travis run (https://travis-ci.org/igmhub/picca/jobs/647506891?utm_medium=notification&utm_source=github_status),
there are a lot of depreciation warning.
It would be good to tackle them before it is too late.
This can be fixed in a few grep.

```144: DeprecationWarning: scipy.arange is deprecated and will be removed in SciPy 2.0.0, use numpy.arange instead
190: DeprecationWarning: scipy.arange is deprecated and will be removed in SciPy 2.0.0, use numpy.arange instead
193: DeprecationWarning: scipy.unique is deprecated and will be removed in SciPy 2.0.0, use numpy.unique instead
199: DeprecationWarning: scipy.zeros is deprecated and will be removed in SciPy 2.0.0, use numpy.zeros instead
201: DeprecationWarning: scipy.zeros is deprecated and will be removed in SciPy 2.0.0, use numpy.zeros instead
251: DeprecationWarning: scipy.array_equal is deprecated and will be removed in SciPy 2.0.0, use numpy.array_equal instead
255: DeprecationWarning: scipy.absolute is deprecated and will be removed in SciPy 2.0.0, use numpy.absolute instead
256: DeprecationWarning: scipy.allclose is deprecated and will be removed in SciPy 2.0.0, use numpy.allclose instead
271: DeprecationWarning: scipy.array_equal is deprecated and will be removed in SciPy 2.0.0, use numpy.array_equal instead
271: DeprecationWarning: scipy.logical_not is deprecated and will be removed in SciPy 2.0.0, use numpy.logical_not instead
277: DeprecationWarning: scipy.allclose is deprecated and will be removed in SciPy 2.0.0, use numpy.allclose instead
281: DeprecationWarning: scipy.array_equal is deprecated and will be removed in SciPy 2.0.0, use numpy.array_equal instead
283: DeprecationWarning: scipy.allclose is deprecated and will be removed in SciPy 2.0.0, use numpy.allclose instead
```

londumas:
Fixed in PR https://github.com/igmhub/picca/pull/681

iprafols:
Fitter and fitter2 still have scipy dependencies

julianbautista:
@iprafols this is solved now right? Can we close this?

iprafols:
Yes, we can close it

PR #681, opened 2020-02-27
Fix codacy and depreciation

londumas:
This heavy PR is actually benign, it fixes:
 - Using the latest version of fitsio does not build, so uses "fitsio==1.0.5"
 - fix trailing space from codacy
 - fix unused "junk" from codacy
 - fix depreciation warning from scipy to numpy

I know this is heavy for review, but @andreufont could you take few minutes to skim it,
the unitest works.
Fixes ticket https://github.com/igmhub/picca/issues/680


andreufont:
Hi @londumas - thank you for taking care of all these changes!

I'm adding @andreicuceu here, he is the new Picca wizzard, and he'll know how to review this much better than I do.

One question though: why did you use "npy" instead of "np" for numpy? I think everyone else in the world uses "np.array" instead of "npy.array", so we need to have a good reason not to follow this! Is it because we use "np" for other variable names? If so, I'd vote for changing that variable name instead.

Andrei could take care of reverting to "np" while he reviews the code.

londumas:
@andreufont, indeed `np` is used for the number of bins in the transverse direction.
I agree we can change that, and go back to `import numpy as np`. I take care of this.
Thus @andreicuceu, waits before doing a review. I will do the changes before.

andreufont:
Thanks @londumas !

londumas:
I will leave you @andreicuceu or @andreufont taking care of it finally.

andreufont:
Hi @andreicuceu - could you take care of review and changing npy for np? Thanks!

andreufont:
Hi @londumas , @andreicuceu - I changed (np,nt) for (npb,ntb), for "number of parallel bins" and "number of transverse bins". I checked that the tests passed, and then changed (npy) for (np). It seems to work fine. I also reviewed the changes, but they were really trivial. I think we can merge.

londumas:
@andreufont, I will have a look this evening.

tutorials/data/picca_plot_chi2_surface.py
```diff
@@ -1,6 +1,7 @@
 #!/usr/bin/env python
 
 import os
+import numpy as npy
```
andreicuceu: Numpy is still npy in this file

tutorials/picca_export_stacked_correlation.py
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/env python
 
+import numpy as npy
```
andreicuceu: Numpy is still npy in this file

andreufont:
Let us wait for the DESI meeting to finish before merging though, since we have many hack sessions using Picca and we do not want to get people confused.

andreufont:
Hi @londumas , @andreicuceu - should we merge this tomorrow?

londumas:
@andreufont and @andreicuceu, this will make all picca no longer backward compatible because you changed `--np` and `--nt`. I would keep them as is simply in the argparser, but indeed change them to `ntb` and `npb` everywhere else.
We should really try not to break backward compatibility.
But do as you prefer

andreufont:
Hi @londumas - you are totally right, we should keep backwards compatibility! 

I will use --np and --nt in the program options, and then internally use nbp and nbt.

andreufont:
Hi @londumas - I believe I have fixed the backwards compatibility. I'll run an end-to-end analysis at NERSC to make sure this is the case.

andreufont:
I will not be able to finish the tests at NERSC for a couple of days, but I noticed that I still get warnings related to NUMBA. I guess this will need a new PR.

andreufont:
picca_deltas, picca_cf, picca_xcf, picca_dmat, picca_xdmat, picca_metal_dmat, picca_metal_xdmat and picca_fitter2 ran well. I just checked that I reproduced results from master branch.

tutorials/picca_export_stacked_correlation.py
```diff
@@ -75,20 +76,20 @@
 
     ### same header
     for i in range(nbData):
-        for k in ['NSIDE','HLPXSCHM','NP','NT','RTMAX','RPMIN','RPMAX']:
+        for k in ['NSIDE','HLPXSCHM','NPB','NTB','RTMAX','RPMIN','RPMAX']:
             assert data[i][k]==data[0][k]
```
londumas: @andreufont, I do not think this is correct, since in the fits files the keys are "NP" and "NT".

tutorials/picca_export_stacked_correlation.py
```diff
@@ -56,7 +57,7 @@
         hep = sp.array(h[2]['HEALPID'][:])
         data[i] = {'RP':rp, 'RT':rt, 'Z':z, 'NB':nb, 'DA':da, 'WE':we,'HEALPID':hep,
             'NSIDE':nside, 'HLPXSCHM':scheme,
-            'NP':np, 'NT':nt, 'RTMAX':rt_max, 'RPMIN':rp_min, 'RPMAX':rp_max}
+            'NPB':npb, 'NTB':ntb, 'RTMAX':rt_max, 'RPMIN':rp_min, 'RPMAX':rp_max}
```
londumas: same here

tutorials/picca_export_stacked_correlation.py
```diff
@@ -129,8 +130,8 @@
         final['CO'] = hh[1]['CO'][:]
         hh.close()
     else:
-        binSizeP = (final['RPMAX']-final['RPMIN']) / final['NP']
-        binSizeT = (final['RTMAX']-0.) / final['NT']
+        binSizeP = (final['RPMAX']-final['RPMIN']) / final['NPB']
+        binSizeT = (final['RTMAX']-0.) / final['NTB']
```
londumas: same here

tutorials/picca_export_stacked_correlation.py
```diff
@@ -171,7 +172,7 @@
 
     h = fitsio.FITS(args.out,'rw',clobber=True)
     head = {}
-    for k in ['NT','NP','RTMAX','RPMIN','RPMAX']:
+    for k in ['NTB','NPB','RTMAX','RPMIN','RPMAX']:
```
londumas: same here

londumas:
@andreufont, Perfect except two files where you refer to non existing NPB and NTB in fits file header

andreufont:
Thanks @londumas - I always forget to check the tutorials folder. It should be done now, ready to merge!

andreufont:
Hi @andreicuceu - the merging is blocked because you requested changes (I think), but I believe these are already addressed, right?

londumas:
@andreufont, thank you very much.

Issue #685, opened 2020-03-24
Fix warning in forecast of fitter2 


andreufont:
When I run fitter2 I get the following warning: 

```
/global/homes/f/font/.conda/envs/picca_plots/lib/python3.8/site-packages/picca-4.0-py3.8.egg/picca/fitter2/chi2.py:51:
SyntaxWarning: "is not" with a literal. Did you mean "!="?
```

It's seems trivial to fix, but since I know that @andreicuceu is working on this, I thought it was better to leave it for now.

londumas:
Indeed this line has little sense. `!=` would be better.
https://github.com/igmhub/picca/blob/58ab96090a2ca90b851276ad1fbacc3622c82f2b/py/picca/fitter2/chi2.py#L51

andreufont:
Definitely, but I'll leave @andreicuceu implement the change.

andreufont:
@andreicuceu - we can close this issue, right?

Are there other branches / issues that we can also remove? It would be good to do a bit of tidying up.

andreufont:
Hi @andreicuceu - this was fixed, right?

andreicuceu:
This was fixed

PR #686, opened 2020-03-25
Fitter mcmc

andreicuceu:
Most of this is new code added in separate files that will not interfere with the fitter2 functionality. Everything that the fitter was doing so far works exactly the same. 
I added three new classes:
1. fitter2 - meant to act as an interface for the fitter. It stores all the data, and chi2 objects. Functionality almost identical to calling the picca_fitter2.py script. This makes it easy to work with the models and the fitter from notebooks, or you can inherit and upgrade the functionality.
2. fitter2_mpi - inherits from fitter2 and extends with parallel functionality using MPI. In particular, the Scan and fastMC functions can now be run in parallel.
3. sampler -  Interface with the PolyChord Nested Sampler. This can be accessed through the fitter2_mpi class.

For supporting these classes, I added a new method in data.py that computes a normalized log-likelihood using the already available chi2 method. I also added a new gaussian prior that returns a log-likelihood (instead of chi2). 

The new libraries used are mpi4py and pypolychord. As these are not required for single-core runs, I created a new file: requirements-mpi.txt, so we can separate the installation of dependencies. 

I also added a fitter2 folder in tutorials with an example chi2.ini file that details all the config for the functionality that was introduced. I will also add example config files for the models, but as those are not relevant to this pull request they will be done separately. 

Also addressed Issue #685.


tutorials/fitter2/chi2_example.ini
```diff
@@ -0,0 +1,123 @@
...
+filename = [path to output file].h5
+
+##########
+
+# Scan config. Start, Stop, Steps
+# One or two parameters only
+# Optional
+[chi2 scan]
+ap = .95 1.15 15
+at = .85 1.15 15
```
andreufont: I think this example is great, but it might take a while to run when not using MPI. 

I wonder whether we should have also an example file that can test the different functionalities, but that would run in a couple of minutes. Doing a scan on a single parameter, doing 10 FastMC mocks, and so on. 

We can open an issue about this, no need to fix everything at the same time.

py/picca/fitter2/control_mpi.py
```diff
@@ -0,0 +1,332 @@
...
+
+        mpi_comm = MPI.COMM_WORLD
+        cpu_rank = mpi_comm.Get_rank()
+        num_cpus = mpi_comm.Get_size()
+
+        # Seed is incremented from the input value by the CPU #
+        np.random.seed(self.chi2.seedfast_mc + cpu_rank)
+        nfast_mc = self.chi2.nfast_mc // num_cpus
+        if self.chi2.nfast_mc % num_cpus != 0:
+            nfast_mc += 1
```
andreufont: Why do we do this here? Please document.

py/picca/fitter2/control_mpi.py
```diff
@@ -0,0 +1,332 @@
...
+            for d in self.chi2.data:
+                if p in d.par_names:
+                    d.pars_init[p] = self.chi2.fidfast_mc[p]
+                    d.par_fixed['fix_'+p] = self.chi2.fixfast_mc['fix_'+p]
+
+        # Compute fiducial model
+        self.chi2.fiducial_values['SB'] = False
+        for d in self.chi2.data:
+            d.fiducial_model = self.chi2.fiducial_values['bao_amp'] \
+                * d.xi_model(self.chi2.k, self.chi2.pk_lin-self.chi2.pksb_lin, self.chi2.fiducial_values)
```
andreufont: Please add a couple of comments here about how you are computing the model

py/picca/fitter2/control_mpi.py
```diff
@@ -0,0 +1,332 @@
...
+            store_d_par_error = store_data_pars[d.name]['error']
+            store_d_par_fixed = store_data_pars[d.name]['fixed']
+            for name in d.pars_init.keys():
+                d.pars_init[name] = store_d_pars_init[name]
+            for name in d.par_error.keys():
+                d.par_error[name] = store_d_par_error[name]
+            for name in d.par_fixed.keys():
+                d.par_fixed[name] = store_d_par_fixed[name]
+
+    def _mpi_fastMC(self):
```
andreufont: Some of the code below is duplicated from chi2.fastMC(), I wonder whether there could be a way to minimise duplication. If we were to change the behaviour of one of these functions, we would now need to synchronise them by hand...

py/picca/fitter2/sampler.py
```diff
@@ -0,0 +1,115 @@
...
+        self.zeff = dic_init['data sets']['zeff']
+        self.data = dic_init['data sets']['data']
+        self.par_names = np.unique([name for d in self.data for name in d.par_names])
+        self.outfile = os.path.expandvars(dic_init['outfile'])
+        self.polychord_setup = dic_init['Polychord']
+
+        self.k = dic_init['fiducial']['k']
+        self.pk_lin = dic_init['fiducial']['pk']
+        self.pksb_lin = dic_init['fiducial']['pksb']
+        self.full_shape = dic_init['fiducial']['full-shape']
```
andreufont: Please add a comment or two about pksb and full-shape

py/picca/fitter2/sampler.py
```diff
@@ -0,0 +1,115 @@
...
+
+    def log_lik(self, pars):
+        ''' Compute log likelihood for all datasets and add them '''
+        pars['SB'] = False
```
andreufont: why do we need to do this about SB? Please add a comment.

tutorials/fitter2/chi2_example.ini
andreicuceu: I reduced the number of runs, so hopefully, it should be doable in a few minutes now. However, it all depends on how many free parameters we have.

py/picca/fitter2/control_mpi.py
andreicuceu: Removing duplication in this case would require a major code reshuffling, as there is no structure that handles the model computation independently. This problem is not localized to these changes either. The original code for fastMC in chi2.py also reuses code from data.py.

andreicuceu:
I addressed all the comments. I still need to add a readme for the fitter two with some run instructions.

py/picca/fitter2/control_mpi.py
andreufont: How do you know that adding one set is enough / better? Is the goal to make sure that all CPUs run exactly the same number of cases? If so, shouldn't the number of extra cases to add depend on the specific case?

andreufont:
Hi @andreicuceu - everything looks great. Minor comments:
 - I added an extra comment about num_cpus and n_fastmc. This is not a show stopper, I'm just curious. 
 - I would start a new Issue about duplicated code, and mention that you'll address this in fitter3.
 - A bit of documentation on running the fitter would be great, but this could be in a new Issue / PR.

py/picca/fitter2/control_mpi.py
andreicuceu: So let me give an example, which I hope is clearer. Say we have 5 cores and we ask for 13 mocks. In this case, there are three options of what we can do:
1. Compute 3 * 5 = 15 mocks as the smallest multiple larger than the number asked.
2. Compute 2 * 5 = 10 mocks as the largest multiple smaller than the number asked.
3. Compute 13 mocks, but for the last set 2 CPUs are idle.

Number 3 is much harder to implement, and wastes computing time. I think number 2 is not ideal because the user wants 13 mocks, so why compute less than asked? Which leaves number one as the best option that I see. And this is the one I did.

Of course, ideally, this shouldn't happen and the number of mocks should be a multiple of the number of CPUs. But I don't want to stop the run with an error if it isn't as it may have queued for a long time to start running.

andreufont: I agree with should do option (1), I just didn't understand the code. 

I got it now, I didn't realyze the "nfast_mc" was actually "nfast_mc_per_processor". I thought it was "updated_nfast_mc". 

Everything makes sense now, sorry for the noise.

PR #687, opened 2020-03-26
Blind analysis

iprafols:
Added blind templates to use with the fitter. 
In picca/constants.py, blind Omega matter is now read from template file


py/picca/constants.py
```diff
@@ -18,7 +18,14 @@ def __init__(self,Om,Ok=0.,Or=0.,wl=-1.,H0=100.,unblind=False):
...
             print("WARNING: The analysis is blinded ! The value of Om is not {} !".format(Om))
-            Om = 0.3  # prov: this should be read from a file
+            # blind test small
+            file = "fitter2/models/DR16_blind_test_small/DR16_blind_test_small.fits"
+            # blind test large
+            #file = "fitter2/models/DR16_blind_test_small/DR16_blind_test_large.fits"
+            # load Om
+            hdu = fits.open("DR16_om0{}.fits".format(aux))
+            Om = hdu[1].header["OM"]
+            hdu.close()
```
TEtourneau: I think you didn't define the variable "aux"
Are you loading any fits file library ?
The syntax looks like astropy.io.fits. I think we should aim to use only one in picca, which is fitsio : 
import fitsio
hdu = fitsio.FITS(file)
Om = hdu[1].read_header()['OM']
hdu.close()

iprafols: I added the import of fitsio at the top and changed to reading with fitsio

iprafols: but there seems to be an error when loading the file

iprafols: there seems to be an error when loading the file, though

TEtourneau: what is the error you get ?

iprafols: OSError: FITSIO status = 104: could not open the named file
failed to find or open the following file: (ffopen)
fitter2/models/DR16_blind_test_small/DR16_blind_test_small.fits

iprafols: I guess the issue is that it is not pointing to a path relative to constants.py

TEtourneau: Yes, I think you need to do it the same way it's done for the PlanckDR16.fits file (the P(k) used in the fitter2)

iprafols: Is it not done by using a configuration file? anyway  found a method to load using the relative path (see the last commit)

TEtourneau: You have an example here :
https://github.com/igmhub/picca/blob/master/py/picca/fitter2/parser.py#L28
(using ressource_filename)

iprafols: changes to load it like this

PR #689, opened 2020-04-02
Pblic data dr16

londumas:
This is near ready commit to add the public data of the DR16 results.
Of course I do not put the data yet, they are on the svn, but I updated the readme.
@andreufont coud you do  review ?
@andreicuceu could you tell me what files are needed for COSMO-MC ?
I think it would be nice to add them there. What do you think ?

### DIFF
--- a/data/README.md
+++ b/data/README.md
@@ -31,12 +31,12 @@ data/deSainteAgatheetal2019/combined_stdFit/combined_stdFit.ap.at.scan.dat
 In ![duMasdesBourbouxetal2020](/data/duMasdesBourbouxetal2020/):
 
 *   ![auto\_full\_stdFit](/data/duMasdesBourbouxetal2020/auto_full_stdFit/): fit results of the combined:
-        * auto-correlation of Lya absorption in Lya region (Lya(Lya)xLya(Lya))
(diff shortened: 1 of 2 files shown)


londumas:
@andreufont, to do the review the best is to have a look at what the new redme looks like.

andreufont:
Hi @londumas - I took a look at the README file, and it looks great. I also tested the plotting script for DR14, the instructions are very clear.

andreufont:
Yes, it would be great to add an update for both CosmoMC and MontePython. I believe these are done by contacting the developers and providing them with the new files, I'm not sure we can do it here.

londumas:
@andreufont, do you know who uses `MontePython`? I know for CosmoMC.

andreufont:
Hi @londumas - do you know anyone who has used our CosmoMC module? I know that at least the group by Julien Lesgourgues has used the the MontePython module, for thier studies of H_0 from BAO+BBN. 

MontePython is python, while CosmoMC is fortran, and specially those that use CLASS instead of CAMB tend to use MontePython, I believe.

andreufont:
Also, it is very easy to do both, Andrei / Pablo / James know how to do it, so I would just do both.

londumas:
@andreufont, I just send an email to that person.

andreufont:
Yes, you are right, now we use it internally in eBOSS. Good call!
________________________________

Issue #683, opened 2020-03-24
Update function calls from SciPy to silence warnings


andreufont:
On a fresh new installation of Picca at NERSC, [following the instructions here](https://desi.lbl.gov/trac/wiki/LymanAlphaWG/Picca_at_NERSC), I get several warnings from scipy (see below for full message). 

We should go through the code and update the calls to scipy and use numpy instead. @ngbusca , @londumas, is there any reason not to do that and stick to the scipy calls, even when there are warnings? 

```
/global/homes/f/font/.conda/envs/picca_plots/lib/python3.8/site-packages/picca-4.0-py3.8.egg/EGG-INFO/scripts/picca_deltas.py:242: DeprecationWarning: scipy.asarray is deprecated and will be removed in SciPy 2.0.0, use numpy.asarray instead
  usr_mask_obs    = sp.log10(sp.asarray(usr_mask_obs))
/global/homes/f/font/.conda/envs/picca_plots/lib/python3.8/site-packages/picca-4.0-py3.8.egg/EGG-INFO/scripts/picca_deltas.py:242: DeprecationWarning: scipy.log10 is deprecated and will be removed in SciPy 2.0.0, use numpy.lib.scimath.log10 instead
  usr_mask_obs    = sp.log10(sp.asarray(usr_mask_obs))
/global/homes/f/font/.conda/envs/picca_plots/lib/python3.8/site-packages/picca-4.0-py3.8.egg/EGG-INFO/scripts/picca_deltas.py:243: DeprecationWarning: scipy.asarray is deprecated and will be removed in SciPy 2.0.0, use numpy.asarray instead
  usr_mask_RF     = sp.log10(sp.asarray(usr_mask_RF))
/global/homes/f/font/.conda/envs/picca_plots/lib/python3.8/site-packages/picca-4.0-py3.8.egg/EGG-INFO/scripts/picca_deltas.py:243: DeprecationWarning: scipy.log10 is deprecated and will be removed in SciPy 2.0.0, use numpy.lib.scimath.log10 instead
  usr_mask_RF     = sp.log10(sp.asarray(usr_mask_RF))
/global/homes/f/font/.conda/envs/picca_plots/lib/python3.8/site-packages/picca-4.0-py3.8.egg/EGG-INFO/scripts/picca_deltas.py:244: DeprecationWarning: scipy.asarray is deprecated and will be removed in SciPy 2.0.0, use numpy.asarray instead
[...]

PR #694, opened 2020-04-13
Update scipy warning

londumas:
Starts fixing https://github.com/igmhub/picca/issues/683, only on warning seen in tests, not others.
Indeed `grep "sp\." -r *` still gives some entries, that either are not triggering warnings or are not tested.

### DIFF
--- a/py/picca/utils.py
+++ b/py/picca/utils.py
@@ -498,8 +498,8 @@ def desi_convert_transmission_to_delta_files(zcat,outdir,indir=None,infiles=None
     print('INFO: Found {} files'.format(fi.size))
 
     ### Stack the transmission
-    lmin = np.lib.scimath.log10(lObs_min)
-    lmax = np.lib.scimath.log10(lObs_max)
+    lmin = np.log10(lObs_min)
+    lmax = np.log10(lObs_max)
     nstack = int((lmax-lmin)/dll)+1
     T_stack = np.zeros(nstack)
     n_stack = np.zeros(nstack)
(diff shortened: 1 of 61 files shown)


py/picca/utils.py
```diff
@@ -727,9 +727,9 @@ def unred(wave, ebv, R_V=3.1, LMC2=False, AVGLMC=False):
 
     # Compute optical portion of A(lambda)/E(B-V) curve
     # using cubic spline anchored in UV, optical, and IR
-    xsplopir = sp.concatenate(([0],10000.0/sp.array([26500.0,12200.0,6000.0,5470.0,4670.0,4110.0])))
-    ysplir = sp.array([0.0,0.26469,0.82925])*R_V/3.1
-    ysplop = sp.array((sp.polyval([-4.22809e-01, 1.00270, 2.13572e-04][::-1],R_V ),
+    xsplopir = sp.concatenate(([0],10000.0/np.array([26500.0,12200.0,6000.0,5470.0,4670.0,4110.0])))
```
andreufont, line 730: should we use numpy concatenate?

londumas: This is one of the example that no warning were trigered by the tests. So either it is not tested, or scipy does not give a warning. I would say that it is un safe to change if it is not tested. What do you think ?

andreufont: I'd probably leave it for now. May be we can start an Issue about removing Scipy when not necessary, but this is not a priority if they don't cause warnings.

londumas:
@andreufont, thanks for the review. I just merged.

PR #696, opened 2020-04-13
update packages to latest version

londumas:
update packages to latest version


londumas:
@andreufont, this is a very simple review that allows to test if there are some version conditions for dependencies.

andreufont:
I'm not sure how to review this PR, since I don't quite know how the requirements files work. May be @andreicuceu would be a better reviewer.

londumas:
@andreufont, I simply ran a `pip install ... --upgrade`, and looked if the test are still runnig well. It they are not, then I report the error in picca, if it is linked to picca or in the related when it is the case. This allows to get the latest packages and see where we get incompabilities or depreciation warning.

andreufont:
Ok, so you did this, and checked that Picca passes all the tests when using the list above. Sounds good, then, we can probably merge.

Issue #591, opened 2019-04-01
latest version of setuptools (40.8.0) does not build


londumas:
Latest version of setuptools (40.8.0) does not build, and fails the Travis tests.
It gives the following error:
```
Traceback (most recent call last):
  File "/home/travis/virtualenv/python2.7.14/bin/picca_deltas.py", line 4, in <module>
    __import__('pkg_resources').run_script('picca==4.0', 'picca_deltas.py')
  File "/home/travis/virtualenv/python2.7.14/lib/python2.7/site-packages/pkg_resources/__init__.py", line 666, in run_script
    self.require(requires)[0].run_script(script_name, ns)
  File "/home/travis/virtualenv/python2.7.14/lib/python2.7/site-packages/pkg_resources/__init__.py", line 1437, in run_script
    .format(**locals()),
pkg_resources.ResolutionError: Script 'scripts/picca_deltas.py' not found in metadata at '/home/travis/build/igmhub/picca/py/picca.egg-info'
```
and
```
Traceback (most recent call last):
  File "/home/travis/virtualenv/python3.6.3/bin/picca_deltas.py", line 4, in <module>
    __import__('pkg_resources').run_script('picca==4.0', 'picca_deltas.py')
  File "/home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages/pkg_resources/__init__.py", line 666, in run_script
    self.require(requires)[0].run_script(script_name, ns)
  File "/home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages/pkg_resources/__init__.py", line 1437, in run_script
    .format(**locals()),
pkg_resources.ResolutionError: Script 'scripts/picca_deltas.py' not found in metadata at '/home/travis/build/igmhub/picca/py/picca.egg-info'

```

londumas:
Working on it in branch https://github.com/igmhub/picca/tree/use_latest_setuptools

londumas:
"De guerre lasse", I asked for some help there: https://stackoverflow.com/questions/55733130/python-setup-py-install-does-not-work-with-latest-setuptools-no-scripts-in

londumas:
Fixed in PR https://github.com/igmhub/picca/pull/698.
This was a very tough one to fix, the clue was in the following depreciation warning.
```
WARNING: Testing via this command is deprecated and will be removed in a future version. Users looking for a generic test entry point independent of test runner are encouraged to use tox.
```

[...]

PR #698, opened 2020-04-13
Use latest setuptools

londumas:
Finally fix issue https://github.com/igmhub/picca/issues/591.
The description is in https://github.com/igmhub/picca/issues/591 and in https://stackoverflow.com/questions/55733130/python-setup-py-install-does-not-work-with-latest-setuptools-no-scripts-in/61193119#61193119.
For short we now run the unitest within .travis using `pytest` instead of `python setup test`.


londumas:
@andreufont, could you do a simple quick review. I does not look like much but it allows us to update from using a very very old version of a package to the latest one.

londumas:
@andreufont, packaging is not the most interesting thing to do indeed. Thanks for the review.

PR #702, opened 2020-04-17
Adding of desiminisv mode to compute deltas with DESI mini SV data

corentinravoux:
picca_deltas.py is not modified, only io.py with a new mode.


londumas:
@corentinravoux, thanks for the work.
Could you explain why the existing routine `read_from_desi` does not work? Your code looks a lot like `read_from_desi`, in what aspect does it differ?

corentinravoux:
Hi Hélion,
DESI mini-SV data are not in the final format of DESI data. The main difference comes from the fact that in mini-SV spectra are classified by tiles and do not use healpix at the moment. read_from_desi and my function are very similar but they take into account the differences in the data.

londumas:
@corentinravoux, thanks for the answer.

andreufont:
Hi @corentinravoux , what is the status of this? Are you still waiting for a review? If so, I could take a look at it.

Waelthus:
I added in some code to allow:
* andes read-in (arms not co-added, whereas for miniSV arms were co-added)
* read-in of zbest-files (with some selection applied automatically)
* an utility script to select quasars from zbest files and merge those to a kind of catalogue (if tihs is used in picca_deltas selection is turned off automatically except for redshift cuts).

Currently this branch is coadding only data on the same tile for the desiminisv branch, I guess we could make this behaviour a keyword argument and default to coadding by healpix or targetid.

As Helion mentioned the changes between the read in for minisv and for normal desi are not super large, I could clean this up into one routine with a flag to avoid code duplication if that's wanted.

If I didn't miss anything, all routines should be able to run on one or multiple zbest files now (only really tested this for the deltas so far, but the changes are in).

bin/picca_co.py
```diff
@@ -181,6 +183,10 @@ def main():
                         help='Number of processors')
 
     args = parser.parse_args()
+    if len(args.drq)==1:
+        args.drq=args.drq[0]
+    if args.drq2 is not None and len(args.drq2)==1:
+        args.drq2=args.drq2[0]
```
iprafols, line 189: why are you doing this change? What happens if args.drq has more than one item in the list? Is this managed by io.read_objects ?

py/picca/utils.py
```diff
@@ -470,3 +470,127 @@ def unred(wave, ebv, R_V=3.1, LMC2=False, AVGLMC=False):
     corr = 1. / (10.**(0.4 * curve))
 
     return corr
+
+
+def catalog_from_zbest(zbestfiles,outfile,zmin,zmax):
```
iprafols, line 475: What is the purpose of this function? Inputs? Outputs? This function should have a docstring explaining that. If I understood it properly its purpose is to generate a catalogue from a list of files. If so, wouldn't this function be better placed in converters.py?

iprafols (changes requested):

bin/picca_co.py
Waelthus: It's to allow people to use multiple catalogs at the same time, which also allows e.g. running with a bunch of zbest files instead of creating a cat beforehand. The original `read_drq` routine cannot cope with a list of files so I changed it that way, the alternative would be always requiring proper catalogs always. Or to modify the `read_drq` to loop over all elements of a list instead of converting the list when it's only one entry

py/picca/utils.py
Waelthus: The idea of the function was to generate a catalog, I can add docstrings and move to converters. Not sure if we still want any of this as there's been some additional work on catalogs in cosmodesi

bin/picca_co.py
iprafols: It makes sense to allow the code to run on multiple catalogues. However, as things stand now the code will break if two catalogues are passed, won't it? Or am I missing something? I would also include the necessary changes in `read_drq` so that the code does not break.

py/picca/utils.py
iprafols: I think we can include it, given that it is already writen. Adding it to `converters.py` will keep the code clean even if we don't need to used it in the future.

andreufont:
Hi @Waelthus , @iprafols - let me see if I understand the changes in this PR. There are 11 files changes:
 - the first 8 of them are changes in the picca_cf.py scripts and similar, to allow them to take multiple arguments for the option --drq. I think it is better to remove these for now, and force the user to provide a single quasar catalog (unless there is a strong reason not to). To keep the main picca scripts simpler / tidier, I think it would be good to move things related to constructing the quasar catalog somewhere else.
 - then we have changes in utils.py and a new file, picca_generate_cat.py. These changes are to construct a quasar catalog from multiple zbest files, isn't it? If we have this, why do we need the option above to allow for more than one drq? If we believe this script will be used widely to construct quasar catalogs, should we move it from tutorials to bin? Or even better, didn't we start a repo in cosmodesi to construct quasar catalogs? Should we merge it there?
 - the remaining changes are in io.py. These are considerably longer changes, but it looks like most of them are related to read_zbest. If we assume that all users will provide a quasar catalog with the right "drq" format, we do not need this function here, right? I believe the changes that we really want to add to master is then just the changes around read_from_minisv_desi, right?

py/picca/io.py
```diff
@@ -259,6 +270,180 @@ def read_drq(drq_filename, z_min, z_max, keep_bal, bi_max=None):
 
     return ra, dec, z_qso, thingid, plate, mjd, fiberid
 
+def read_zbest(zbestfiles,zmin,zmax,keep_bal,bi_max=None):   #this needs more docs comparable to the ones of read_drq
```
iprafols, line 273: docstring required

py/picca/io.py
```diff
@@ -346,11 +531,19 @@ def read_data(in_dir,
     """
     userprint("mode: " + mode)
     # read quasar characteristics from DRQ catalogue
-    ra, dec, z_qso, thingid, plate, mjd, fiberid = read_drq(drq_filename,
-                                                            z_min,
-                                                            z_max,
-                                                            keep_bal,
-                                                            bi_max=bi_max)
+
+    try:
```
iprafols, line 535: given that function `read_data` has de `mode` variable, wouldn't it be better to solve this with an if/else block instead of with a try/except?

py/picca/io.py
```diff
@@ -1303,6 +1510,179 @@ def read_from_desi(nside,
 
     return data, num_data
 
+def read_from_minisv_desi(nside,
```
iprafols, line 1513: can you check the names of the variables to match the rest of the functions? They seem to be the old names, prior to standardisation. The list of changes is here: https://docs.google.com/spreadsheets/d/1PougvkVWN_H5BubOCOs1ylkS0wnFSyDbOc3cvPEJXoc/edit?usp=sharing

py/picca/io.py
```diff
@@ -1303,6 +1510,179 @@ def read_from_desi(nside,
...
+            if i==0:
+                print("reading all-band coadd as in minisv dataset")
+        except (KeyError,IOError):
+            if i==0:
+                print("couldn't read the all band-coadd, trying single band as introduced in Andes reduction")
+            bands=['B','R','Z']
+
+        for str_band in bands:
+            dic={}
+            dic['LL'] = sp.log10(h['{}_WAVELENGTH'.format(str_band)].read())
```
iprafols, line 1614: use `np`

py/picca/io.py
```diff
@@ -1303,6 +1510,179 @@ def read_from_desi(nside,
...
+            dic['FL'] = h['{}_FLUX'.format(str_band)].read()
+            dic['IV'] = h['{}_IVAR'.format(str_band)].read()*(h['{}_MASK'.format(str_band)].read()==0)
+            w = sp.isnan(dic['FL']) | sp.isnan(dic['IV'])
```
iprafols, line 1617: use `np`

iprafols (changes requested):

iprafols:
Hi @Waelthus, @andreufont

- I had missed the changes in io.py, thanks Andreu  to point it out. I reviewed it and send some comments.

- I agree with Andreu that the code would be cleaner if it only accepts one catalogue at a time. 

- I'm not sure about not needing read_zbest

- Can you run `yapf --style google` and then `pylint` to make sure the code follows the python standards?

andreufont:
I think we should have a very simple, clear and documented way to generate a quasar catalog from any combination of zbest files, and hide this from Picca. It should take a couple of minutes (most likely seconds) to generate it.

A different problem is that the current format of the quasar catalog is SDSS-oriented, with lots of keywords that do not make sense in DESI. I would start a new issue about this, it is related to what we have discussed earlier about making Picca less SDSS-oriented.

alxogm:
> The major question for this PR is if the mode to run on zbest files or zbest-like cats should be included or removed. So far, except for myself, I think the only person that actually tried it out at some point is @Andrea-MG (see discussion along https://desisurvey.slack.com/archives/C012T7WJZ5M/p1591193991092200), not sure if she still does.

Waelthus:
ok, given the discussion in the telecon I'll remove the additional stuff about zbest. Catalogue creation can then be done using [Paul's tools](https://github.com/cosmodesi/desi_lya_sv/blob/martini/bin/createqsocat.py). I'll double check that it works properly on it's outputs again. 
Is it possible/allowed to put e.g. a cat from this tool and some corresponding miniSV data into a travis test? Then in principle this could be done as well to make sure future changes don't break those analyses...

iprafols:
Given that mini-SV data is not public yet, I'm not sure we can add it to the repo yet. However we can do the test with a flag to skip it if the file is not present

julianbautista:
Thanks Michael.
Yes we can put a fake catalog with the same format for a travis test.

julianbautista:
Hi all, 
I tried to run this branch on andes data following Corentin's tutorial and it didn't work for any of the BAO/from* options. 
- It says that the "night" information cannot be found. I see the line 1446 of py/picca/io.py is commented out. 
- Or it cannot find the THING_ID for the DESI_mini_SV_catalog version. The qsocat.fits doesn't have the info.

bin/picca_co.py
Waelthus: ok, these changes are removed now.

py/picca/utils.py
Waelthus: the catalog making is removed now

py/picca/io.py
Waelthus: function is removed

py/picca/io.py
Waelthus: this particular case has been removed, similar try/excepts have been changed to more clear if/else conditions

py/picca/io.py
Waelthus: yes, some of them were, should be fixed now

py/picca/io.py
Waelthus: yep, never got why picca used sp. anyway for those things

Waelthus:
now it runs again, catalog making, running on multiple cats and generating a cat have been removed. The naming convention is now followed, I think.

Waelthus:
There was one more bug in read_from_desi being introduced in the merging, fixed it and double checked on NERSC. So we should now be able to do desi models as well as miniSV data again.
There's also now a travis test for the minisv I/O, including generating a mock cat and mock spectra (both in a reduced miniSV format) and running picca_deltas after (currently only checking if the run is successful, not comparing outputs).

iprafols (approved):

Issue #688, opened 2020-03-27
Drop support for python2?


andreufont:
Now that python2 is officially abandoned by python, we should stop supporting it in Picca. Do we know of anyone still using Picca from python2?

It is just annoying to keep using "future" functions instead of going full-on python3.

iprafols:
I agree that we should go full-on python 3

andreufont:
@londumas - any reason to keep support for python2? 

Otherwise, after the DR16 paper is accepted, we can drop support for python2. We will have to do it sooner or later!

londumas:
@andreufont, is there a way to just remove the `future` imports and have the code still be both python 2 and python 3 compatible ?

andreufont:
I don't think so... But also, why do we want to keep that compatibility? Do we know of anyone using Picca with python2?

PR #703, opened 2020-05-01
fixed issue697

iprafols:


iprafols:
There seems to be problems when installing llvmlite when using python2. Maybe another reason to go through with issue #688?

iprafols:
This fix was also applied in branch standardisation, now merged into master

PR #707, opened 2020-05-14
llvmlite==0.31.0 for python2

londumas:
llvmlite==0.31.0 for python2, because version 0.32.1 is not compatible with python2

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,8 @@ python:
     - "3.6"
 
 install:
-    - pip install -r requirements.txt
+    - if [[ $TRAVIS_PYTHON_VERSION == 2.7 ]]; then pip install -r requirements-python2.txt; fi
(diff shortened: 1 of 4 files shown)


londumas:
@andreufont, this PR fixes the python2 dependencies problem with e.g. llvmlite.
Dependencies are starting to drop python2, and thus are breaking the travis tests for this version of python2. This PR fixes this issue by lisiting the required version of the dependancies for python2 is a separte file. This will allow to be up to date with the latests version for python3

iprafols:
Don't you think it would be easier to just drop python2 support, as suggested in #688 ?

londumas:
@iprafols, that is a way to drop python2. We keep track of the version of the package needed for python2 in requirements-python2, and move on by default for python3 using the simple requirements.txt

andreufont:
I think I agree with @iprafols, I'd favor the option of dropping python2 altogether. If we don't do it now, we'll do it in a year from now, and meanwhile we'll have to keep using __future__ modules.

Do we know of anyone who has used the python2 version in the last year?

PR #708, opened 2020-06-05
Correct the term used to model the gaussian smoothing in mocks : the …

TEtourneau:
…term is squared when applied to the P(k)


TEtourneau:
It is the pk_gauss_smoothing function, defined in the fitter2/pk.py file.
In the mock, we apply a gaussien smoothing to the delta field:
W(k) = exp(- k^2 \sigma^2 / 2)
where \sigma is the size of the voxels.
This term is squared when applied to the power spectrum.

julianbautista:
Ok I see. This is ok for me to merge but I don't have the rights for some reason.

julianbautista:
@iprafols could you look into this, and make the same change in your branch? thanks

iprafols:
Hi @julianbautista, since my branch does not modify the fitter, then this should be included automatically when we merge with master. I also don't have the rights to merge the pull request. Maybe @andreufont will have them/be able to give them to us?

andreufont:
Hi @TEtourneau - this is only used when fitting mocks, right? I just want to make sure that this is not impacting the DR16 results on data.

Also, do you mind adding a line or two of comments, since you are already modifying that function? 

Once you clarify this, I'll be happy to merge.

andreufont:
Also, before merging, I need someone to review this. It looks like @julianbautista has already reviewed it, but it does not show up as reviewed on GitHub. You will have to actually hit the submit review button, or similar.

@iprafols , @julianbautista - I will see if I can add you as admin of Picca, so that you can also merge branches.

TEtourneau:
Yes this pk function is only used when fitting mocks.
Plus, I added some comments.
Tell me if it's OK for you to merge.

py/picca/fitter2/pk.py
```diff
@@ -423,13 +423,15 @@ def pk_hcd_Rogers2018_uv_cross(k, pk_lin, tracer1, tracer2, **kwargs):
 def pk_gauss_smoothing(k, pk_lin, tracer1, tracer2, **kwargs):
     """
     Apply a Gaussian smoothing to the full correlation function
-
+    Args : k, sigma_smooth par and per
```
andreufont, line 426: This docstring does not seem right. It says "k, sigma_smooth", but the arguments are "k, pk_lin, tracer1..."

py/picca/fitter2/pk.py
```diff
@@ -423,13 +423,15 @@ def pk_hcd_Rogers2018_uv_cross(k, pk_lin, tracer1, tracer2, **kwargs):
...
+    return : P(k) multiplied by G(k)^2
+    with G(k) = exp(-kpar^2 sigma_par^2 - kperp^2 sigma_perp^2)
     """
     kp  = k*muk
     kt  = k*np.sqrt(1.-muk**2)
     st2 = kwargs['per_sigma_smooth']**2
     sp2 = kwargs['par_sigma_smooth']**2
-    return sp.exp(-(kp**2*sp2+kt**2*st2)/2.)
+    return sp.exp(-(kp**2*sp2+kt**2*st2)/2.)**2
```
andreufont, line 434: We should use numpy instead of scipy here, for consistency with the rest of the Picca.

Also, why do you divide by 2 the exponent, and then square it? Couldn't you use the definition of G(k) as described in the docstring?

iprafols (changes requested):
Hi Thomas, given that these are small changes it should be trivial to merge master into here and then we can accept the PR.

Also, I see that this file is still using sp (even in master). Can you update this file to using np?

TEtourneau:
Thanks for the review !

Actually, I did the minimal changes to have a correct pk_gauss_smoothing function without breaking the code.
There are still a lot of scipy used in this file. I guess it would be a proper solution to have a dedicated branch that remove all scipy uses in fitter2.
Similarly, the args given are obviously not used. But again, I didn't want to break the code and so I did the minimal changes.
Do you want me to change these arguments ? I don't think it is very relevant since Andrei is working on a proper fitter2.

iprafols:
I think we can reopen issue #680 and address a new pull request to fix scipy to numpy in both fitter and fitter2, for consistency with the rest of picca. But the docstring should be fixed before accepting the PR

TEtourneau:
I merged the master branch into my branch.
I also updated the docstring: there are more details, and I specified the arguments that are not used (I kept them not to break the code).
Is it OK like this ?

iprafols (approved):
I don't understand why are there some parameters that are not used, but I take it this is not the place th fix that

iprafols:
@andreufont  are you happy with this documentation?

andreufont (approved):

TEtourneau:
> I don't understand why are there some parameters that are not used, but I take it this is not the place th fix that

I don't know. I think this is something coded by Helion, he probably copied/pasted some other function and didn't pay attention.
@andreufont Thanks for the merge !

andreufont:
I believe this is caused by the black magic in how fitter2 figures out which model to use. You can define extra functions with new models, but the input has to have this form and arguments? Not sure, @andreicuceu can probably explain better.

andreicuceu:
@andreufont you are right, all pk components have to have the same signature defined in the pk class (lines 10-26 of pk.py). And btw, for the kwargs of all these functions, we actually pass the full parameter list, it's just that each function extracts only what it needs from the dictionary. But this is impossible to document as it is run dependent, so I think for each function we should just document the stuff that is actually used (as was done here).

PR #713, opened 2020-06-26
fix python2 incompatibility for iminuit

londumas:
requires the latest version compatible with python2 for iminuit.
Update other packages to latest version just for clarity.
Fix failing tests for python2
```
iminuit==1.3.10
```


londumas:
@iprafols This PR fixes the crash of the unitest, could you have a look ?
It is very simple.

iprafols:
I cannot merge the pull request, but since there are only changes in the versions of the packages, and the travis test passes, I think the pull request is fine

PR #718, opened 2020-07-19
update iminuit requirements

londumas:
update iminuit requirements after they corrected their bug: https://github.com/scikit-hep/iminuit/issues/458#event-3562286426


londumas:
@andreufont, this is a very benign change to take into account the bug fixes for minos in iminuit.
It has no consequences on picca or the DR16 analyses

PR #724, opened 2020-08-28
Standardisation

iprafols:
This commit contains a merge from master with all the conflicts resolved


bin/picca_pk2fits.py
```diff
@@ -1,6 +1,7 @@
 #!/usr/bin/env python
 
 from astropy.io import fits
+import numpy as np
 import scipy as sp
```
julianbautista, line 5: there is no more need of scipy ?

py/picca/fitter/parameters.py
```diff
@@ -1,3 +1,4 @@
+import numpy as np
 import scipy as sp
```
julianbautista, line 2: scipy

bin/picca_pk2fits.py
iprafols: updated

py/picca/fitter/parameters.py
iprafols: updates

iprafols:
I rerun the tests on the commit 165d123 which originally passed to double check that there is no problem with versions. Not sure why travis tests now fail

I fixed commit ad4ce1c, sp.interpolate was changed to np.interpolate but this is not in numpy

iprafols:
With the new fix the tests pass, so I think we can go ahead and merge

julianbautista (approved):

Issue #725, opened 2020-08-31
Travis tests passing, but still having tracebacks in some parts


Waelthus:
This is copied over from slack.
@iprafols @julianbautista 
I just tried to merge master (post-standardization now) into one of my branches (minisv_deltas) and while Travis tells me things are fine, there are some tracebacks in the travis output. I double checked the output for master itself (https://travis-ci.org/github/igmhub/picca/builds/722140282 line numbers below are from here) and they are there, too.  I think it's a bit weird that travis gives green light even though something obviously did not run through.

line 1600:
```Traceback (most recent call last):
  File "/home/travis/virtualenv/python3.6.7/bin/picca_export_cross_covariance.py", line 4, in <module>
    __import__('pkg_resources').run_script('picca==4.0', 'picca_export_cross_covariance.py')
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/pkg_resources/__init__.py", line 665, in run_script
    self.require(requires)[0].run_script(script_name, ns)
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/pkg_resources/__init__.py", line 1470, in run_script
    exec(script_code, namespace, namespace)
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/picca-4.0-py3.6.egg/EGG-INFO/scripts/picca_export_cross_covariance.py", line 48, in <module>
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/fitsio/header.py", line 345, in __getitem__
    raise KeyError("unknown record: %s" % item)
KeyError: 'unknown record: HLPXSCHM'
```

line 1884:
```
Traceback (most recent call last):
  File "/home/travis/virtualenv/python3.6.7/bin/picca_export_co.py", line 4, in <module>
    __import__('pkg_resources').run_script('picca==4.0', 'picca_export_co.py')
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/pkg_resources/__init__.py", line 665, in run_script
    self.require(requires)[0].run_script(script_name, ns)
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/pkg_resources/__init__.py", line 1470, in run_script
    exec(script_code, namespace, namespace)
[...]

PR #726, opened 2020-09-02
Travis fix

Waelthus:
This fixes part of Issue #725

One of the tracebacks shown by travis on current master is fixed. The other one ('picca_export_co') I didn't yet find the reason for. If I'm running the test on files in `picca/py/picca/test/data` everything works smoothly, but when running the test it fails, so there might be some difference between saved outputs in the directory and when calling the routines in the test.

I also changed the travis test so that routines which do not compare fits or hdf5 files (i.e. all `picca_export_*` and `picca_co`) will fail the test in cases where the code does not complete. The current master does not check for completion in these cases and thus returned a pass in travis without actually passing everything.

### DIFF
--- a/bin/picca_co.py
+++ b/bin/picca_co.py
@@ -321,7 +321,7 @@ def main():
     units = ['h^-1 Mpc', 'h^-1 Mpc', '', '']
     results.write([r_par, r_trans, z, num_pairs],
                   names=['RP', 'RT', 'Z', 'NB'],
-                  headerer=header,
+                  header=header,
                   comment=comment,
                   units=units,
                   extname='ATTRI')
--- a/bin/picca_export_cross_covariance.py
+++ b/bin/picca_export_cross_covariance.py
@@ -44,8 +44,8 @@ if __name__ == '__main__':
         hdul = fitsio.FITS(filename)
         header = hdul[1].read_header()
         nside = header['NSIDE']
-        head = hdul[2].read_header()
-        healpix_scheme = head['HLPXSCHM']
+        header2 = hdul[2].read_header()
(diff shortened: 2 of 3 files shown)


iprafols (approved):

iprafols:
I checked the changes until better variable names. We should check why are travis tests failing before actually doing the merge

iprafols (approved):

Waelthus:
ok, I think I found the other bug. Writing the outputs of `picca_co.py`  contained a typo that lead to not writing proper headers, but this failed silently. So there should ideally be travis tests actually comparing output files for everything. For the moment as a small fix I only included checking the return code which did not raise an error when running `picca_co.py`, but at least raises an error when running `export_picca_co.py`.

iprafols:
For build on commit 080f334, the code seems to be failing at line 129 of bin/picca_export_co.py. The error message is related to not finding the card "NOBJ" in the header of files co_DD.fits.gz and co_RR.fits.gz. I checked those files in the test suite and they seem to be fine. This card is present on them so probably this is an error originated on bin/picca_co.py

iprafols (approved):

Waelthus:
> For build on commit [080f334](https://github.com/igmhub/picca/commit/080f334a37a1667f6fc15f6d0729ac5f9b087703), the code seems to be failing at line 129 of bin/picca_export_co.py. The error message is related to not finding the card "NOBJ" in the header of files co_DD.fits.gz and co_RR.fits.gz. I checked those files in the test suite and they seem to be fine. This card is present on them so probably this is an error originated on bin/picca_co.py

Yes, it's fixed now.

Issue #684, opened 2020-03-24
Lots of Numba warnings when measuring correlations


andreufont:
On a fresh new installation of Picca at NERSC, following the instructions here, I get several warnings from numba (see below for full message).

I don't know much (anything) about numba, so it would be great to have some help here.
```
  warnings.warn(errors.NumbaWarning(warn_msg,
/global/homes/f/font/.conda/envs/picca_plots/lib/python3.8/site-packages/numba-0.48.0-py3.8-linux-x86_64.egg/numba/object_mode_passes.py:187: NumbaDeprecationWarning: 
Fall-back from the nopython compilation path to the object mode compilation path has been detected, this is deprecated behaviour.

For more information visit http://numba.pydata.org/numba-doc/latest/reference/deprecation.html#deprecation-of-object-mode-fall-back-behaviour-when-using-jit

File "../../../../../../homes/f/font/.conda/envs/picca_plots/lib/python3.8/site-packages/picca-4.0-py3.8.egg/picca/xcf.py", line 168:
<source missing, REPL/exec in use?>

  warnings.warn(errors.NumbaDeprecationWarning(msg,
/global/homes/f/font/.conda/envs/picca_plots/lib/python3.8/site-packages/numba-0.48.0-py3.8-linux-x86_64.egg/numba/object_mode_passes.py:177: NumbaWarning: Function "fill_dmat" was compiled in object mode without forceobj=True, but has lifted loops.
```

londumas:
Is it related to issue https://github.com/igmhub/picca/issues/634 ?

andreufont:
It might be. That was an actual error, and it was "fixed" by specifying a particular version of numba, but clearly it didn't solve all the issues.

julianbautista:
These errors are happening because the functions under @jit cannot be compiled in nopython mode. 
See more details here: http://numba.pydata.org/numba-doc/latest/reference/deprecation.html#deprecation-of-object-mode-fall-back-behaviour-when-using-jit

I tried to modify the code to make it work in nopython mode but I failed. 
So I go with the easy solution for now. 
As recommended I've added:
`@jit(forceobj=True)`
This removes the error messages. 

However, adding forceobj=True is almost identical to not having the `@jit`.  
I did a quick test and the performance is indeed the same. I will keep the jit though for now.

iprafols:
I'm not sure about the performance being the same. The Travis test on the PR #731 failed due to the tests being too long so we need to find a different solution

[...]

PR #731, opened 2020-09-09
Numba warnings

julianbautista:
I'm replacing all 
`@jit`
with
`@jit(forceobj=True)`

That removes the warnings we see when computing the correlation functions, discussed in https://github.com/igmhub/picca/issues/684

### DIFF
--- a/py/picca/cf.py
+++ b/py/picca/cf.py
@@ -189,7 +189,8 @@ def compute_xi(healpixs):
     return weights, xi, r_par, r_trans, z, num_pairs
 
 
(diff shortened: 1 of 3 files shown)


iprafols (approved):

iprafols:
Let's wait until Travis is finished, then we can double check that the warnings are gone in the logs there and merge the pull request

iprafols:
The jobs were terminated due to exceeding the time limit

julianbautista:
I removed the changes. Let's close this until we find a real solution.

PR #732, opened 2020-09-09
function __add__ from Forest renamed to coadd

iprafols:
I modified the function name and Travis seems to be fine with it so this might be a bit tricky to revise. After that, I looked at the places where two Forest instances were summed and replaced the sum by the coadd function. Note that I only found this to happen in the reading functions


julianbautista (approved):
Looks good to me. Couldn't find more locations that call this function.

iprafols:
Let's merge then

PR #733, opened 2020-09-09
changing __xor__ by get_angle_between

julianbautista:
This is another change needed that will help understand the code. 
People had questions about this bit multiple times and it's confusing.


iprafols (approved):
I couldn't find any other instances that need changing. Looks fine to me

Issue #712, opened 2020-06-25
Fiducial cosmology parameters missing in correlation function files


julianbautista:
The correlation function code uses values for Omega_m and Omega_r to convert log_lambda into comoving distances. These values need to be written in the headers of the cf_*.fits files. 

This should apply for the auto, cross, distortion matrices and metal matrices.

julianbautista:
They are now in the headers of the correlation files.

PR #736, opened 2020-09-11
Fiducial cosmo parameters

julianbautista:
I'm adding the fiducial cosmological parameters to the headers of the correlation function files (cf.fits.gz, xcf.fits.gz, etc.) as requested in #712 

The files in py/picca/test/data were updated accordingly.


iprafols (approved):
Looks fine to me

PR #740, opened 2020-09-23
Faster cf (removes numba warnings)

julianbautista:
This branch includes new "fast" versions of previous functions with proper numba "nopython" implementation. 
I kept the previous versions for reference if needed. 

This removes the warning messages when computing the correlations or distortion matrices. 

As a bonus this improves the speed of the auto correlation by a factor of at least 3, the auto distortion matrix by a factor of 1.3 but no significant improvement for the cross. 

I tested that I have the same output files for DR16.


py/picca/cf.py
```diff
@@ -158,27 +158,28 @@ def compute_xi(healpixs):
...
-                r_trans[:len(rebin_r_trans)] += rebin_r_trans
-                z[:len(rebin_z)] += rebin_z
-                num_pairs[:len(rebin_num_pairs)] += rebin_num_pairs.astype(int)
+                         delta1.weights, delta1.delta, 
+                         delta2.z, delta2.r_comov, delta2.dist_m, 
+                         delta2.weights, delta2.delta, 
+                         ang, same_half_plate, 
+                         weights, xi, r_par, r_trans, z, num_pairs)
+
+                #xi[:len(rebin_xi)] += rebin_xi
```
iprafols, line 177: why are these commented? I take it the values are now updated in the functions themselves, aren't they? If that is correct, maybe we should remove these lines

py/picca/cf.py
```diff
@@ -279,6 +280,95 @@ def compute_xi_forest_pairs(z1, r_comov1, dist_m1, weights1, delta1, z2,
     return (rebin_weight, rebin_xi, rebin_r_par, rebin_r_trans, rebin_z,
             rebin_num_pairs)
 
+@jit(nopython=True)
+def compute_xi_forest_pairs_fast(z1, r_comov1, dist_m1, weights1, delta1, z2,
```
iprafols, line 284: why don't we replace this function with compute_xi_forest_pairs? I mean, is there a particular reason as to why we would like to keep the slow version? Similarly for the other updated functions

py/picca/cf.py
```diff
@@ -596,6 +686,201 @@ def compute_dmat_forest_pairs(log_lambda1, log_lambda2, r_comov1, r_comov2,
...
+                + eta6[k] * log_lambda_minus_mean2[j] + 
+                + eta7[k] * log_lambda_minus_mean1[i] + 
+                + eta8[k] * log_lambda_minus_mean1[i] * log_lambda_minus_mean2[j] 
+                - eta1[i+num_pixels1*k] 
+                - eta2[j+num_pixels2*k] 
+                - eta3[i+num_pixels1*k] * log_lambda_minus_mean2[j] 
+                - eta4[j+num_pixels2*k] * log_lambda_minus_mean1[i])
+                )
 
 def compute_metal_dmat(healpixs, abs_igm1="LYA", abs_igm2="SiIII(1207)"):
```
iprafols, line 885: I take it we don't have the fast version for this function, do we? Is it planned for a later commit?

py/picca/xcf.py
```diff
@@ -128,32 +128,34 @@ def compute_xi(healpixs):
...
-                weights[:len(rebin_weight)] += rebin_weight
-                r_par[:len(rebin_r_par)] += rebin_r_par
-                r_trans[:len(rebin_r_trans)] += rebin_r_trans
-                z[:len(rebin_z)] += rebin_z
-                num_pairs[:len(rebin_num_pairs)] += rebin_num_pairs.astype(int)
+                         weights_qso, ang,
+                         weights, xi, r_par, r_trans,
+                         z, num_pairs)
+                         
+                #xi[:len(rebin_xi)] += rebin_xi
```
iprafols, line 153: as in the other file

iprafols (approved):
I think the changes are fine, but do consider my comments on keeping the old functions before merging

py/picca/cf.py
julianbautista: I left them because I also left the old versions of the functions and I would be easy to come back.

py/picca/cf.py
julianbautista: There were no warnings on this one. I don't plan to update it for now. But you can add this to the todo list for other people. It's not easy though.

py/picca/cf.py
iprafols: ok, so maybe we want to add a comment specifying that these are deprecated commands (also for the older functions) that will be removed later

iprafols (approved):

PR #745, opened 2020-09-24
Making code less SDSS dependent : part one

julianbautista:
In this PR, there are several things being implemented:

1) simplifying io.py with the usage of astropy.table.Table to store the metadata. This makes it easy to propagate metadata for SDSS or DESI or other types. There are some modes that I didn't update yet (I wonder if anyone uses them).

2) removed some warning messages in picca_deltas.py

3) removed the dependency from Delta in picca_deltas.py

4) improving the readability of the logs in picca_deltas.py (e.g, when it prints the var_lss, eta, fudge, etc)

The next step is to actually modify the picca.data.QSO and Forest classes to remove the explicit need of plate, mjd, fiber.


py/picca/io.py
```diff
@@ -1375,116 +1112,111 @@ def read_from_minisv_desi(in_dir,
...
-                flux = (ivar*spec['FL'][w_t]).sum(axis=0)
-                ivar = ivar.sum(axis=0)
-                w = ivar > 0.
-                flux[w] /= ivar[w]
+                ivar = spec['IV'][w_t]*1
+                flux = spec['FL'][w_t]*1
 
                 if pk1d is not None:
-                    reso_sum = spec['RESO'][w_t].sum(axis=0)
+                    reso_sum = spec['RESO'][w_t]*1
```
Waelthus, line 1205: why the "*1"? I don't think this is the same operation.

Waelthus (changes requested):
Looks good in general, there's still a bunch of statements saying "Not implemented" in `io.py`, so those need to be fixed for sure. Also some changes might break compatibility with earlier code versions, so people would need to be informed before merging as previous analysis cannot be run anymore without chanegs of either catalogs or submission scripts.

Waelthus:
@andreufont : I don't think we can really rely on travis for all the hiccups, there's just too many options people might use that are currently not covered in the test, and some will fail, e.g. running on desi mocks using a DRQ. I guess we should actually start working on improving the tests before the next big PR.

bin/picca_deltas.py
```diff
@@ -60,6 +59,57 @@ def get_metadata(data):
...
+        exposures_diff = forest.exposures_diff
+        if forest.exposures_diff is not None:
+            exposures_diff /= mean_expected_flux_frac
+        ivar = forest.ivar / (eta + (eta == 0)) * (mean_expected_flux_frac**2)
+
+        forest.delta = delta
+        forest.weights = weights
+        forest.exposures_diff = exposures_diff
+        forest.ivar = ivar
+
```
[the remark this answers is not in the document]
julianbautista: Yes. The Delta class is barely used for continuum fitting. 
The goal is to break the code into three main components (continuum fitting, correlations, fitters). 
Yes, there are no other uses of Delta in this part of the code.

bin/picca_deltas.py
```diff
@@ -403,14 +458,13 @@ def main():
     ### Correct multiplicative pipeline inverse variance calibration
     if args.ivar_calib is not None:
         try:
-            hdul = fitsio.FITS(args.ivar_calib)
-            log_lambda = hdul[2]['LOGLAM'][:]
-            eta = hdul[2]['ETA'][:]
+            hdul = fitsio.FITS(args.ivar_calib, ext=2)
```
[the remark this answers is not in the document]
julianbautista: yes, fixed.

py/picca/prep_del.py
```diff
@@ -216,17 +220,16 @@ def chi2(eta, var_lss, fudge):
...
+                var_delta[index * num_var_bins:(index + 1) * num_var_bins] - variance)
             weights = var2_delta[index * num_var_bins:(index + 1) *
                                  num_var_bins]
             w = num_qso[index * num_var_bins:(index + 1) * num_var_bins] > 100
             return np.sum(chi2_contribution[w]**2 / weights[w])
 
         minimizer = iminuit.Minuit(chi2,
-                                   forced_parameters=("eta", "var_lss",
-                                                      "fudge"),
+                                   name=("eta", "var_lss", "fudge"),
```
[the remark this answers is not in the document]
julianbautista: yes, this removes a Deprecation Warning. The warning itself suggested the change!

py/picca/data.py
```diff
@@ -456,8 +431,6 @@ def __init__(self,
                 Modified Julian Date of the observation.
             fiberid: integer
                 Fiberid of the observation.
-            order: 1 or 0
```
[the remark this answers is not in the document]
julianbautista: This is in Forest, isn't it?

bin/picca_deltas.py
```diff
@@ -434,44 +488,46 @@ def main():
                                          log_file=log_file,
                                          keep_bal=args.keep_bal,
                                          bi_max=args.bi_max,
-                                         order=args.order,
                                          best_obs=args.best_obs,
                                          single_exp=args.single_exp,
                                          pk1d=args.delta_format,
                                          spall=args.spall)
```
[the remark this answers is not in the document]
julianbautista: I'm not sure I understand the suggestion. best_obs, single_exp, spall are options for some of the read_from_* functions.

py/picca/io.py
```diff
@@ -184,91 +146,83 @@ def read_drq(drq_filename, z_min, z_max, keep_bal, bi_max=None):
...
-    hdul = fitsio.FITS(drq_filename)
+    userprint('Reading catalog from ', drq_filename)
+    catalog = Table(fitsio.read(drq_filename, ext=1))
+
+    keep_columns = ['RA', 'DEC', 'Z']
+    if 'desi' in mode:
+        obj_id_name='TARGETID'
+        catalog.rename_column('TARGET_RA', 'RA')
+        catalog.rename_column('TARGET_DEC', 'DEC')
+        keep_columns += ['TARGETID', 'TILEID', 'PETAL_LOC', 'NIGHT', 'FIBER']
```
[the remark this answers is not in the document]
julianbautista: Yes. Could you send me a script where this is run on desi mocks? Or point me to someone that does that. I agree this is not covered by Travis and has to be added. 
I added 
`if 'desi' in mode and 'TARGETID' in catalog.colnames:`
which should cover the cases you mention.

py/picca/io.py
julianbautista: I want to make sure we are doing copies of the data. 
Also I didn't understand the operations with ivar (some weighted average).

py/picca/io.py
```diff
@@ -1375,116 +1112,111 @@ def read_from_minisv_desi(in_dir,
...
-                filenames.append(f_in)
-                break
-
-    z_table = {t:z for t,z in zip(thingid,z_qso)}
-
-    for index,filename in enumerate(filenames):
-        userprint("\rread tile {} of {}. ndata: {}".format(index,len(filenames),num_data))
+    for entry in catalog:
+        fi = (f"{entry['TILEID']}/{entry['NIGHT']}/"+
+              f"coadd-{entry['PETAL_LOC']}-{entry['TILEID']}-{entry['NIGHT']}.fits")
```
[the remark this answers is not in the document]
julianbautista: Ok fine. I will leave it as a comment and put back your glob for now.

bin/picca_deltas.py
```diff
@@ -434,44 +488,46 @@ def main():
...
+                w = mask['frame'] == 'RF'
+                mask_rest_frame = np.array([np.log10(mask['wave_min'])[w], 
+                                            np.log10(mask['wave_max'])[w]]).T
+            else:
+                mask_rest_frame = np.array([])
+            if 'RF_DLA' in mask['frame']:
+                w = mask['frame'] == 'RF_DLA'
+                mask_rest_frame_dla = np.array([np.log10(mask['wave_min'])[w], 
+                                                np.log10(mask['wave_max'])[w]]).T
+            else:
```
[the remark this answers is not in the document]
julianbautista: Yes. It's still a lot of noisy code. I was planning to send a Table directly to forest.mask instead of these arrays.

julianbautista:
@andreufont yes, sorry, I started doing one change and I did others in between, but they are quite minor (warnings).

I'm running DR16 tests but on S82 data and comparing with master. I'm also running DESI mini SV. 
I'm bugging other people to provide tests for mocks. 

In any case, with git is easy to come back to previous versions and track problems.

bin/picca_deltas.py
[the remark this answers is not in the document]
Waelthus: yes, that would actually be a good idea to keep everything more clean. Generating one table of those and then passing them to mask and add_dla and getting the right columns from there. This would also be more flexible in case one would like to later add additional stuff to the masking, e.g. BAL_RF or something...

py/picca/prep_del.py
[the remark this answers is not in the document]
Waelthus: ok, then it should be fine.

py/picca/data.py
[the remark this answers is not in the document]
Waelthus: Oh, you're right. My bad. Yes, remove it from the initializer (which allows getting rid of it in `read_from_*`), but we need to set it manually somewhere else.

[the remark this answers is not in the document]
julianbautista: Yes, it's in picca_deltas.py. I set them after read_data().

bin/picca_deltas.py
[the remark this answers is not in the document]
Waelthus: What I meant is that you removed the order from the initilizer of Forest and as an argument for all the `read_from_*` routines, which I guess makes things less messy. But I'd suggest to still set the order in io.read_data(...), e.g. just before returning using the same 4 lines you've used here to set it. The reason would be to keep the picca_deltas cleaner as this doesn't actually do anything with the forests yet.

py/picca/io.py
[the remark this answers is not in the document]
Waelthus: about the script, please remind me when NERSC is back,

about the change to `if 'desi' in mode and 'TARGETID' in catalog.colnames` this would keep things in the cat on eBOSS names, which I guess would be removed in `read_from_desi`, that's why I suggested the mapping

py/picca/io.py
Waelthus: Ok, I think the sum was used here to do an average over all selected spectra (in case one has the same targetid in the sample mulitple times), that converts an [nspec, ndiag, npix] array to an [ndiag, npix] array so cannot be removed here. Also the *1 for a copy is not really good style ;-) why not use .copy() instead

All the ivar operations are to do inverse variance weighted coadding between different spectra with same THINGID/TARGETID within the same coadd-file. The actual coadd routine, currently only coadds the different arms I think.

All of those routines probaby need a bit of an update to be more clear...

py/picca/io.py
```diff
@@ -1152,46 +935,37 @@ def read_from_spplate(in_dir,
...
+                 f" in {time_read:.3f} per spec. "+
+                 f" Progress: {len(pix_data)}"+
+                 f" of {len(catalog)} ")
         hdul.close()
 
     data = list(pix_data.values())
     return data
 
-
+#-- Not implemented yet
```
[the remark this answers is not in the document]
julianbautista: Done

py/picca/io.py
```diff
@@ -857,52 +673,44 @@ def read_from_pix(in_dir,
...
+            )
         if log_file is not None:
             log_file.write("{} read\n".format(t))
-
     hdul.close()
 
     return pix_data
 
-
+#-- Not implemented yet
```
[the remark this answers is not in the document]
julianbautista: Done

bin/picca_deltas.py
```diff
@@ -353,68 +403,62 @@ def main():
...
-                                           fill_value="extrapolate",
-                                           kind="nearest")
-            hdul.close()
-        except (OSError, ValueError):
-            userprint(("ERROR: Error while reading flux_calib"
-                       "file {}".format(args.flux_calib)))
-            sys.exit(1)
-
-    ### Correct multiplicative pipeline inverse variance calibration
+        hdul = fitsio.read(args.flux_calib, ext=1)
```
iprafols, line 444: why are we removing the try/except block here? I think it is useful to inform the user if an error is occuring here

iprafols, line 444: hdul here should be renamed to hdu since it is now just a Header Data Unit, and not a Header Data Unit List

bin/picca_deltas.py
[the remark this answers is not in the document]
iprafols: I'm not sure what would be the best place for this. It is true that moving it to read_data(...) would keep picca_deltas cleaner, but adding it to read_data(...) feels weird in itself since the order is not actually read. I think this may be why Julian chose to move it here.

How about we add this as a method `set_order` of Forest? It would have the save 4 lines. I think this way we could have the benefits from all points of view

bin/picca_deltas.py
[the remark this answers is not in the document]
iprafols: You seem to agree on this (and I do as well), but are we thinking to add this in part 2?

bin/picca_deltas.py
```diff
@@ -705,13 +761,16 @@ def main():
...
     data_bad_cont = []
     for healpix in sorted(data.keys()):
-        deltas[healpix] = [
-            Delta.from_forest(forest, get_stack_delta, Forest.get_var_lss,
-                              Forest.get_eta, Forest.get_fudge,
-                              args.use_mock_continuum)
-            for forest in data[healpix]
-            if forest.bad_cont is None
-        ]
+        for forest in data[healpix]:
```
iprafols, line 764: Why are you removing the list comprehension? How does this affect performance?

py/picca/data.py
```diff
@@ -12,30 +12,6 @@
 from picca.utils import userprint, unred
 from picca.dla import DLA
 
-
-def get_variance(var_pipe, eta, var_lss, fudge):
```
iprafols, line 16: If I recall properly, this was left as a function since it was used in multiple places. I understand removing it and making the computation explicitly in class Forest but, what happened to the other places this was used?

py/picca/data.py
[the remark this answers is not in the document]
iprafols: This is addressed by the other question in picca_deltas

py/picca/io.py
```diff
@@ -148,127 +111,111 @@ def read_absorbers(filename):
...
+    w &= catalog['Z'] >= z_min
+    userprint(f" and z >= {z_min}        : nb object in cat = {np.sum(w)}")
+    w &= catalog['Z'] < z_max
+    userprint(f" and z < {z_max}         : nb object in cat = {np.sum(w)}")
 
     ## BAL visual
-    if not keep_bal and bi_max is None:
-        try:
-            bal_flag = hdul[1]['BAL_FLAG_VI'][:]
+    if keep_bal == False and bi_max is None:
```
iprafols, line 186: instead of `keep_bal == False` the standard is to type `not keep_bal`

py/picca/io.py
```diff
@@ -285,14 +232,11 @@ def read_dust_map(drq_filename, extinction_conversion_r=3.793):
         associated with the observation. Values are the extinction for that
         line of sight.
     """
-    hdul = fitsio.FITS(drq_filename)
-    thingid = hdul[1]['THING_ID'][:]
-    ext = hdul[1]['EXTINCTION'][:][:, 1] / extinction_conversion_r
-    hdul.close()
-
+    hdul = fitsio.read(drq_filename, ext=1)
```
iprafols, line 235: `hdul` here should be renamed to `hdu`

py/picca/io.py
```diff
@@ -770,18 +578,18 @@ def read_from_mock_1d(filename,
...
 
-    for t, r, d, z, p, m, f in zip(thingid, ra, dec, z_qso, plate, mjd,
-                                   fiberid):
-        hdu = hdul["{}".format(t)]
-        log_file.write("file: {} hdus {} read  \n".format(filename, hdu))
-        lambda_ = hdu["wavelength"][:]
-        log_lambda = np.log10(lambda_)
+    for entry in catalog:
+        thing_id = entry['THING_ID']
+        hdu = hdul[f'{thing_id}']
```
iprafols, line 583: why not use the .format() ? I think it is a bit clearer than the f"  " notation

py/picca/io.py
```diff
@@ -791,48 +599,31 @@ def read_from_mock_1d(filename,
...
+                   entry['RA'], entry['DEC'], entry['Z'], 
+                   entry['PLATE'], entry['MJD'], entry['FIBERID'],
                    exposures_diff, reso, mef))
 
     hdul.close()
 
     return pix_data
 
-
+#-- Is anyone using this?
```
iprafols, line 612: I don't think so, but I would keep it for a bit longer. Maybe we can add a DeprecationWarning if this function is used

py/picca/io.py
```diff
@@ -1375,116 +1088,118 @@ def read_from_minisv_desi(in_dir,
...
-                forest_temp = Forest(spec['log_lambda'],flux,ivar,t,ra[w_t][0],
-                    dec[w_t][0],z_table[t], p, m, f, order,
+                forest_temp = Forest(spec['log_lambda'], flux, ivar,
+                    entry['TARGETID'],
+                    entry['RA'], entry['DEC'], entry['Z'],
+                    entry['TILEID'], entry['NIGHT'], entry['FIBER'],  
                     exposures_diff, reso_in_km_per_s)
                 if forest is None:
-                    forest = copy.deepcopy(forest_temp)
+                    forest = forest_temp
```
iprafols, line 1202: why are you not using copy here but using it on line 1057?

py/picca/data.py
iprafols: Ok, I saw that you modified the other places as well

iprafols (changes requested):

py/picca/io.py
julianbautista: We should probably use copy just to make sure. I'm adding it back

py/picca/io.py
julianbautista: I disagree. How to solve this conflict ? :)

bin/picca_deltas.py
julianbautista: I removed because I was having the error message but it was not informative enough. 
Removing the try/except block I have more information on the error and the code stops anyway.

julianbautista: seriously...

py/picca/io.py
iprafols: I'm not strong headed here. I guess we can leave this with the f notation, but maybe @Waelthus can break the tie

bin/picca_deltas.py
iprafols: hahahha, you can leave it as hdul if you want, but this was just to let you know my reasoning behind using these two names in the rest of the code

iprafols: Ok, I see

bin/picca_deltas.py
julianbautista: Because we don't create Delta objects anymore. We modify them. 
Performance here is definitely not a bottleneck.

py/picca/io.py
julianbautista: ok

bin/picca_deltas.py
iprafols: makes sense

iprafols (commented):
I approve this changes, but there are still unresolved questions that should be addressed before merging, particularly those signaled by Michael

bin/picca_deltas.py
Waelthus: Agreed, why rename all variables to clearer names if immediately after standardization we stop following the naming rules ;-)

bin/picca_deltas.py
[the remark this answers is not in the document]
Waelthus: Couldn't one even add it in here? Sounds sufficiently simple, I'll have a look how much of a change this would be

bin/picca_deltas.py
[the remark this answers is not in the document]
Waelthus: This does sound like a good compromise to me. Another option that just came to my mind would be to have the data dictionary actually be an object to allow having a way to put all the 
```
for (heal)pix in data:
        for forest in data[(heal)pix]:
```
constructions into a function (or iterator or something), maybe even having parameters that are dataset and not really spectrum dependent such as e.g. order in there. But this might be too much of a change...

py/picca/io.py
Waelthus: I don't know if somebody uses it, but agree with Ignasi about the warning

py/picca/io.py
Waelthus: I tend to prefer the f-string for variable names or very short operations and the format for longer operations, in this case I'd be on @julianbautista's side

iprafols: ok, let's keep it with the f format then

bin/picca_deltas.py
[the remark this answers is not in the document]
iprafols: Yes, I think it would be too much of a change

py/picca/io.py
julianbautista: added a warning

bin/picca_deltas.py
julianbautista: fitsio.read gives a recarray when reading a table, not an HDU. But I'm calling it hdu now.

py/picca/io.py
julianbautista: In this part of the code, there cannot be two spectra for one single object/targetid. I think the sum was to convert an 2D array into a 1D array. 
I use np.where instead of a mask array so this sum() is not needed. I tested and it works.

iprafols (commented):
Changes look fine

py/picca/io.py
Waelthus: Ok, I only had this ever do the 2d to 1d conversion (it definitely does so with current data). Personally I like the mask array syntax more than np.where, but that's probably a question of taste.
If I'm not misreading the code having two spectra of the same target on the same tile in the same night would lead to coadding them in the old version, but would just pick the first of them in the new. Given that the outputs we're using are already coadded this should never happen. Might still be worth putting in a check and issuing a warning/raising an exception if there's ever more than one index in w_t.

I'm still against the "*1" as copy() is clearer to the reader (but I do not think the copying is even necessary in this case), i.e. a future maintainer might be tempted to remove a "*1" assuming it does nothing.

bin/picca_deltas.py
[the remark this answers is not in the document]
Waelthus: In this case I'd opt for @iprafols suggestion to add a set_order method. Also: rename the `pix` in the for loops to healpix for consistency

py/picca/io.py
julianbautista: I've added the warning and added .copy() for each array.

bin/picca_deltas.py
iprafols: but does it not have a header? If it is a recarray you can rename it to something else

julianbautista:
sorry I don't think this is that important. Feel free to change it in the
future.

julianbautista:
I tested this branch and I obtain the same results as master with
- eBOSS DR16 data on Stripe82 
- eBOSS quickquasar Saclay mocks (by Julianna and Thomas)

I'm still waiting for NERSC to come back so I can test the deltas in P1D format by Michael.

iprafols (commented):
changes look fine

py/picca/io.py
iprafols: This is addressed with commit d464ea8

bin/picca_deltas.py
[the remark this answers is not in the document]
Waelthus: Ok, did the changes, it affects `data.py`, `picca_deltas.py` and `picca_plotSpec.py`. Should I commit it in here, or should we just leave this as is and I open a PR?

py/picca/io.py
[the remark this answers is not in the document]
Waelthus: ok, the current approach seems to work on with a drq-style desi mock catalog and the london mocks, so I'd consider this solved.

py/picca/io.py
```diff
@@ -844,6 +627,8 @@ def read_from_pix(in_dir,
     Returns:
         List of read spectra for all the healpixs
     """
+    userprint('DeprecationWarning: this method will be deprecated.')
```
Waelthus, line 630: wouldn't it be better to use `warnings.warn("this method will be deprecated.", DeprecationWarning)`. No show stopper, but might be good practice for the future

bin/picca_deltas.py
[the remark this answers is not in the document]
julianbautista: I think you can commit it here. Thanks!

Waelthus (approved):
Ok, given that tests run and all remaining changes are minor cosmetics, I'd say it's good to go. Probably issue a warning that output file structure changed for `mode==desi` due to a change in nside from hardcoded to automatically estimated, but everything else can easily be done later in seperate PRs.

bin/picca_deltas.py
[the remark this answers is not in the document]
Waelthus: Done!

bin/picca_deltas.py
```diff
@@ -494,50 +494,26 @@ def main():
...
-                w = mask['frame'] == 'RF_DLA'
-                mask_rest_frame_dla = np.array([
-                    np.log10(mask['wave_min'])[w],
-                    np.log10(mask['wave_max'])[w]
-                ]).T
-            else:
-                mask_rest_frame_dla = None
+            col_log_min=Column(np.log10(mask['wave_min']),name='log_wave_min')
+            col_log_max=Column(np.log10(mask['wave_max']),name='log_wave_max')
+            mask.add_columns([col_log_min,col_log_max])
```
julianbautista, line 505: Thanks for these changes. This can be even simpler thanks to the dictionary-like behaviour of Tables. You can do:
```
mask['log_wave_min'] = np.log10(mask['wave_min'])
mask['log_wave_max'] = np.log10(mask['wave_max'])
```

Waelthus:
Oh cool, thanks! Didn't even know that this works ;) feel free to do the change

On October 1, 2020 2:12:20 PM GMT+02:00, Julian Bautista <notifications@github.com> wrote:
>@julianbautista commented on this pull request.
>
>
>
>> -                    np.log10(mask['wave_min'])[w],
>-                    np.log10(mask['wave_max'])[w]
>-                ]).T
>-            else:
>-                mask_rest_frame = np.array([])
>-            if 'RF_DLA' in mask['frame']:
>-                w = mask['frame'] == 'RF_DLA'
>-                mask_rest_frame_dla = np.array([
>-                    np.log10(mask['wave_min'])[w],
>-                    np.log10(mask['wave_max'])[w]
>-                ]).T
>-            else:
>-                mask_rest_frame_dla = None
>+            col_log_min=Column(np.log10(mask['wave_min']),name='log_wave_min')
>+            col_log_max=Column(np.log10(mask['wave_max']),name='log_wave_max')
>+            mask.add_columns([col_log_min,col_log_max])
>
>Thanks for these changes. This can be even simpler thanks to the dictionary-like behaviour of Tables. You can do:
>```
>mask['log_wave_min'] = np.log10(mask['wave_min'])
>mask['log_wave_max'] = np.log10(mask['wave_max'])
>```
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/igmhub/picca/pull/745#pullrequestreview-500281047
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

bin/picca_deltas.py
Waelthus: indeed, this is even better, done!

julianbautista:
Can we merge this now ?

iprafols (approved):

Issue #737, opened 2020-09-12
picca not compatible with python3.8 in MacOS


julianbautista:
After updating my python version in my mac, picca stopped working. 
The reason is that Python 3.8 on MacOS by default now uses "spawn" instead of "fork" as start method for new processes in multiprocessing. 
Adding `multiprocessing.set_start_method("fork")` in the first line below each `if __name__ == "__main__"` makes it work.
I don't think we need to solve this for MacOS users because most people use unix based systems to run picca, but we better have this documented here.

iprafols:
Does this fix work also for unix? If so, then I would add these changes to master

julianbautista:
unix doesn't need this fix because multiprocessing is implemented differently. But I haven't tested though.

andreufont:
Hi @julianbautista - Some might use picca on a laptop to debug stuff, or run small tests. If it is easy to fix, and it doesn't have any bad side effect, it might still be worth it. Maybe we can ask at the picca telecon, or in an email, whether anyone is using Picca on MacOS.

iprafols:
Note that this is also happening in NERSC with the new version of python

iprafols:
The fix proposed by Julian doesn't seem to work. When running the tests, the subprocess module is used to run the different codes. For the first code, the fix works perfectly. However, for the second code, the system thinks the multiprocessing.set_start_method has already been called, and the code crashes as this function can only be called once. However, if a try/except block is used, then the second code runs with the 'spawn' starting method and therefore crashes. 

I will look at other possible fixes

For some reason, when running the tests the multiprocessing thinks that it has already set the start method for the runs after picca_deltas, and it doesn't do so, resulting in a failure of the successive tests.

iprafols:
I fixed the problem by requiring a specific context for the multiprocessing in each of the files. This should work irrespective of the default context in the different OS.

iprafols:
@julianbautista I belive this is not an issue anymore. Unless you complain shortly, I will close the issue

PR #749, opened 2020-10-14
modified travis python version

iprafols:
When I run picca automatic tests (on branch master, updated) at NERSC and on my local laptop the automatic tests fail, but Travis says it is fine. It seems that the problem comes from the way pool is interacting with the classes. I checked that before line `data_fit_cont = pool.map(cont_fit, sorted_data)`, the variable `Forest.log_lambda_max_rest_frame` is correctly valued, but after the command it is set to `None`. 

I add this pull request to check if the problem lies with the python version (I'm using version 3.8, whereas travis was set on version 3.7), or has something to do with the OS

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: python
 python:
-    - "3.6"
+    - "3.8"
 
 install:
     - pip install -r requirements.txt


julianbautista:
Hi Ignasi. Just to mention that this is exactly the problem in #737

iprafols:
Thanks, I had missed it. I guess I'll  address it then

Waelthus (commented):
Ok, @iprafols your changes are ok. I additionally converted the Wick-Computations to numba that compiles with the nopython option and thus works also with newer versions, same for `co.py`'s `compute_xi_forest_pairs`.
The Wick 3-QSO terms seem to be broken even in the current master (the call syntax doesn't even match the function definition), I noted that halfway through the changes... So if we want to have those terms some fixup is needed. But I'm not enough of an expert on this to decide if we need those. But currently they cannot be used in either the new nor the old version, I think. They are also not part of the test_suite.

iprafols:
Great! Given that I started the PR, I cannot review your changes (though they look fine with me), let's wait until someone can review them before merging

Issue #753, opened 2020-10-30
bug found when computing Lya(Lya) x Lya(Lyb)


iprafols:
I found a bug when computing the autocorrelation between two different forests with the same absorption line. It seems that picca is ignoring the second set of deltas. I've identified the bit of code that is responsible for this bug and I will create a PR as soon as I fix it

iprafols:
After fixing the initial bug, there seems to be something else hidden as I do not recover the exact same correlation function as in the DR16 results computed by Andreu back when we published the results
![cf_LYA_in_LYA_LYA_in_LYB_z_0_10-exp](https://user-images.githubusercontent.com/9192975/97872786-e7c70d00-1d16-11eb-96a2-bb6d79058e93.png)

iprafols:
I don't think that the problem is with the deltas since I can reproduce the following correlation functions: Lya x Lya, Lya x QSO, Lya(Lyb) x QSO

iprafols:
This is now fixed with the last commit on PR #754

PR #754, opened 2020-10-30
fixed functions fill_neighs

iprafols:
I have fixed the problems in issue #753 by changing the comparison made in the function fill_neighs of `picca/cf.py`. Now it does not check for the flag `x_correlation` when deciding to use the second set of deltas to compute the neighbours. Instead, it checks for the presence of the second set.

I also discovered that the same bug was occuring in `picca/co.py` so I also fixed it as well.

### DIFF
--- a/py/picca/cf.py
+++ b/py/picca/cf.py
@@ -84,7 +84,7 @@ def fill_neighs(healpixs):
                 nside, [delta.x_cart, delta.y_cart, delta.z_cart],
                 ang_max,
                 inclusive=True)
-            if x_correlation:
+            if data2 is not None:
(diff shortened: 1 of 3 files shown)


Waelthus (approved):

iprafols:
Tests on NERSC show that the problem is fixed, travis tests are probably failing due to h5py version change

Waelthus (approved):

Waelthus:
ok, found the issue regarding testing: the new h5py will create object arrays where the old one created byte-string arrays for "list of fixed pars" (and other lists of strings). I'll let the tests check for this and convert the dtypes.

Issue #752, opened 2020-10-27
Picca automatic testing


Waelthus:
Currently automatic testing requires a run of all tests. 
We should spilit this up into subtests to allow for easier debugging and eventually simpler addition of tests and thus better test coverage.
This requires tests to be independent from previous outputs which are mostly already saved in the repo anyway.

iprafols:
This will soon be fixed for the continuum fitting part of the code when the delta_extraction branch is merged

iprafols:
PR #878  partially solved this for the continuum fitting. Corrections and Masks testing still needs to be improved

corentinravoux:
@iprafols Should we close this issue or you still think tests should be more separated ?

iprafols:
Let's close it

PR #755, opened 2020-11-03
Disentangle testing

Waelthus:
First attempts to attack #752. Currently only splitting the test into multiple files by topic, adding outputs of `export*`, `picca_deltas` and the randomly generated catalogues to the data dir such that tests can run autonomously without requiring previous tests to be run.

The only exception is the test for `picca_deltas` which still needs to generate random mock spectra as we probably don't want to save a couple GB of fake data in the repo...


py/picca/test/test_1_deltas.py
```diff
@@ -0,0 +1,427 @@
...
+            for fname in glob.glob(
+                    f'{self._branchFiles}/Products/Delta_Pk1D_MiniSV/Delta/delta-*.fits.gz'
+            ):
+                path2 = fname
+                path1 = f"{self._masterFiles}/test_delta/Delta_Pk1D_MiniSV/{os.path.basename(fname)}"
+                self.compare_fits(path1, path2, "picca_deltas.py")
+        return
+
+
+    def test_delta_Pk1D(self):
```
iprafols, line 386: ordering of the methods is weird. Could we put all the test methods together?

iprafols (commented):
Related question: are you also planning on not using the subprocess.call and calling the scripts directly? or do you want to do this later on?

py/picca/test/test_1_deltas.py
Waelthus: yes, I didn't clean this up yet, thanks!

Waelthus:
> Related question: are you also planning on not using the subprocess.call and calling the scripts directly? or do you want to do this later on?

I'll try implementing this for picca_deltas and see how well it works. It's just not in yet. Structurized it first. But this could also be a new PR which probably is a cleaner change.

Waelthus:
Ok, making things importable does create a bunch of extra hassle when it comes to installation as by default depending on python configuration you couldn't import the scripts in bin (e.g. on my notebook I can, but travis cannot). The simplest fix for this is moving the bin directory to be it's own package as in the last commits, or making it a subpackage of picca itself like the fitter.

One benefit is that ERRORS (and also warnings) can be summarized at the very end of a test run including a traceback (from the actual script not the test script that called subprocess) which makes it more useful. Atm I just use the argument parser to parse the function call, i.e. this requires a string list to be given to the function which is ugly, but it is a simpler change than properly splitting the parsing and the actual main routine, mainly because there are a lot of arguments (most of them typically being used with there default values).

So those changes would create a lot of changes in code structure which we should probably think about in more detail. But also make things more easy to debug or call from jupyter notebooks etc...

iprafols:
@Waelthus  what's the status of this branch? Are you still planning on working on it a bit more or should I go ahead with the revision?

Waelthus:
You can go ahead, the one thing I wanted to look at again was trying to get the imports of picca_* working without changing the bin directory to be within py/ (which works but is inelegant and demands from people to change their scripts).

I also noticed that travis has become much slower in executing the tests compared to a couple months back, i.e. they seem to queue for a while before actually being run, while earlier they basically started right after committing/PRing. At least it feels longer than before. Not sure if that might be an issue and if we could do much about it except for e.g. switching CI provider. But anything like this should be done in a seperate PR anyway...

Waelthus:
Ok, I did a couple more things here allowing to start testing for deltas, pk1d and fitter in the main process, for all the things left in test_cov that is prepared as well and should now only require changes to the actual test. Those are currently broken, I'll fix this later today.

Waelthus:
There're some more changes I made here:
-- test_3_cov: now imports the picca_* files into the main process, it also reloads cf and xcf where it matters (some variables are declared globally in there, so importing the package twice within the same process leads to variables from previous tests still lying around; I guess changing code to have those variables treated in a safer way might be some work, won't do that here)
-- test_1_deltas: now modifies seed at test execution where needed and not only at initial setup, this is required to allow those tests to be run in any order (previously the deltas test needed to be run before deltas_miniSV) as needed when executing e.g. on multiple CPUs, changed output data files accordingly
-- travis setup: now runs previously failing tests first, stopping after 2 failures
-- structure/setup.py: the bin directory is moved to py/picca/bin and setup.py is searching for executables there, this makes files be accessible via both imports and the shell, it's also a less weird setup than the previous usage of py/picca_bin

Waelthus:
@iprafols now I'll actually not change anything more before your pass...

Waelthus:
this has been merged into `test_github_action_problems` and thus will be merged with #789

Waelthus:
closes #752

Issue #758, opened 2020-11-27
picca_export and picca_export_coadd_zint give different results


jfarr03:
As far as I understand, running either `picca_export.py` or `picca_export_coadd_zint.py` on the same input data should yield the same exported correlation functions. However, when running these two scripts on the eBOSS DR16 analysis from mocks, using `picca_export` seems to work ok, whereas `picca_export_coadd_zint` gives a non-positive definite covariance warning. Using v4 of picca (as was used for the DR16 analysis), this issue goes away. 

There seem to be some slight differences between `picca_export` and `picca_export_coadd_zint` but I haven't found any explanation for this issue yet. I wonder whether any of the changes made to picca in the last 6 months might be causing it? I've added a guide to reproducing the issue below.

At any rate, perhaps this raises a question of whether we need two scripts that have overlapping functionality. Would it be better instead just to have `picca_export` and `picca_coadd_zint`? It would mean running more 2 scripts rather than 1 sometimes, but as we would like to run most of our analyses in an automated way I don't think this is really an issue? 

(In addition to the above, there are some issues with backwards-compatibility, and running `picca_export` on the DR16 mock correlations requires a "hack" of sorts to work properly. This isn't a major problem, but it might be good to add a workaround if we want to make re-analysis of the DR16 results easier?)

################################################################################

To reproduce the issues, go to `/global/cfs/projectdirs/desi/mocks/lya_forest/picca/london/v9.0/test_export` at NERSC, and then run the following with the master branch of picca installed:

`
picca_export.py --data eboss_correlations/eboss-0.0/cf_z_0_10.fits --out cf_exp_eboss_0.0_m_e.fits.gz --dmat eboss_correlations/eboss-0.0/dmat_z_0_10.fits    
`
`                                
picca_export_coadd_zint.py --data eboss_correlations/eboss-0.0/cf_z_0_10.fits --out cf_exp_eboss_0.0_m_ecz.fits.gz
`

The `picca_export` command should raise an error relating to the keyword "OMEGAM" in the header, whereas the `picca_export_coadd_zint` should run properly but return a warning that the covariance is not positive definite.

To address the issue with "OMEGAM", switch to the picca branch `export_backwards_compatible_dr16`. This just comments out the offending lines of `picca_export` as a temporary solution. Once this branch is installed, then run:

`
[...]

PR #761, opened 2020-11-27
Fix picca export coadd zint

jfarr03:
Address #758 .


jfarr03:
@iprafols this fixes a bug in `picca_export_coadd_zint` - without this fix the script gives unnecessary warnings about non-positive definite covariances

iprafols (approved):

Issue #764, opened 2020-12-10
Weights when coadding distortion matrices


jfarr03:
When coadding correlations and associated distortion matrices from several redshift bins, the distortion matrices are combined in a weighted sum where the weights are taken from the associated correlation:

https://github.com/igmhub/picca/blob/53828dc007b580c91504be22b922cc3dcabfabf7/bin/picca_export_coadd_zint.py#L93

Is there a particular reason for doing this rather than using the weights from the distortion matrix calculation itself? 

I'm currently updating this script as discussed at the telecon on 2nd December, and so can switch to using the dmat weights instead if we would like to.

iprafols:
They should be the same, shouldn't they? If they are the same, then doing it this way optimizes the code. That said, since this is a very fast bit of the code, I guess you can change it to using the dmat weights for more readability

jfarr03:
Statistically yes I guess they should but I don't know if they'd be literally identical? I noticed that there's a comment a few lines below this which seems to be to do with weights (though I'm not sure _exactly_ what it means or where it came from):

https://github.com/igmhub/picca/blob/53828dc007b580c91504be22b922cc3dcabfabf7/bin/picca_export_coadd_zint.py#L95

At any rate, we need to open up both the correlation and dmat files so I think the speed difference would be negligible really?

iprafols:
In that case I'd say to change it to improve readability (but I'd double check that the speed difference is indeed negligible)

jfarr03:
Sure I'm just writing the changes into a branch now and will run some tests to check both that the output is the same and that the speed difference is negligible

PR #768, opened 2020-12-17
Improve coadd export

jfarr03:
This PR deletes the script `picca_export_coadd_zint.py` as it has duplicated functionality with `picca_export.py`. In its place, a new script `picca_coadd_zint.py` is created, which largely replicates the coaddition procedures of `picca_export_coadd_zint.py`, but leaves the process of "export"ing to `picca_export.py`.

There are a small number of changes to the coaddition procedure:

- When coadding distortion matrices, the weights stored in the `dmat` file are used, rather than weights from the corresponding correlation files (as discussed in #764). This causes some slight differences in the output distortion matrices, though they are smaller than 1% for all elements from my tests. I'm not aware of any reason that we should not use the distortion matrix weights, but would be interested in others' views! Tests also suggest there is no discernible difference in the time this takes. 

- When coadding correlations, a check is carried out to see if all input correlations have measurements from the exact same HEALPix pixels. If this is true (as is often the case), correlations are added straightforwardly as arrays rather than using a dictionary to carefully match HEALPix pixels. The array approach is ~20x faster from my tests.

- Additional checks are put in place to make sure that the input correlations and distortion matrices have the same properties (e.g. same nside for HEALPix pixelisation)

- Distortion matrices are no longer found by replacing the string "cf" with the string "dmat" in the path to the correlation files, as was the case before. A list of distortion matrix files is now specified at input. This allows for more flexible naming and locations of files, which is potentially useful when analysing mocks as we will have separate correlations for each realisation, but probably only one set of distortion matrices for all realisations. It is also perhaps slightly less robust than the previous approach as it is possible to point the script to the wrong dmat files - I can add a warning to make it more robust if this is a concern.

These are all relatively minor changes, but please let me know if there are any issues. The script is not quite fully ready to be merged as there are some parts of the old code that are retained in comments for easier comparison with old methods.


bin/picca_coadd_zint.py
```diff
@@ -0,0 +1,458 @@
...
+import numpy as np
+
+from picca.utils import userprint
+
+
+def coadd_correlations(fi,fout):
+    """Coadds correlation functions measured in different redshift intervals.
+
+    Args:
+        fi: list of strings
```
iprafols, line 15: I'd suggest to rename this variable to `input_files`

bin/picca_coadd_zint.py
```diff
@@ -0,0 +1,458 @@
...
+            List of the paths to the input correlations to be coadded.
+        fout: string
```
iprafols, line 17: Similarly this could be renamed to `output_file`

bin/picca_coadd_zint.py
```diff
@@ -0,0 +1,458 @@
...
+    r_trans_max = -1.e6
+    z_cut_min = 1.e6
+    z_cut_max = -1.e6
+
+    # check to see if all files have the same HEALPix pixels
+    same_healpixs = True
+    for file in fi:
+        with fitsio.FITS(file) as hdul:
+            if len(hdul[2]['HEALPID'][:]) != len(healpixs):
+                same_healpixs *= False
```
iprafols, line 50: why the * ? Why not simply do `same_healpix = False`?

bin/picca_coadd_zint.py
```diff
@@ -0,0 +1,458 @@
...
+            elif (hdul[2]['HEALPID'][:] != healpixs).all():
+                same_healpixs *= False
```
iprafols, line 52: same here

bin/picca_coadd_zint.py
```diff
@@ -0,0 +1,458 @@
...
+    # loop over files
+    for file in fi:
+
+        # open the file and read the header
+        userprint("coadding file {}".format(file))
+        hdul = fitsio.FITS(file)
+        header = hdul[1].read_header()
+
+        # check that the header properties match those from the first file
+        for h in headers_to_check_match:
```
iprafols, line 73: maybe rename `h` for `entry` or `key`?

bin/picca_coadd_zint.py
```diff
@@ -0,0 +1,458 @@
...
+            assert header[h] == headers_to_check_match_values[h]
+
+        # add weighted contributions from this file to coadd variables
+        weights_aux = hdul[2]['WE'][:]
```
iprafols, line 77: just to double check, `weights_aux` and `weights_total_aux` here are used to sum over all the different healpix, aren't they? If so, why does it not matter the value of `same_healpixs`?  Am I missing something here?

bin/picca_coadd_zint.py
```diff
@@ -0,0 +1,458 @@
...
+            healpixs = hdul[2]['HEALPID'][:]
+            for index, healpix in enumerate(healpixs):
+                userprint(" -> coadding healpix {}".format(healpix),end='\r')
+                if healpix in xi:
+                    xi[healpix] += hdul[2]["DA"][:][index] * weights_aux[index]
+                    weights[healpix] += weights_aux[index, :]
+                else:
+                    xi[healpix] = hdul[2]["DA"][:][index] * weights_aux[index]
+                    weights[healpix] = weights_aux[index]
+            print('')
```
iprafols, line 106: I think we should remove this print or, at least, convert it to `userprint`

bin/picca_coadd_zint.py
```diff
@@ -0,0 +1,458 @@
...
+                  names=['HEALPID', 'WE', 'DA'],
+                  comment=['Healpix index', 'Sum of weight', 'Correlation'],
+                  header=header2,
+                  extname='COR')
+
+    results.close()
+
+    return
+
+def coadd_dmats(fi,fout):
```
iprafols, line 213: I'd suggest similar variable naming changes as in the previous function

bin/picca_coadd_zint.py
```diff
@@ -0,0 +1,458 @@
...
+        r_par_dmat = np.zeros(hdul[2]['RP'][:].size)
+        r_trans_dmat = np.zeros(hdul[2]['RT'][:].size)
+        z_dmat = np.zeros(hdul[2]['Z'][:].size)
+        num_pairs_dmat = 0.
+
+        # get values of header entries to check with other files
+        header = hdul[1].read_header()
+        headers_to_check_match_values = {h:header[h] for h in headers_to_check_match}
+
+    ### OLD WEIGHTS
```
iprafols, line 240: I think it'd be best to remove these comments before merging

bin/picca_coadd_zint.py
```diff
@@ -0,0 +1,458 @@
...
+    # loop over files
+    for file in fi:
+
+        # open the file and read the header
+        userprint("coadding file {}".format(file))
+        hdul = fitsio.FITS(file)
+        header = hdul[1].read_header()
+
+        # check that the header properties match those from the first file
+        for h in headers_to_check_match:
```
iprafols, line 264: similar name change suggestion as in the previous function

bin/picca_coadd_zint.py
```diff
@@ -0,0 +1,458 @@
...
+            assert header[h] == headers_to_check_match_values[h]
+
+        # add weighted contributions from this file to coadd variables
+
+        ### OLD WEIGHTS
```
iprafols, line 269: same here, I'd remove the comments before merging

bin/picca_coadd_zint.py
```diff
@@ -0,0 +1,458 @@
...
+        #    weights_total_aux = hdul_corr[2]['WE'][:].sum(axis=0)
+        #dmat += hdul[1]['DM'][:] * weights_total_aux[:,None]
+        #weights_total += weights_total_aux
+        ###
+        weights_dmat_aux = hdul[1]['WDM'][:]
+        dmat += hdul[1]['DM'][:] * weights_dmat_aux
+        weights_dmat += weights_dmat_aux
+        ###
+
+        ### OLD WEIGHTS
```
iprafols, line 280: same here

bin/picca_coadd_zint.py
```diff
@@ -0,0 +1,458 @@
...
+        r_par_min = np.min([r_par_min,header['RPMIN']])
+        r_par_max = np.max([r_par_max,header['RPMAX']])
+        r_trans_max = np.max([r_trans_max,header['RTMAX']])
+        z_cut_min = np.min([z_cut_min,header['ZCUTMIN']])
+        z_cut_max = np.max([z_cut_max,header['ZCUTMAX']])
+
+        hdul.close()
+
+    # normalize
+    ### OLD WEIGHTS
```
iprafols, line 304: same here

bin/picca_coadd_zint.py
iprafols, line 213: Just double checking. Why does this function does not check for `same_healpix` as is done in the previous function? Is it because the distortion matrix is already saved merging the information  from the different healpixs?

bin/picca_coadd_zint.py
jfarr03: Agreed, I'll address this in a pull shortly

bin/picca_coadd_zint.py
jfarr03: Not sure to be honest, I'll change this in a commit shortly to make it simpler

bin/picca_coadd_zint.py
jfarr03: Agreed, I'll address this in a commit shortly

bin/picca_coadd_zint.py
jfarr03: Yes `weights_aux` is the weights in each separation bin and in each pixel, whereas `weights_total_aux` has been summed over pixels.

The value of `same_healpixs` doesn't matter as `weights_total_aux` is only used to provide appropriate weights for calculation the average redshift/separation etc of each separation bin, and we're only interested in the mean over all pixels in those cases. I.e. it doesn't matter if two correlations have measurements in completely different sets of pixels, we still would compute the mean z/r_par/r_trans in each bin over all pixels. Does that make sense? 

At any rate, using `same_healpixs` doesn't change any of the quantities being computed relative to the old method, it just makes things much much faster when our correlations have the exact same footprint

bin/picca_coadd_zint.py
jfarr03: I think this print is needed otherwise subsequent printed output is placed on top of the information about which pixels are being coadded. Will change to `userprint` in a commit thought

bin/picca_coadd_zint.py
jfarr03: Agreed, I'll address this in a commit shortly

bin/picca_coadd_zint.py
jfarr03: Yes we don't measure the distortion matrix in HEALPix pixels, I guess because we're not interested in computing its variance via subsampling

jfarr03:
Hi @iprafols , thanks for taking a look at everything - I've addressed all comments either with changes or in replies. Let me know what you think :)

bin/picca_coadd_zint.py
iprafols: Yes, this makes sense. I just wanted to make sure I was not missing anything

iprafols (approved):

PR #770, opened 2020-12-23
Fixed iminuit version

jfarr03:
Fixed iminuit version to 1.4.9 while tests on v2.x are ongoing as v1.5.x has know issues (see https://github.com/scikit-hep/iminuit/releases), and v2.x has different syntax.


iprafols (approved):

londumas:
@jfarr03 and @iprafols could you tell me what was the issue with the latest iminuit?

iprafols:
@londumas, the 1.5.x series of iminuit is known to have some issues and so we decided to revert back to 1.4.9 until version 2 is stable

londumas:
@iprafols, Which problems did you observe? The unittests were totally fine with the latest release.

iprafols:
We found some DR16 fits that were not converging anymore. Also, the fits were giving different results. Travis tests passed as the results of the new fits on the test set were updated as well.

londumas:
> Travis tests passed as the results of the new fits on the test set were updated as well.

Well that is not true, I actually looked and checked that the results were the same to the machine precision.

Did you take the time to understand were were the errors? The differences? What was causing them?

iprafols:
At a later stage, the travis tests failed due to some updates from the Travis machine. We checked that it was the change of libraries as previous tests that passed were failing again. Thus, we decided to revert to a more stable version of iminuit

londumas:
@iprafols, any way it is not my job anymore. Have fun.

PR #771, opened 2021-01-13
Pass drq mode to xcf

Waelthus:
Changes to make picca_xcf work with SV catalog.
New mode argument (defaulting to sdss) will be passed to read_objects and read_drq (the latter had this implemented already), using TILE/NIGHT/FIBER to initialize QSO objects PLATE/MJD/FIBERID attributes for the moment.

Also followed @SamYoules suggestion to flip the order of readin for catalog, spectra.

Still needs to be tested.


cramirezpe (approved):
Tested on Blanc data release obtaining the same output as a previous picca version.

Waelthus:
The test suite does run through on my notebook now as well, so this should allow running all codes using a drq also with the new cats without breaking sdss compatibility. We should probably include tests as well performing the whole analysis chain on a desi-like cat and dataset (so far this is only done for picca_deltas) to ensure we don't break this again. But that's for another PR.

SamYoules (approved):
Change to order of reading catalog and deltas is fine.

2199433g (approved):

PR #772, opened 2021-01-19
Changed zcat default downsampling to None

SamYoules:
I changed the default argument for downsampling-nb in tutorials/desi/picca_convert_zcat.py to None, as it was crashing if the mocks had less than the default value of 700,000.

### DIFF
--- a/py/picca/converters.py
+++ b/py/picca/converters.py
@@ -343,7 +343,7 @@ def desi_from_ztarget_to_drq(in_path,
                     cat[key] = cat[key][w]
                 userprint((" and downsampling : nb object in cat = {}, nb z > "
                        "{} = {}").format(cat['RA'].size, downsampling_z_cut,
-                                        (cat["Z"] > downsampling_z_cut).sum()))
+                                        z_cut_num))
             else:
                 userprint(("WARNING:: Trying to downsample, when nb QSOs with "
                            "z > {} = {} and downsampling = {}").format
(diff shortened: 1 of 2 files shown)


Waelthus:
While this change would make it not crash with default values for undersized cats, it would also make the script not resample anymore for oversized ones. Not sure if anyone would be affected by the change in default, but I would think that an actual check if the cat is larger than the value submitted would be the better approach as it fixes the underlying issue. Could be done e.g. here: https://github.com/igmhub/picca/blob/334235699c9667bc78d8f5209fc74330ce3c75da/py/picca/converters.py#L329

SamYoules:
> While this change would make it not crash with default values for undersized cats, it would also make the script not resample anymore for oversized ones. Not sure if anyone would be affected by the change in default, but I would think that an actual check if the cat is larger than the value submitted would be the better approach as it fixes the underlying issue. Could be done e.g. here:
> 
> https://github.com/igmhub/picca/blob/334235699c9667bc78d8f5209fc74330ce3c75da/py/picca/converters.py#L329

Thanks Michael. How about if after line 336, I put:

if select_fraction > 1.0:
    select_fraction = 1.0

along with a warning that there are fewer than z>downsampling_z_cut in the catalog?

Waelthus:
not sure if that would work right, then it would still choose objects until it has gotten all of them. How about doing something like:
```
if select_fraction<1:
  the rest of the stuff within the downsampling if condition   
else:
  warning...```

SamYoules:
Ok, thanks.

Waelthus (approved):

Waelthus:
looks ok to me.

PR #773, opened 2021-01-26
Add DR16 BAO scans for external use with script and documentation

bzicher:
Add an extra folder with BAO scans run on finer grids and a script to extract just the alphas and chi2/likelihood for use in external libraries. Also added output text files with just the BAO scans and documentation to explain the difference between these, the other folders here that contain the scans with all the parameters and the official eBOSS likelihoods (which are in DM/DH instead of alphas).


andreicuceu:
By the way, this is @andreicuceu doing the pull request. I accidentally started it from another browser where I was logged in with another github account.

PR #774, opened 2021-01-26
Add DR16 BAO scans for external use with script and documentation

andreicuceu:
Add an extra folder with BAO scans run on finer grids and a script to extract just the alphas and chi2/likelihood for use in external libraries. Also added output text files with just the BAO scans and documentation to explain the difference between these, the other folders here that contain the scans with all the parameters and the official eBOSS likelihoods (which are in DM/DH instead of alphas).


data/duMasdesBourbouxetal2020/BAO_for_external_use/README
```diff
@@ -0,0 +1,34 @@
+1. The official eBOSS DR16 scans can be found here: https://svn.sdss.org/public/data/eboss/DR16cosmo/tags/v1_0_0/likelihoods/
```
andreufont, line 1: @andreicuceu - I would clarify that in that link they can find the likelihoods used in the "DR16 cosmology paper" (and add the arXiv number to Alam et al. 2020).

I would also clarify in the README which are the scans used in the DR16 Lya BAO analysis, since these are not the same.

data/duMasdesBourbouxetal2020/BAO_for_external_use/README
```diff
@@ -0,0 +1,34 @@
...
+
+This is the README for those files:
+The DR16 Lyman-alpha BAO likelihoods are provided separately for the auto (LYAUTO) and cross correlation (LYXQSO). They can be treated as independent. The format for each is:
+
+Column 1:  (DM/rd) at z=2.334
+Column 2:  (DH/rd) at z=2.334
+Column 3:  likelihood (relative to the best point on the grid)
+
+2. Files containing the full scan output with all parameters can be found here: https://github.com/igmhub/picca/tree/master/data/duMasdesBourbouxetal2020. Note those results were run on a smaller grid compared to the official results.
```
andreufont, line 10: Which files are these? You should list the file names, otherwise it is confusing. Where these the files used in the Lya BAO paper? If so, say so.

andreufont: Are these HDF5 files? If so, specify it.

andreufont (commented):
I've added a couple of comments to the README, but I think it looks good. I was able to run the script and reproduce a couple of files. 

It would be good to send an email to:
 - Jim, as eBOSS chair, to check that he is ok with this
 - Ignasi & Michael, as Picca developers, to let them know of the PR
 - Pablo, to make sure this reflects what is in CosmoMC and in Cobaya, and that both have DR16 modules already available and in in their master branch.
 - James, to ask him how to update the MontePython module with the DR16 files

andreicuceu:
@andreufont checkout the updated documentation. I will also send the emails.

andreufont (approved):
Looks great, happy for this to be merged (I guess after the call tomorrow?).

PR #775, opened 2021-02-09
change io to use spectra from deep by default

Waelthus:
This is to allow using the deep spectra by default in DESI SV mode, there's 2 flags for picca_deltas to allow changing to all or allow to use the previous behaviour, i.e. selecting only spectra from specific nights.

### DIFF
--- a/py/picca/io.py
+++ b/py/picca/io.py
@@ -1083,6 +1083,7 @@ def read_from_minisv_desi(in_dir, catalog, pk1d=None, usesinglenights=False, use
             f"{entry['PETAL_LOC']}-{entry['TILEID']}-{entry['NIGHT']}"
             for entry in catalog
         ]
+        #this uniqueness check is to ensure each petal/tile/night combination only appears once in the filelist
         petal_tile_night_unique = np.unique(petal_tile_night)
     
         filenames = []
@@ -1090,6 +1091,7 @@ def read_from_minisv_desi(in_dir, catalog, pk1d=None, usesinglenights=False, use
           for ptn in petal_tile_night_unique:
             if ptn in os.path.basename(f_in):
                 filenames.append(f_in)
+                break
     else:
         if useall:
             files_in = glob.glob(os.path.join(in_dir, "**/all/**/coadd-*.fits"),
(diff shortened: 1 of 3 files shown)


py/picca/io.py
```diff
@@ -1069,19 +1076,39 @@ def read_from_minisv_desi(in_dir, catalog, pk1d=None):
...
+            files_in = glob.glob(os.path.join(in_dir, "**/all/**/coadd-*.fits"),
+                         recursive=True)
+        else:
+            files_in = glob.glob(os.path.join(in_dir, "**/deep/**/coadd-*.fits"),
+                         recursive=True)
+        petal_tile = [
+            f"{entry['PETAL_LOC']}-{entry['TILEID']}"
+            for entry in catalog
+        ]
+        petal_tile_unique = np.unique(petal_tile)
```
andreufont, line 1104: Why do we need this "unique" call here? Do we have expect more than one file for this combination of petal and tile?

andreufont (commented):
The code looks good, but I haven't tested the branch. @2199433g - could you take a look at it?

2199433g (approved):
Picca_deltas works for me with --use-all!

Waelthus:
ok, seems to do its job with single nights or the new default to use deep stacks as well

py/picca/io.py
Waelthus: yes, this goes over the objects in the cat not over the files, only files that have objects in the cat will then be read from disk at all.

andreufont:
We can discuss this at the telecon today, and merge right after? 

I'm still confused about why we need to use petal_tile_unique instead of directly petal_tile, but I guess that's ok. What would happen if we didn't add that line? Maybe add a comment explaining this to the code?

Waelthus:
I think that without the line files might appear multiple times in the tile list and thus might be read-in multiple times. But I didn't test for this specifically...

andreufont:
Even if we are only reading coadd files under /deep/ ?

If this happened, it might be better to write an error message and crash, unless we understand why this would happen.

Waelthus:
it happens due to the 2-level for-loop after that loops over all files and for each file over all cat entries. I now let it break when the first match is found so that each file that has at least one match in the cat should appear exactly once in the filelist. The unique is thus not strictly needed, but should speed up (but this gain is probably unimportant) the double loop that follows right after (which is ugly, but was the easiest I came up with when writing this piece last year).

andreufont:
I see. We can keep it as it is for now, but it would be good if we found a nice way to do this long-term. We should be using this function quite a lot in the next few months.

Thanks for getting this started, by the way, it has been very useful.

Waelthus:
Yes, I didn't want to be overly sophisticated here, also because we're likely replacing those routines soon.

PR #776, opened 2021-02-12
save iter file at each iteration

Waelthus:
This saves the iter (or delta_attributes) file after each continuum fitting iteration. It also computes the stack after each iteration. If the latter is too expensive we could only do this for the last iteration.

### DIFF
--- a/bin/picca_deltas.py
+++ b/bin/picca_deltas.py
@@ -776,10 +776,10 @@ def main():
             ],
                           names=['loglam_rest', 'mean_cont', 'weight'],
                           extname='CONT')
-            var_pipe_values = np.broadcast_to(var_pipe_values.reshape(1, -1),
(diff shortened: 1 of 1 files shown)


iprafols (approved):
changes look fine

iprafols (changes requested):
Why are the tests not passing?

Waelthus:
I'm trying to figure that out. I'm getting a numpy error in the end about a broadcasting (1,100) -> (10,10). Not sure where that comes from in detail yet...

Waelthus:
was an issue where `np.broadcast` was used on `var_pipe_values` and overwrote the array, for all but the last iteration this was replaced in the iteration, but for the last one only mean_cont is computed and the array isn't replaced, which made the broadcast fail on this. Now I just generate a new variable for writing this output

iprafols (approved):

PR #777, opened 2021-02-16
Adapt for new mocks without reso matrix for each spectrum

Waelthus:
This reads the resolution matrix from the truth file to allow running picca deltas in Pk1d mode for light quickquasar spectra. This is skipped for the BAO analysis which never uses the resolution anyway

### DIFF
--- a/py/picca/io.py
+++ b/py/picca/io.py
@@ -1008,7 +1008,8 @@ def read_from_desi(in_dir, catalog, pk1d=None):
                     spec["RESO"] = hdul[f"{color}_RESOLUTION"].read()
                 else:
                     try:
-                        filename_truth=filename.replace('spectra-','truth-')
+                        filename_truth=f"{in_dir}/{healpix//100}/{healpix}/truth-{in_nside}-{healpix}.fits"
+
                         with fitsio.FITS(filename_truth) as hdul_truth:
                             spec["RESO"] = hdul_truth[f"{color}_RESOLUTION"].read()
                     except IOError:
--- a/py/picca/io.py
+++ b/py/picca/io.py
@@ -1212,7 +1212,10 @@ def read_from_minisv_desi(in_dir, catalog, pk1d=None, usesinglenights=False, use
                 flux = spec['FL'][w_t].copy()
 
                 if pk1d is not None:
-                    reso_sum = spec['RESO'][w_t].copy()
+                    if len(spec['RESO'])>1:
+                        reso_sum = spec['RESO'][w_t].copy()
+                    else:
+                        reso_sum = spec['RESO'][0].copy()
                     reso_in_km_per_s = np.real(
                         spectral_resolution_desi(reso_sum, spec['log_lambda']))
                     exposures_diff = np.zeros(spec['log_lambda'].shape)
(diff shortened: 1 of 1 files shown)


py/picca/io.py
```diff
@@ -1005,6 +1006,19 @@ def read_from_desi(in_dir, catalog, pk1d=None):
...
+                elif pk1d is not None:
+                    filename_truth=f"{in_dir}/{healpix//100}/{healpix}/truth-{in_nside}-{healpix}.fits"
+                    try:
+                        with fitsio.FITS(filename_truth) as hdul_truth:
+                            spec["RESO"] = hdul_truth[f"{color}_RESOLUTION"].read()
+                    except IOError:
+                        userprint(f"Error reading truth file {filename_truth}")   
+                    except KeyError:
+                        userprint(f"Error reading resolution from truth file for pix {healpix}")
+                    else:
```
iprafols, line 1018: Probably there is an identation error here

py/picca/io.py
```diff
@@ -1005,6 +1006,19 @@ def read_from_desi(in_dir, catalog, pk1d=None):
...
+                        if not reso_from_truth:
```
iprafols, line 1019: I don't see the usefulness of having this here. Maybe the whole block goes somewhere else? I mean the fact that you change it `reso_from_truth` to True does not imply anything here as the reading is done earlier. Maybe this block goes as an else at line 1009 and the elif there is really and if with and additional condiditon `if pk1d is not None and reso_from_truth:`

py/picca/io.py
```diff
@@ -1033,7 +1047,10 @@ def read_from_desi(in_dir, catalog, pk1d=None):
                 flux = spec['FL'][w_t].copy()
 
                 if not pk1d is None:
-                    reso_sum = spec['RESO'][w_t].copy()
+                    if not reso_from_truth:
+                        reso_sum = spec['RESO'][w_t].copy()
```
iprafols, line 1051: what is w_t and where is it defined?

iprafols (changes requested):

py/picca/io.py
Waelthus: this is only excecuted when it actually found a truth file with the right info (i.e. no exception on the command before), that's where the else comes in. The if is then in case the variable hasn't been set before to set it and print a message.

py/picca/io.py
Waelthus: `w_t = np.where(targetid_spec == entry[id_name])[0]` from ~15 lines further up, it's the index of the currently used spectrum (RESO is a (nspec,ndiag,npix) array). If `reso_from_truth` was set RESO is only an (ndiag,npix) array and the selection is not needed

Waelthus: If reso does not exist in any form at this point this will fail.

py/picca/io.py
iprafols: Ok, I see, I thought might be the else for the if. Just double checking

py/picca/io.py
iprafols: This is explained by the previous block being properly indented

py/picca/io.py
iprafols: ok

iprafols (approved):

PR #780, opened 2021-03-18
Update the template pk computation

andreicuceu:
Just a few updates to the picca_compute_pk_pksb.py script. We are now writing to file the sigma8 and f (growth rate) values at both redshift 0 and the reference redshift. Previously we were writing sigma8(z=0) and f(z=zref) with no reference to the redshift.
Also fixed a bug when calling linspace and updated to logspace for simplicity.


andreufont (approved):
Yes, definitely better to store both f(z) and sigma_8(z) at the relevant redshifts.

andreicuceu:
Someone else needs to merge this, as I'm not authorized to do it.

andreufont:
@iprafols - I'd be in favor of giving Andrei the power to merge PR in Picca.

iprafols:
@andreufont, I agree

PR #782, opened 2021-05-12
reinstate before error in push

londumas:
reinstate before error in push.
Made an error by pushing directly to master. Reinstate the previous master

### DIFF
--- a/bin/picca_Pk1D.py
+++ b/bin/picca_Pk1D.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+#!/usr/bin/env python
 """Compute the 1D power spectrum
 """
 import argparse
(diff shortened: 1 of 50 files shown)


iprafols (changes requested):
I'm not sure we want to remove these changes. Basically it is just a matter of replacing dots in imports by actual package names. It makes more sense to keep the package names. Maybe the only thing I would change back is the requirements: for some reason your accidental push reverted to older requirements

Waelthus:
I agree with @iprafols here, relative imports with the dot notation are not always fully stable anyway, so it might make sense to do this change, but please change back the req.

londumas:
Sorry, I went to fast.
This Pr added relative imports which are the proper and way more stable way to do packaging.
I messed up the Pr processes, due to remaining privilegies. If you do not like the PR, do not hesitate to reinstate the status of master before. And again sorry for my messed-up

PR #784, opened 2021-05-31
new method for convergence

moonlovist:
add jean-marc's convergence method;
Fix eta and fudge parameters for mocks


andreicuceu:
Hi @iprafols, I noticed that a lot of the changes in this pull request are Jupyter notebook checkpoints. These shouldn't go into master, so I was thinking we should add a flag to .gitignore about these checkpoints. As it's very relevant to this pull request you could just include it in here. If this is gonna take a while before it goes into master I can do a separate pull request just for that. Let me know what you think

iprafols:
Hi @moonlovist before merging we need to make sure the tests pass. The reason the tests are not passing in your case is that  we are actually modifying what the code does, so we expect the deltas and correlations to be different. However, we need to update the test suite accordingly. You would need to create the tests files for your run so that the travis test pass. It'd be good to add some plots or something to this PR so that we can be convinced that the new test files are correct

iprafols:
Hi @andreicuceu, 
Yes, I totally agree, but it seems this branch might take a while to merge so maybe we can go ahead with a separate PR

iprafols (changes requested):
We need the tests to pass before we can consider merging this. Also, this is a PR on the old version of picca_deltas, isn't it?

PR #785, opened 2021-06-02
Add healpix analysis for minisv mode

Waelthus:
adding a mode to create healpix based coadds on tilebased DESI data.


py/picca/io.py
```diff
@@ -340,6 +352,8 @@ def read_data(in_dir,
     if mode in ["desi", "spcframe", "spplate", "spec", "corrected-spec"]:
         if mode == "desi":
             pix_data = read_from_desi(in_dir, catalog, pk1d=pk1d)
+        elif mode == "desiminisv_healpix":
```
iprafols, line 355: what's the difference between modes "desi" and "desiminisv_healpix"? Why is the new mode necessary? Do we need to blind data from both modes (e.g. minimal blinding strategy)?

Waelthus: "desi" is reading from and storing to healpix based format
"desi-minisv" is reading from and storing to tile based format (not coadding objects across tiles)
"desiminisv_healpix" is reading from tile based format, coadding same objects across tiles and storing to healpix based format

Waelthus: we do not have any data in healpix format now, so until everest comes out the "desi" mode can only be used on mocks, also everest will afaik not provide Y1 data, but only SV data, so Y1 data cannot be used until July (?) using this mode when the first healpix coadds of Y1 data will probably appear.
the desiminisv_healpix however uses picca to do the coadding on a data format that already exists, so blinding would be an issue here already now

iprafols: So "desi" mode is able to read mocks and data (assuming there was data)?

Waelthus: I'm not completely sure yet (as there are no data in the exact final format yet), there might be some adjustments that need to be made...

Waelthus:
let me make double sure that it's actually exiting before merging

Waelthus:
Ok, now this should actually work

iprafols:
This branch was merged into blind_analysis, so this PR is no longer necessary

PR #786, opened 2021-06-08
Add blind analysis structure

andreicuceu:
We implement the minimal blinding strategy by adding a "blinding" string to all fits files created by picca that allows you to select the blinding strategy. For now the relevant ones are "none" or "minimal" (only added when reading DESI data), but new ones can be added after we decide on a long term blinding strategy. We also modify the name of the data columns (DELTA for deltas, and DA for the correlations) to add "_BLIND". This protects against older versions of picca accidentally running the unblinded analysis.

This branch also contains some small changes related to strategy B (blinding the coordinate cosmology), and two test blind templates. These changes don't do anything for now, and the two templates will be replaced in the future with the blind template that will actually be used for the relevant blinding strategy (if necessary).


Waelthus:
should we increase the version to 4.1 to allow easily checking if those changes have been included in an installation? With the `delta_extraction` (where we'd actually remove some old eBOSS functionality as well) being envisioned as `5.0`?

andreicuceu:
Yeah, I think it would be good to move to a new version. 4.1 is fine with me.

Waelthus:
Ok, from my point of view we can merge after the github actions PR is merged.

iprafols:
Fine with me too. I also agree about the version number

Waelthus (approved):
I did not check changes line by line, but given that different analyses work on the branch, and that this merge enables us to look at DESI data I'm approving this once tests are through.

PR #791, opened 2021-07-05
setup had issues

Waelthus:
The previous setup file made trouble when installing without developing mode (i.e. `pip install .` instead of `pip install -e .`). This should fix those issues


calumgordon (approved):
Works for me!

PR #793, opened 2021-07-23
changing coadding to always return the forest with data should there be one without data

Waelthus:
It looks like there was a bug in the coadding routine (see https://desisurvey.slack.com/archives/CTJ0GRSR3/p1627042008070600 and following). The coadding routine in  py/picca/data.py previously returned the current object should either it (self) or the forest to be coadded to it (other) contain no data. 
In case of coadding multiple bands this led to trouble when B (which is treated) first contains no useful data (e.g. when looking at the MgII forest that mostly lives in R/Z). I changed it to always return the forest with useful data from the coadding routine (note that it is also used as a method sometimes, this usecase still needs to be fixed).

### DIFF
--- a/py/picca/io.py
+++ b/py/picca/io.py
@@ -1296,7 +1296,7 @@ def read_from_minisv_desi(in_dir, catalog, pk1d=None, usesinglenights=False, use
                 if forest is None:
                     forest = copy.deepcopy(forest_temp)
                 else:
-                    forest.coadd(forest_temp)
+                    forest = forest.coadd(forest_temp)
             if not usehealpix:
                 if plate_spec not in data:
                     data[plate_spec] = []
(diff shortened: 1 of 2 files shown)


py/picca/io.py
```diff
@@ -1307,17 +1307,17 @@ def read_from_minisv_desi(in_dir, catalog, pk1d=None, usesinglenights=False, use
                     data[in_healpixs[w_t]] = []
                 #this might be slow, but would coadd objects with the same targetid even if on multiple tiles
                 do_append=True
+#                if forest.log_lambda is None:
+#                    breakpoint()
                 for index_coadd,forest_existing in enumerate(data[in_healpixs[w_t]]):
                     if forest_existing.thingid==forest.thingid:
-                        forest_existing.coadd(forest)
+                        forest_existing = forest_existing.coadd(forest)
```
Waelthus, line 1314: does this work? or would this be destroying inter-tile coadding and we need to use `data[in_healpixs[w_t]]=forest_existing.coadd(forest)` instead?

Waelthus: the method call is ok here, as in this case forest_existing is a pre-existing forest (and actually the `forest_existing=` is unneccessary), but need to check for validity when adding forests to data to ensure this.

calumgordon (approved):
This works for me, but so did the previous commit actually.

Waelthus:
yes, the changes in the last commit were just ensuring that things are treated right (before they could've in principle went wrong but didn't for the data at hand) and cosmetics

PR #794, opened 2021-07-28
Forest metadata

iprafols:
This branch makes as default the saving of a metadata table, and adds a few fields to the existing table

### DIFF
--- a/bin/picca_deltas.py
+++ b/bin/picca_deltas.py
@@ -52,6 +52,16 @@ def get_metadata(data):
                     column_values.append(0)
         tab[field] = np.array(column_values)
 
+    mean_delta2_values = []
+    for healpix in data:
+        for forest in data[healpix]:
+            if "delta" in forest.__dict__ and not forest["delta"] is None:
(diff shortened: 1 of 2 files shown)


Waelthus (changes requested):
ok, tests fail currently on Pk1d_MiniSV when comparing the stack of delta_attributes. Not sure why this could happen, the only change that is not just adding more info should be the mean_snr change in `data.py`...
They also seem to fail due to a difference in the metadata files produced on my notebook with the ones the github actions produce for one of the p0

Waelthus (approved):

Waelthus:
Note that the test on the new metadata is still failing (with at least one of the p1 values from the fitting being outside the defined accuracy bounds) when running on github actions, but not on my notebook (where the relevant test-outputs have been created), we need to figure out why this happens after the holidays.

Waelthus:
The results on the metadata seem to be actually bimodal, not sure why. But github is now storing results, so we can easily compare e.g. the failed and successful test results of this branch

PR #798, opened 2021-09-24
Read desi healpix coadd

alxogm:


andreufont:
@alxogm - we are ready to merge the "strategy E" blinding to the master branch.

andreufont:
I always make mistakes with GitHub... I closed the ticket accidentally!

andreufont:
What I meant is that we are ready to merge the corr_yshift branch (after the Picca telecon), and we should then finish any remaining detail and merge this one.

py/picca/io.py
```diff
@@ -349,14 +357,37 @@ def read_data(in_dir,
...
+            desi_prefix = 'spectra'
+            desi_nside = 16
+            pix_data, is_mock = read_from_desi(in_dir, catalog, desi_prefix, desi_nside, pk1d=pk1d)
+            
+        elif mode == 'desi_healpix':
+            pix_data=[]
+            for survey_type in ['SV1','SV3','DESI'] :
+                desi_nside = 64
+                if survey_type == 'DESI':
+                    catalog_ = catalog[catalog['DESI_TARGET']>0]
```
alxogm, line 371: I'm using the columns "DESI_TARGET", "SV1_DESI_TARGET" and "SV3_DESI_TARGET" to infer the location of the spectra files. I think it would be clearer if the catalog had a column indicating from which survey we should read the spectra. In the current catalog there are no repited TARGETIDS, and I think there are targets from more than one survey mode, but there could be in general. What do you think? Also, if we request that column to be in the catalog, we can simplify some of this lines...

py/picca/io.py
```diff
@@ -984,9 +989,17 @@ def read_from_desi(in_dir, catalog, pk1d=None):
...
         id_name = 'TARGETID'
-        plate_name = 'TILEID'
-        mjd_name = 'NIGHT'
-        fiberid_name = 'FIBER'
+        if 'TILEID' in catalog.colnames:
+            plate_name = 'TILEID'
+            mjd_name = 'NIGHT'
+            fiberid_name = 'FIBER'
+        else:
+            plate_name = 'TARGETID'
```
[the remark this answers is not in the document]
alxogm: I would like set these variable to None for desi data and mocks, but it currently breaks in the Forest instance, should I leave it as is for now and open an issue ticket to fix it later on?

alxogm (commented):
Almost done,

alxogm:
I think is almost ready for review (need to clean some unwanted spaces) and I raised a couple of questions to discuss in the telecom.

py/picca/io.py
andreufont: We should ask about this to the data team, but I think there are "SURVEY" or "FLAVOUR" keywords somewhere. Otherwise, DESI_TARGET might already be enough. At any rate, this seems good enough for now.

py/picca/io.py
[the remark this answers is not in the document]
andreufont: Right, let's just set these to whatever, and let's open an issue.

alxogm:
OK, I think is ready. I've tested it with a catalog produced with the latest update of `createqsocat.py` and everything seems to work well

py/picca/io.py
```diff
@@ -349,14 +361,28 @@ def read_data(in_dir,
...
-    if mode in ["desi_mocks","desi","desi_survey_tilebased", "spcframe", "spplate", "spec", "corrected-spec"]:
-        if mode in ["desi", 'desi_mocks']:
-            pix_data, is_mock = read_from_desi(in_dir, catalog, pk1d=pk1d)
-            if (not is_mock) and ('TILEID' in catalog.colnames) and np.any((catalog['TILEID']<60000)&(catalog['TILEID']>=1000)):
-                print("you are trying to run on DESI survey tiles!")
+    if mode in ["desi_mock","desi_healpix","desi","desi_survey_tilebased", "spcframe", "spplate", "spec", "corrected-spec"]:
+        if mode in ["desi_mock", "desi"]: #I still don't think we need two different modes since we are checking if truth files exist...
+            desi_nside = 16
+            desi_prefix = f'spectra-{desi_nside}'
+            pix_data, is_mock = read_from_desi(in_dir, catalog, desi_prefix, desi_nside, pk1d=pk1d)
```
andreufont, line 368: Should we also check here for blinding? This is supposed to be a mode to read DESI data, but we are not checking for blinding. If we are sure that this is only ever going to be used for mocks, then I'd suggest deleting the mode "desi" and only keeping "desi-mock", "desi-healpix" and "desi-survey-tilebased"

andreufont (approved):
I tested the branch on Everest, and it worked great. Thanks @alxogm !

I just added a minor comment about dropping mode="desi", but we could merge now and fix that later if you wanted.

andreufont:
@alxogm - just to be clear, we should first merge the blinding branch to this one, and then merge to master.

andreufont (approved):

py/picca/io.py
alxogm: Thanks for catching that the blinding check was not indented correctly, the intent is to always check for blinding. I'll fix it. I'm also in favor to deprecate "desi" mode, but I'm not sure if p1d uses it, so I'd leave it for now an open an issue...

alxogm:
> @alxogm - just to be clear, we should first merge the blinding branch to this one, and then merge to master.

sure, I'll fix the comment about the blinding and merge the blinding branch into this one.

py/picca/io.py
alxogm: OK, I ended up checking for blinding since just fixing the indentation was not the best way to do it.

PR #799, opened 2021-09-28
Improve raw analysis and parallel delta loading

andreicuceu:
Improve the raw analysis by adding an option for linear binning when computing the deltas. Also add parallel loading of transmission files/writing of deltas (only for raw analysis). I also parallelised read_deltas in io.py, so everything that uses it should automatically read deltas in parallel. We need to test this and maybe add one more flag for the weight type of the raw deltas.


py/picca/converters.py
```diff
@@ -362,17 +363,217 @@ def desi_from_ztarget_to_drq(in_path,
...
-                                             in_dir=None,
-                                             in_filenames=None,
-                                             lambda_min=3600.,
-                                             lambda_max=5500.,
-                                             lambda_min_rest_frame=1040.,
-                                             lambda_max_rest_frame=1200.,
-                                             delta_log_lambda=3.e-4,
-                                             max_num_spec=None):
-    """Convert desi transmission files to picca delta files
+def read_transmission_file(filename, num_bins, objs_thingid, lambda_min=3600.,
```
iprafols, line 366: This function should be in io.py

py/picca/converters.py
```diff
@@ -362,17 +363,217 @@ def desi_from_ztarget_to_drq(in_path,
...
+        deltas[healpix].append(
+            Delta(thingid[index2], ra[index2], dec[index2], z[index2],
+                  thingid[index2], thingid[index2], thingid[index2],
+                  rebin_log_lambda, rebin_ivar, None, rebin_flux, 1, None,
+                  None, None, None, None, None))
+
+    return deltas, stack_flux, stack_weight
+
+
+def write_delta_from_transmission(deltas, mean_flux, flux_variance, healpix, out_filename,
```
iprafols, line 500: This function should be in io.py

py/picca/io.py
```diff
@@ -1435,21 +1460,36 @@ def read_deltas(in_dir,
...
-            if num_data > max_num_spec:
+    for delta in results:
+        if delta is not None:
+            deltas += delta
+            num_data = len(deltas)
+            if (max_num_spec is not None) and (num_data > max_num_spec):
                 break
+    deltas = deltas[:max_num_spec]
+
+    # deltas = []
```
iprafols, line 1478: Removed commented version or add a comment specifying that it is the old way of loading data

py/picca/io.py
```diff
@@ -1435,21 +1460,36 @@ def read_deltas(in_dir,
...
-        num_data = len(deltas)
-        if max_num_spec is not None:
-            if num_data > max_num_spec:
+    for delta in results:
+        if delta is not None:
+            deltas += delta
+            num_data = len(deltas)
+            if (max_num_spec is not None) and (num_data > max_num_spec):
                 break
+    deltas = deltas[:max_num_spec]
```
iprafols, line 1476: This line is repeated (see line 1495). I would delete this here since there it also updates num_data

iprafols (changes requested):
I'm not sure the linear binning will work with just the changes implemented here. We need to address at least also the functions in prep_del.py and the coadding and rebinning in data.py. There may be other places where changes are required as well, to be double-checked. 

That said, I would be happy to merge once the minor points I signalled are fixed. I think the parallelistion in the reading is long overdue

bin/picca_export.py
```diff
@@ -185,6 +185,7 @@ def main(cmdargs):
             z_dmat = z.copy()
         hdul.close()
     else:
+        dmat_name = 'DM'
```
andreufont, line 188: Why is this change here? Has this leaked from your separate work on the blinding strategy?

andreufont:
Thanks @iprafols , I think we can implement these minor changes.

Note, however, that the changes in this PR regarding log-lambda vs lin-lambda only affect the "deltas from transmission", so I don't think you have to worry about prep_del.py or the rebinning in data.py, since these are not involved in the script. We directly write delta files without any continuum fitting and without creating any Forest object.

bin/picca_export.py
andreicuceu: The version of master I started working from had a bug related to blinding (now fixed I think) and this is me doing a temporary fix so this branch works. Before merging I'll pull master into this and get rid of this line.

andreufont: Sounds good.

andreufont:
@iprafols, @andreicuceu - now that we are done with the more pressing pull requests, it might be time to merge this one. We should also parallelise reading DESI data (it takes over an hour to read all spectra now in picca_deltas...), but we should do this in another PR.

andreufont:
Yeah, I wasn't convinced about putting these to the (already too large) io.py. 

Putting these into raw_analysis.py or raw_io.py could be an option.

iprafols:
Since they are io functions I would rather add them to a raw_io.py file than raw_analysis. 
I agree with @andreufont  that the io.py file is already too large

tutorials/desi/picca_convert_transmission.py
```diff
@@ -2,52 +2,62 @@
 
 import argparse
 
-from picca import converters
+from picca import raw_io
```
andreufont, line 5: Out of curiosity, why is this script in "tutorials/desi/"? I'd have imagine this would be in the main "/bins/" folder.

andreicuceu:
Hi @iprafols and @andreufont I moved the functions to a new raw_io.py file, and updated an executable that runs the raw analysis, that is under tutorials for some reason. On my side I am happy to merge, but you can have another look to see if everything makes sense.

andreufont:
I'd move the executable to /bin/ and merge.

iprafols (approved):

iprafols:
It seems we are ready to merge, right?

andreicuceu:
Yep

PR #800, opened 2021-09-28
Add simple strategy E implementation.

andreicuceu:
This is adds a quick version of strategy E (corr_yshift) where we use an external file (saved on NERSC) to store the difference between a blind model and an unblind model. This difference gets added to the measured correlation at the picca_export step. Still needs testing, and the real file to be produced in a common DESI directory at NERSC.

### DIFF
--- a/bin/picca_export.py
+++ b/bin/picca_export.py
@@ -253,10 +253,15 @@ def main(cmdargs):
 
     # Check if we need blinding and apply it
     if 'BLIND' in data_name or blinding != 'none':
-        if blinding != 'corr_yshift':
+        if blinding == 'corr_yshift':
+            userprint("Blinding using strategy corr_yshift.")
+        elif blinding == 'minimal':
             blinding = 'corr_yshift'
-            print("Only strategy E called 'corr_yshift' is allowed for now."
-                  " This will be applied automatilly.")
+            userprint("Only strategy E called 'corr_yshift' is allowed for now."
+                      " This will be applied automatilly.")
+        else:
+            raise ValueError("Expected blinding to be 'corr_yshift' or 'minimal'."
+                             " Found {}.".format(blinding))
 
         if args.blind_corr_type is None:
             raise argparse.ArgumentError("Blinding strategy 'corr_yshift' requires"
@@ -264,7 +269,7 @@ def main(cmdargs):
 
         # Read the blinding file and get the right template
         blinding_filename = ('/global/cfs/projectdirs/desi/users/acuceu/notebooks/'
-                             'vega_models/blinding/blinding_file/test_2.h5')
+                             'vega_models/blinding/blinding_file/blinding_model_2.h5')
         if not os.path.isfile(blinding_filename):
             raise RuntimeError("Missing blinding file. Make sure you are running at"
                                " NERSC or contact picca developers")
--- a/bin/picca_export.py
+++ b/bin/picca_export.py
@@ -25,11 +25,12 @@ def main(cmdargs):
         required=True,
         help='Correlation produced via picca_cf.py, picca_xcf.py, ...')
 
-    parser.add_argument('--out',
-                        type=str,
-                        default=None,
-                        required=True,
-                        help='Output file name')
+    parser.add_argument(
+        '--out',
+        type=str,
+        default=None,
+        required=True,
+        help='Output file name')
 
     parser.add_argument(
         '--dmat',
@@ -62,10 +63,17 @@ def main(cmdargs):
         required=False,
         help='Remove a correlation from shuffling the distribution of los')
 
-    parser.add_argument('--do-not-smooth-cov',
-                        action='store_true',
-                        default=False,
-                        help='Do not smooth the covariance matrix')
+    parser.add_argument(
+        '--do-not-smooth-cov',
+        action='store_true',
+        default=False,
+        help='Do not smooth the covariance matrix')
+
+    parser.add_argument(
+        '--blind_corr_type',
+        default=None,
+        choices=['lyaxlya', 'lyaxlyb', 'qsoxlya', 'qsoxlyb'],
+        help='Type of correlation. Required to apply blinding in DESI')
 
     args = parser.parse_args(cmdargs)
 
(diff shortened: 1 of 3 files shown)


bin/picca_export.py
```diff
@@ -240,6 +241,44 @@ def main(cmdargs):
         'R-parallel', 'R-transverse', 'Redshift', 'Correlation',
         'Covariance matrix', 'Distortion matrix', 'Number of pairs'
     ]
+
+    # Check if we need blinding and apply it
+    if 'BLIND' in data_name or blinding != 'none':
```
iprafols, line 246: Why not put `if blinding == "corr_yshift"`? Otherwise add a TODO saying to change this once we are ready to unblind

bin/picca_export.py
```diff
@@ -240,6 +241,44 @@ def main(cmdargs):
...
+        if blinding != 'corr_yshift':
+            print('Only strategy E called "corr_yshift" is allowed for now.'
+                  ' This will be applied automatilly.')
+
+        # Figure out what which correlation it is
+        if len(xi) == 50:
```
iprafols, line 252: I would change the order of these ifs. Otherwise, there could be some binning raising an AssertionError instead of a ValueError with the correct message. I would do something like
```
if 'qso' in qrgs.data.lower():
    if len(xi) != 100:
        raise ValueError('Unknown binning. Cannot blind. Please raise an issue'
                                    ' or contact picca developers.')
    if 'lyb' in args.data.lower():
        blind_corr = 'qsoxlyb'
    else:
        blind_corr = 'qsoxlya'
else:
    if len(xi) != 50:
        raise ValueError('Unknown binning. Cannot blind. Please raise an issue'
                                    ' or contact picca developers.'))
    if 'lyb' in args.data.lower():
        blind_corr = 'lyaxlyb'
    else:
        blind_corr = 'lyaxlya'
```

bin/picca_export.py
```diff
@@ -240,6 +241,44 @@ def main(cmdargs):
...
+            if 'lyb' in args.data or 'Lyb' in args.data:
+                blind_corr = 'qsoxlyb'
+            else:
+                blind_corr = 'qsoxlya'
+        else:
+            raise ValueError('Unknown binning. Cannot blind. Please raise an issue'
+                             ' or contact picca developers.')
+
+        # Read the blinding file and get the right template
+        blinding_path = ('/global/cfs/projectdirs/desi/users/acuceu/notebooks/'
```
iprafols, line 268: Consider renaming it to `blinding_filename`. Add a check that makes sure the file exists and complain otherwise
```
if not os.path.isfile(blinding_path):
    raise ValueError("Missing blinding file. Make sure you are running at"
                                " NERSC or contact picca developers")
```

iprafols (changes requested):

bin/picca_export.py
iprafols: Resolved by commit #728c397

bin/picca_export.py
andreicuceu: Yeah this has to do with erring on the safe side (blind as much as possible). When unblinding we would would run picca_deltas without the blinding. So the change will be in picca_deltas not here, and this code doesn't get used at that point. But let's talk about this tomorrow.

bin/picca_export.py
iprafols: Yes that could work, but also add it for the qsoxlyb

iprafols: What I am worried is that at some point we might want to double the binning in r_par o r_trans thus having 100 pixels and then entering the wrong branch

bin/picca_export.py
iprafols: We could also add an 
```
# TODO: remove this when we are ready to unblind
else:
    raise ValueError("Expected blinding strategy corr_yshift")
```
but let's discuss it tomorrow morning

andreufont: I think it is ok as it is now, since it allows you to run picca_export.py on correlations function files that were computed using the minimal strategy, and it blinds them with the new strategy corr_yshift

bin/picca_export.py
andreufont: As I mentioned on Slack, we could add --blind-corr-type as an option to picca_export.py that would specify the correlation type. It would only be required when running on blind data.

bin/picca_export.py
```diff
@@ -240,6 +242,48 @@ def main(cmdargs):
         'R-parallel', 'R-transverse', 'Redshift', 'Correlation',
         'Covariance matrix', 'Distortion matrix', 'Number of pairs'
     ]
+
+    # Check if we need blinding and apply it
+    if 'BLIND' in data_name or blinding != 'none':
+        if blinding != 'corr_yshift':
+            print('Only strategy E called "corr_yshift" is allowed for now.'
+                  ' This will be applied automatilly.')
```
andreufont, line 250: As I say above I think this is ok, but we should consider resetting at this point the variable "blinding":
blinding = 'corr_yshift'

iprafols (approved):

andreufont:
If someone runs the script with --blind-corr-type lyaxlya on a file that was not blind, should we crash? It might help us find some problematic runs...

iprafols:
We can have it crush, but I'm nto sure which type of problematic run this would help detect

alxogm:
If I understood correctly, when picca_export.py is run in correlations computed under the minimal blinding, it will  apply the new blinding  right?  in this case shouldn't DA_BLIND and DM_BLIND be just DA and DM again ?

andreufont:
@alxogm - I think we want to keep the "BLIND" label in all relevant column names, as we do whenever we deal with main-survey data that should be blind. 

Your comment, however, made me realisat that we should make changes in fitter2 and Vega so that they can run with free BAO parameters. We can open issues in Picca / Vega and fix these later.

iprafols:
I would keed the names as well. Otherwise in the future we might have trouble distinguishing between blind/unblind correlation files

alxogm:
>Your comment, however, made me realise that we should make changes in fitter2 and Vega so that they can run with free BAO parameters. We can open issues in Picca / Vega and fix these later.

Exactly, that and plotting scripts is why I ask... 

User's plotting scripts, like the one used [here ](https://desi.lbl.gov/trac/wiki/LymanAlphaWG/how_to_run_picca#no1 )will not distinguish between blindings, and prevent to plot the wedges  or if we modify then will plot wedges even if only minimal blinding was set... Maybe not many actually produced files with the minimal blinding, and is not really important...

andreufont:
> > Your comment, however, made me realise that we should make changes in fitter2 and Vega so that they can run with free BAO parameters. We can open issues in Picca / Vega and fix these later.
> 
> Exactly, that and plotting scripts is why I ask...
> 
> User's plotting scripts, like the one used [here ](https://desi.lbl.gov/trac/wiki/LymanAlphaWG/how_to_run_picca#no1)will not distinguish between blindings, and prevent to plot the wedges or if we modify then will plot wedges even if only minimal blinding was set... Maybe not many actually produced files with the minimal blinding, and is not really important...

I think it is useful that the plotting scripts know that they have to be worried about blind data, so I would keep DA_BLIND as the name and force everyone to think about how their plottings scripts should deal with data from the different strategies

alxogm:
> 
> I think it is useful that the plotting scripts know that they have to be worried about blind data, so I would keep DA_BLIND as the name and force everyone to think about how their plottings scripts should deal with data from the different strategies

Sounds good to me.

bin/picca_export.py
```diff
@@ -240,6 +250,41 @@ def main(cmdargs):
...
+        else:
+            raise ValueError("Expected blinding to be 'corr_yshift' or 'minimal'."
+                             " Found {}.".format(blinding))
+
+        if args.blind_corr_type is None:
+            raise argparse.ArgumentError("Blinding strategy 'corr_yshift' requires"
+                                         " argument --blind_corr_type.")
+
+        # Read the blinding file and get the right template
+        blinding_filename = ('/global/cfs/projectdirs/desi/users/acuceu/notebooks/'
```
iprafols, line 271: This needs to be changed to the actual folder

iprafols (changes requested):

iprafols (approved):

andreufont (approved):

PR #806, opened 2021-10-01
adapted fitter2 to the corr_yshift blinding

andreufont:
It seems to be working, but it would be great if someone else could give it a shot. 

You can take a look at the scripts that I have put together in 
https://github.com/cosmodesi/desi_lya_sv/tree/master/everest_healpix_e2e/scripts

### DIFF
--- a/py/picca/fitter2/data.py
+++ b/py/picca/fitter2/data.py
@@ -33,20 +33,18 @@ class data:
         h = fitsio.FITS(fdata)
         head = h[1].read_header()
 
-        blinding = "none"
+        self._blinding = "none"
         if 'BLINDING' in head:
-            blinding = head["BLINDING"]
+            self._blinding = head["BLINDING"]
 
-        if blinding == "minimal":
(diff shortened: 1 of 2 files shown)


py/picca/fitter2/data.py
```diff
@@ -287,7 +286,7 @@ def __init__(self,dic_init):
                     self.z_met[(self.tracer1['name'], m)] = hmet[2]["Z_{}_{}".format(self.tracer1['name'],m)][:]
                     
                     metal_mat_name = "DM_{}_{}".format(self.tracer1['name'], m)
-                    if self._blind:
+                    if not self._blinding == 'none':
```
iprafols, line 289: change to `if self._blinding != 'none'`

iprafols (changes requested):

py/picca/fitter2/data.py
andreufont: I have changed this as requested

iprafols (approved):

PR #809, opened 2021-10-04
Revert "Read desi healpix coadd"

iprafols:
Reverts igmhub/picca#798. Will remerge once more tests are done


andreufont:
Hi @iprafols - I was able to run with the new master branch without an issue, I don't think we need to revert this PR.

Issue #807, opened 2021-10-04
Update continuum fitting and fitter2 to support newer versions of iminuit


andreufont:
We are currently forced to use a fairly old version of iminuit (1.49), but there are much newer versions available.

iprafols:
After PR #825 and the everest tag we moved to iminuit 2

PR #817, opened 2021-10-21
Draft (Picca iminuit)

SamYoules:
Test: trying to merge master changes into this branch


Waelthus:
Ok, this is now at a point where things run through and produce output with similar format for fitter2. There are differences in the results that we'll need to have a look at

Waelthus:
Ok, now fitter2 actually works and passes the test, still not sure why picca_deltas does not.
Utils have updated code for iminuit2, but not tested.
Fitter1 and picca_* relying on it are removed with this PR

Waelthus:
Ok, found an issue in prep_del, where limits for the `master` were different from the limits in this branch, but changing that did not fix the problems, i.e. if I didn't overlook something, the new code should give the same results as before, but does not. Plots for delta_attributes are attached. There seems to be an issue for loglam>~3.7 and loglam_rest>~3.06 (i.e. at the high-z end where less data is available in the test) where the choice of iminuit version changes results. There's (nearly) no difference in the mean continuum itself, the stack or the number of pixels used.

[CONTmean_cont_compare_iminuit.pdf](https://github.com/igmhub/picca/files/7497226/CONTmean_cont_compare_iminuit.pdf)
[CONTweight_compare_iminuit.pdf](https://github.com/igmhub/picca/files/7497227/CONTweight_compare_iminuit.pdf)
[STACKstack_compare_iminuit.pdf](https://github.com/igmhub/picca/files/7497228/STACKstack_compare_iminuit.pdf)
[STACKweight_compare_iminuit.pdf](https://github.com/igmhub/picca/files/7497229/STACKweight_compare_iminuit.pdf)
[WEIGHTnb_pixels_compare_iminuit.pdf](https://github.com/igmhub/picca/files/7497232/WEIGHTnb_pixels_compare_iminuit.pdf)
[WEIGHTvar_lss_compare_iminuit.pdf](https://github.com/igmhub/picca/files/7497233/WEIGHTvar_lss_compare_iminuit.pdf)
[WEIGHTeta_compare_iminuit.pdf](https://github.com/igmhub/picca/files/7497230/WEIGHTeta_compare_iminuit.pdf)
[WEIGHTfudge_compare_iminuit.pdf](https://github.com/igmhub/picca/files/7497231/WEIGHTfudge_compare_iminuit.pdf)

py/picca/fitter2/chi2.py
```diff
@@ -274,16 +285,22 @@ def fastMC(self):
 
     def minos(self):
         if not hasattr(self,"minos_para"): return
-
         sigma = self.minos_para['sigma']
+        cl=scipy.stats.norm.cdf(sigma, loc=0, scale=1)-scipy.stats.norm.cdf(-sigma, loc=0, scale=1)
```
iprafols, line 289: Could the difference in the results be originated here? Or is this not entering the tests?

py/picca/fitter2/chi2.py
```diff
@@ -294,15 +311,19 @@ def export(self):
         g=f.create_group("best fit")
 
         ## write down all parameters
-        for i, p in enumerate(self.best_fit.values):
+        for i, p in enumerate(self.best_fit.values.to_dict().keys()):
             v = self.best_fit.values[p]
             e = self.best_fit.errors[p]
-            if p in self.best_fit.list_of_fixed_param():
+            if p in [name for (name,fix) in self.best_fit.fixed.to_dict().items() if fix]:
```
iprafols, line 317: The list comprehension `[name for (name,fix) in self.best_fit.fixed.to_dict().items() if fix]` appears multiple times in the following lines. Maybe it's good to assign it to a variable before the `for` in lone 314

py/picca/fitter2/chi2.py
```diff
@@ -379,10 +406,11 @@ def export(self):
         if hasattr(self, "minos_para"):
             g = f.create_group("minos")
             g.attrs['sigma'] = self.minos_para['sigma']
-            minos_results = self.best_fit.get_merrors()
+            minos_results = self.best_fit.merrors # self.best_fit.get_merrors()
```
iprafols, line 409: probably we can remove the comment, right? I don't think it is giving us any relevant information

iprafols (changes requested):

py/picca/fitter2/chi2.py
Waelthus: nope, this is part of fitter2 which by now is giving the correct result. The change is because nomenclature in minos calls went from N-sigma to giving minos the confidence range you want, the `cl=...` here is just for doing the conversion.

It might be that we get different results for an N-d minos fit with N>1, this is not part of the test case and I'm not certain if the sigma used for previous versions was supposed to mean 1σ-confidence range of a 1-d Gaussian (i.e. 68%) or 1σ-confidence range of a N-d Gaussian (which would be <68%). In the implementation so far I assume the former, but it might be the latter...

py/picca/fitter2/chi2.py
Waelthus: yes, we could. This list-comprehension is what the deprecation warning of list_of_fixed_param() was suggesting to use, so I just used this for the moment (it also ensures we're not accidentally using the list of another fit than the most recent one).

Waelthus: ok, I did change to define a variable in the minos and export functions for `fixed_pars` and `varied_pars`, mostly for more verbose code.

py/picca/fitter2/chi2.py
Waelthus: did just that

py/picca/fitter2/chi2.py
iprafols: ok so then I guess the comments might be a bit misleading?

iprafols (approved):

iprafols:
The changes look fine, we just need to figure out why the tests are failing

Waelthus:
I'm pretty puzzled by the failing of the tests. 
For the fitter things seem to be giving the same answer (for every single quantity in the hdf5 except the one number [I think it is the number of calls in the last fitting step]) that cannot be retrieved anymore.
For deltas with constant weights (i.e. delta P1d tests) everything works perfectly fine, but for deltas with standard weights there is a difference wrt previous tests in some of the properties, but using exactly the same number of pixels (i.e. I think it's not to having spectra that are close to the cuts taken and randomly are included or not depending on numerical effects or sth)

Waelthus:
Ok, I found the issue. The difference was due to the following (in prep_del.compute_var_stats):
The `master` version checked for `migrad_ok` during the computation which is deprecated and meant that the function minimum is valid.
This branch instead checked for `accurate`, which means that the covariance matrix estimate is accurate and not just an approximation (i.e. different condition than before).

In case this check returns False both versions return hard coded default values for eta/var_lss/fudge=1/0.1/1e-7 for the offending wavelength bin. In the `master` tests this always happens for the last bin, while here it happened for the last 2 bins due to the stronger condition 

> loglam    eta      var_lss  fudge    chi2     num_pix
> 3.727e+00 1.00e+00 1.00e-01 1.00e-07 1.62e-09 2845
> 3.736e+00 1.00e+00 1.00e-01 1.00e-07 0.00e+00 1186

I just changed the check for `accurate` to a check for `valid` which has the same meaning as the previous `migrad_ok` and tests now pass.

Waelthus:
@iprafols: I guess this can be merged whenever Vega has been updated accordingly.

Waelthus:
closes #807

iprafols:
Perfect! good job fixing this! @andreicuceu  can you let us know when you can port this to Vega as well?

SamYoules:
Thanks for taking it over Michael!

Best wishes,
Sam

Issue #815, opened 2021-10-20
Make sure that DLA/BAL Catalogs are supported and work as expected


Waelthus:
For the BAL side there has been work by @LaurEnnesser.
For the DLA side ideally the DLA finder group should try to implement a similar approach to the cats.

We do want to be able to run the full analysis with the everest tag, so this needs to be done before.

zoujiaqi99:
Hi, which column names of the current DLA catalog do I need to change?  What do you mean about 'a similar approach'?

Waelthus:
For the BAL catalog the current idea is to add BAL information to the QSO catalog as additional columns to the main ZCATALOG extension. If that's feasible for DLAs, one could do the same for those to unify the treatment of "contaminants".

The other part is to ensure that whichever catalog format we will use for DLAs, picca needs to be able to read it. I'm not sure this is currently the case, but I haven't looked at your catalogs yet. Are those available on NERSC and accessible for everybody?

@alxogm: Have you (or someone else) looked into the DLA cats from @zoujiaqi99 already?

andreufont:
I agree with @iprafols , we should probably use a different catalog for DLAs than for BALs. 

Thank you @zoujiaqi99 for providing the catalog! Regarding column names in your catalog, this is the current list:              
```
TTYPE1  = 'TARGET_RA'                                                           
TTYPE2  = 'TARGET_DEC'                                                          
TTYPE3  = 'ZQSO    '                                                            
TTYPE4  = 'Z       '                                                            
TTYPE5  = 'TARGETID'                                                            
TTYPE6  = 'S/N     '                                                            
TTYPE7  = 'DLAID   '                                                            
TTYPE8  = 'NHI     '                                                            
TTYPE9  = 'DLA_CONFIDENCE'                                                      
TTYPE10 = 'NHI_STD '                                                            
TTYPE11 = 'ABSORBER_TYPE'                                                       
TTYPE12 = 'QSO without BAL'                                                     
TTYPE13 = 'DLA     '                                                            
TTYPE14 = 'Assoc_DLA'                                                           
TTYPE15 = 'Interesting_flag'                                                    
TTYPE16 = 'Comments'     
```                                                       

Some of these are common with other DESI files (TARGET_RA, TARGET_DEC, TARGETID). These are great. 

The function in Picca that reads DLA positions (read_dlas in io.py) currently expects to find columns names 'Z' and 'NHI', so I think we can leave these as they are for now. 

[...]

PR #822, opened 2021-11-04
Bal tools update

LaurEnnesser:
BAL masking has been updated for DESI.


Waelthus:
closes #815

Waelthus:
please tell when this is ready for reviewing...

LaurEnnesser:
Yes, sorry, this is ready for review

py/picca/bal_tools.py
```diff
@@ -12,50 +12,59 @@
...
     Returns:
-        A dictionary with BAL information. Keys are the THING_ID
+        A dictionary with BAL information. Keys are the TARGETID
         associated with the BALs. Values are a tuple with its AI
         (*_CIV_450) and BI (*_CIV_2000) velocity.
 
     """
+    if survey == 'EBOSS':
+        id_name = 'THING_ID'
+        ext_name = 1
```
Waelthus, line 32: given that the previous code used BALCAT here, maybe change to `extname='BALCAT'`. Or is this to allow both a BAL catalog and an eBOSS style drq with BALs included? (then maybe add a comment about this)

bin/picca_deltas.py
```diff
@@ -613,18 +613,27 @@ def main(cmdargs):
 
     ### Mask BALs
     if not args.bal_catalog is None:
-        userprint("INFO: Adding BALs found in BAL catalog")
-        bal_cat = bal_tools.read_bal(args.bal_catalog)
+        userprint("INFO: Masking BALs from BAL catalog")
+        bal_cat = bal_tools.read_bal(args.bal_catalog,'EBOSS')
```
Waelthus, line 617: is this right? wouldn't this only allow using EBOSS cats? Or is reading DESI cats just unneccessary because DESI BALs are always fed into the QSO cat?

bin/picca_deltas.py
```diff
@@ -613,18 +613,27 @@ def main(cmdargs):
...
-                    mask_obs_frame_bal = []
-                    mask_rest_frame_bal = bal_tools.add_bal_rest_frame(
-                        bal_cat, forest.thingid, args.bal_index)
-                    forest.mask(mask_rest_frame_bal)
+                    bal_mask = bal_tools.add_bal_mask(
+                        bal_cat, forest.thingid, args.bal_index, 'THING_ID')
+                    forest.mask(bal_mask)
                     num_bal += 1
         log_file.write("Found {} BAL quasars in forests\n".format(num_bal))
+    elif (args.keep_bal is True) & (args.bal_catalog is None):
```
Waelthus, line 627: as this is already an `elif` on the previous `if not args.bal_catalog is None`, the second half is superfluous.

bin/picca_deltas.py
```diff
@@ -915,7 +924,7 @@ def main(cmdargs):
                                               delta.fiberid)
                     },
                     {
-                        'name': 'THING_ID',
+                       'name': 'THING_ID',
```
Waelthus, line 927: shouldn't this raise an indentation error? I think this space should not be removed

Waelthus:
Did you also check that io.read_drq (https://github.com/igmhub/picca/blob/de7c3edcb7a7069fcba9c94693a382d5b5144b72/py/picca/io.py#L124) is doing what it should regarding BALs? Also if keep_bal is not set?

bin/picca_deltas.py
iprafols: I'd run pylint to check what is the correct spacing

bin/picca_deltas.py
iprafols: Reading the bal_tools.read_bal the survey is set to 'DESI' by default, but with this it is always called usin  as 'EBOSS'. To fix this, I would add a parsing option (with 'EBOSS' as default if you want) and then pass `args.survey` to this function

LaurEnnesser: Is args.survey already an argument for picca_deltas? I don't see it available.

The EBOSS quasar catalogs do not contain the BAL information needed for masking, so separate BAL catalogs were made. As a result, in EBOSS, I need a way to match the quasars in both catalogs. I have been using 'THINGID' for simplicity, but this results in a naming clash with DESI, where we start using 'TARGETID'. I'm currently not aware of any easy way to switch between the naming conventions of the two surveys, but if there is I'd be happy to incorporate it.

 For DESI, we're planning on having BAL information added to the quasar catalogs as an afterburner, so no additional BAL catalog is needed.

iprafols: No, it is not an argument yet. My suggestion is to create one so that we can choose the survey we want to use (currently we are always running in eBOSS mode).

I'm not sure what you mean in the second and third paragraphs.

bin/picca_deltas.py
```diff
@@ -617,48 +613,23 @@ def main(cmdargs):
...
         userprint("INFO: Masking BALs")
         bal_cat = bal_tools.read_bal(bal_catalog_to_read,args.survey)
         num_bal = 0
         for healpix in data:
             for forest in data[healpix]:
-                bal_mask = bal_tools.add_bal_mask(bal_cat, forest.thingid)
-                if len(bal_mask) > 0:
+                bal_mask = bal_tools.add_bal_mask(bal_cat, forest.thingid,
+                        args.survey)
+            forest.mask(bal_mask)
```
iprafols, line 628: This should be inside the loop, right?

py/picca/bal_tools.py
```diff
@@ -46,17 +44,22 @@ def read_bal(filename,survey):  ##Based on read_dla from picca/py/picca/io.py
...
         objectid: str
             Identifier of quasar
-        id_name: str
-            Column name of quasar identification ("THING_ID" in EBOSS)
+        survey: str
+            Survey catalog comes from. 
     """
+    if survey == 'eboss':
+        id_name = 'THING_ID'
+    else:
```
iprafols, line 60: the `else` should be `elif survey == 'desi'` and there should be an extra `else` with an error

py/picca/bal_tools.py
```diff
@@ -102,6 +105,9 @@ def add_bal_mask(bal_catalog, objectid, id_name='TARGETID'):
         for line in lines.values():
             min_wavelength = np.log10(line * (1 - min_velocities[vel] / light_speed))
             max_wavelength = np.log10(line * (1 - max_velocities[vel] / light_speed))
-            bal_mask.add_row([max_wavelength,min_wavelength,'RF'])
+            if survey == 'eboss':
```
iprafols, line 108: This seems weird. Why would the mask for DESI be ordered differently than the mask from eBOSS? The function forest.mask is the same irrespective of the survey being analyzed. Can you double-check that this is correct and necessary? I would go with simply `bal_mask.add_row([min_wavelength,max_wavelength,'RF'])`

iprafols (changes requested):

py/picca/bal_tools.py
iprafols: addressed in 06a81ce

bin/picca_deltas.py
iprafols: addressed in 53937dd

iprafols (approved):

PR #824, opened 2021-11-11
add desi_healpix as mode for picca_metal_xdmat

andreufont:
Small fix to run metal matrix for the cross-correlation in Everest healpix data


alxogm (approved):
Thanks for catching it, I thought all scripts were updated in a past PR but this one was missed.

PR #829, opened 2021-12-07
Look for LOS_ID in delta files

andreufont:
The current master branch looks for THING_ID in the delta files, but the new continuum branch (delta_extraction) writes instead LOS_ID (because DESI and other surveys do not have THING_ID).

In the long run we should rewrite the rest of Picca to be less SDSS-scentric (and scrap PLATE, MJD, FIBER), but meanwhile this pull request provides a quick fix so that we can measure correlations from the new delta files. 

I also used this PR to fix a couple of minor bugs that I encountered.

### DIFF
--- a/py/picca/data.py
+++ b/py/picca/data.py
@@ -1201,7 +1201,12 @@ class Delta(QSO):
         fitiing. See equations 5 and 6 of du Mas des Bourboux et al. 2020
         """
         # 2nd term in equation 6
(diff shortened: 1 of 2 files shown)


Waelthus (approved):
looks good to me. If one wanted to depending on survey also PLATE/MJD/FIBERID or NIGHT/PETAL/TILE would be available, but reading those would only complicate I/O for this step

PR #832, opened 2021-12-08
Delta extraction true continuum

alxogm:
Not ready to merge, but I open de PR to discuss issues/questions.

I'm testing with this configuration

```
[general]
overwrite = True
quiet = False
out dir = /global/cscratch1/sd/alxogm/mocks/picca_dev/true_cont_stack/deltas/

[data]
type = DesiData
input directory = /global/cfs/cdirs/desi/mocks/lya_forest/develop/london/qq_desi/v9.0/v9.0.0/desi-2.0-4/spectra-16/
catalogue = /global/cscratch1/sd/alxogm/mocks/picca_dev/true_cont/test_cat.fits

#/global/cfs/cdirs/desi/mocks/lya_forest/develop/london/qq_desi/v9.0/v9.0.0/desi-2.0-4/zcat.fits
rebin = 3
lambda min = 3600.0
lambda max = 7235.0
lambda rest min = 1040.0
lambda rest max = 1200.0
wave solution = log
mode = healpix
nspec=100

[corrections]
num corrections = 0
optical-depth = 0.0023 3.64 LYA

[masks]
num masks = 0

[expected flux]
type = TrueContinuum
input directory = /global/cfs/cdirs/desi/mocks/lya_forest/develop/london/qq_desi/v9.0/v9.0.0/desi-2.0-4/spectra-16/
```

TO DO:

Update mean flux used
Update var_lss used, and see If there are iterations needed. Not sure of the stack_deltas part...
Update what is written to delta_attributes file, if any.
and more ...


py/picca/delta_extraction/data_catalogues/desi_data.py
```diff
@@ -223,35 +223,45 @@ def read_from_healpix(self, catalogue):
...
              (healpix, survey)), group in zip(enumerate(grouped_catalogue.groups.keys),
                                     grouped_catalogue.groups):
-
-            input_directory = f'{self.input_directory}/{survey}/dark'
-            filename = (
+            input_directory = f'{self.input_directory}/'
+            filename_truth = (
+                f"{input_directory}/{healpix//100}/{healpix}/truth-{in_nside}-"
+                f"{healpix}.fits")
+            if not os.path.isfile(filename_truth):
```
alxogm, line 250: hmm I just noticed I have to change in_nside to 64 here for the data. But I wanted to check with @iprafols he agrees with this in order to read the data with the same DesiData type, or maybe he prefers to write a separate method even though the way it is working now it is this function the one indicating whether is mock or data...

py/picca/delta_extraction/expected_fluxes/true_continuum.py
```diff
@@ -0,0 +1,392 @@
...
+                self.logger.info(f"Rejected forest with los_id {forest.los_id} "
+                                 f"due to {forest.bad_continuum_reason}")
+                continue
+            # get the variance functions and statistics
+            log_lambda = forest.log_lambda
+            stack_delta = self.get_stack_delta(log_lambda)
+            var_lss = self.get_var_lss(log_lambda)
+
+
+            mean_expected_flux = forest.continuum #* stack_delta ##not sure of this stack_delta
```
alxogm, line 372: I'm not sure we need the stack_delta here...

py/picca/delta_extraction/expected_fluxes/true_continuum.py
```diff
@@ -0,0 +1,392 @@
...
+        mean_optical_depth = np.ones(forest.log_lambda.size)
+        tau, gamma, lambda_rest_frame = 0.0023, 3.64, 1215.67
+        w = 10.**forest.log_lambda / (1. + forest.z) <= lambda_rest_frame
+        z = 10.**forest.log_lambda / lambda_rest_frame - 1.
+        mean_optical_depth[w] *= np.exp(-tau * (1. + z[w])**gamma)
+
+        forest.continuum *= mean_optical_depth
+        return forest
+
+    def read_var_lss(self):
```
alxogm, line 204: @andreicuceu  I did this only to have something working, the var_lss in this case was from a run of picca_deltas in a mock, that I had from other tests... so please do change this function as we discussed...

py/picca/delta_extraction/data_catalogues/desi_data.py
iprafols: Hi @alma, I added a fix to the desi_extraction branch, I'm not sure I merged but it not I'll do it right now

py/picca/delta_extraction/expected_fluxes/true_continuum.py
iprafols: You are multiplying the true continuum by the mean expected flux, so I'd say you don't

py/picca/delta_extraction/data_catalogues/desi_data.py
iprafols: ok, so weirdly enough this was not set to 64 even when the merge was done. I will push a commit for this

andreicuceu:
Hi @alxogm and @iprafols, I added the var_lss from the raw analysis. There are three new files in the expected_flux folder, so I made a new subfolder for them (more will probably be added in the future). For now, the three options are log binning (with the standard bin size), linear binning with 2.4A bins, and linear binning with 3.2A bins. I can make more as needed. These files are automatically written when running the raw analysis, for a new binning we just need to run the raw analysis with that binning first, and then add the stats file to picca.

I haven't tested this code yet, so if anyone wants to give it a try that would be great.

andreicuceu:
sigma_lss_comp

alxogm:
I have tested the different var_lss files can be used, but I'm still looking at the results. @iprafols I have merged the delta_extraction branch into this one, but it seems the blinding is not being updated to `none`, it is using the default `corr_yshift`

iprafols:
Hi @alxogm , 

I have pulled the changes and still found some bug in the set_blinding function. Weirdly this was fixed for me in the delta_extraction branch, so I merged it again. Now the problem should be fixed (I think)

alxogm:
> Hi @alxogm ,
> 
> I have pulled the changes and still found some bug in the set_blinding function. Weirdly this was fixed for me in the delta_extraction branch, so I merged it again. Now the problem should be fixed (I think)

Oh, its possible I fixed it in the delta_extraction branch and merged to the other branch just locally... thanks for merging...

py/picca/delta_extraction/astronomical_objects/forest.py
```diff
@@ -523,6 +523,10 @@ def rebin(self):
             w1 = w1 & (self.lambda_ / (1. + self.z) < Forest.lambda_max_rest_frame)
             w1 = w1 & (self.ivar > 0.)
             if w1.sum() == 0:
+                self.lambda_ = np.array([])
```
alxogm, line 526: @iprafols The issue I was having in `compute_delta_stack` was due to a short forest not being rejected for `wave solution = lin`, but it was for `wave solution = log`, for which there was some negative flux bins... and the differences comes from those lines being present or not. Were those lines really missing?

iprafols: @alxogm, yes they were really missing. Good detective work!

py/picca/delta_extraction/data_catalogues/desi_healpix.py
```diff
@@ -186,16 +185,15 @@ def read_file(self, filename, catalogue, forests_by_targetid):
             targetid = row["TARGETID"]
             w_t = np.where(targetid_spec == targetid)[0]
             if len(w_t) == 0:
-                self.logger.warning(
-                    f"Error reading {targetid}. Ignoring object")
+                #self.logger.warning(
```
iprafols, line 188: why are you removing this comments?

py/picca/delta_extraction/data_catalogues/desi_healpix.py
```diff
@@ -208,7 +206,7 @@ def read_file(self, filename, catalogue, forests_by_targetid):
                     "targetid": targetid,
                     "ra": row['RA'],
                     "dec": row['DEC'],
-                    "z": row['Z'],
+                    "z": row['Z']
```
iprafols, line 209: This change is not doing anything

py/picca/delta_extraction/astronomical_objects/forest.py
iprafols: I'll add a commit to the delta extraction branch with this

py/picca/delta_extraction/data_catalogues/desi_healpix.py
alxogm: I thought I have restored it in the last commit, I had to comment it out when I was testing the parallelization, but should not be removed.

alxogm:
Comparison between picca_deltas and picca_delta_extraction.
Mock used: `/global/cfs/cdirs/desi/mocks/lya_forest/develop/london/qq_desi/v9.0/v9.0.0/desi-2.0-4/spectra-16/`

Used 200K skewers.
 Both picca_deltas and picca_delta_extraction ended with a remaining sample of 169436 forests.
 
 For the delta attributes comparison, I did not mask for `NB_PIXELS>0` because this is not being saved in `picca_delta_extraction` , but I don't think this is affecting the correlations, is it? 

Screen Shot 2022-02-02 at 20 10 37

The difference at the correlation level is quite large at some "random" points 

Screen Shot 2022-02-02 at 20 29 15

iprafols:
Hi @alxogm, you don't need to filter `NB_PIXELS>0` in the delta attributes file since we are now saving the interpolation functions (which are the ones used in the computation later on)

alxogm:
> sigma_lss_comp

Hi @andreicuceu can you point me to what mock you used to create this var_lss files? did you produced the raw deltas for that mock? Thanks?

alxogm:
Here I compare the delta attributes using the same plotting than Ignasi. The label says delta_extraction_true, referring to the branch, not the use of the true continuum. There are some funny differences at low and high observed waelengths, but ratios are below 1e-11. 
Screen Shot 2022-02-08 at 10 00 42

Regarding the use of the true continumm, I compare just a mean continuum and the stack of deltas computed using the true continuum. We have differences as expected, but I don't quite like the behavior of the delta stack  small and large observed wavelength.  

Screen Shot 2022-02-08 at 11 10 20
Screen Shot 2022-02-08 at 11 10 16

However, what I'm doing now is the comparison with the raw mocks, which I is more relevant. But I'm not sure if you want to merge this or leave it open until we have final comparison?

iprafols:
Hi @alxogm , 
Maybe we can discuss these plots on the next picca task force telecon. Hopefully by then you will have the comparison with the raw analysis. 
In any case we can discuss whether to proceeed with the merge or not then

alxogm:
@iprafols @andreufont  as discussed earlier I've changed from draft to actual PR.

py/picca/delta_extraction/data_catalogues/desi_healpix.py
```diff
@@ -118,7 +119,7 @@ def read_data(self):
 
         return False, is_sv
 
-    def read_file(self, filename, catalogue, forests_by_targetid):
+    def read_file(self, filename, catalogue,forests_by_targetid):
```
iprafols, line 122: why are you removing the space? The standard coding practices say to keep it

iprafols (approved):
We should add this class to the tutorials and to the data model

alxogm:
> We should add this class to the tutorials and to the data model

I can do it for the tutorial, but not sure how to do it for the data model...

iprafols:
I'll do it tomorrow morning, don't worry

andreufont (commented):
Thanks @alxogm for putting this together. Quick comment regarding the filenames with mean_flux and var_lss from mocks. Should we specify which mocks do these come from? Maybe something like: colore_v9_lin_2.4.fits.gz? There might be equivalent mocks from Saclay at some point...

If we wanted to go an extra mile, we could also store there some metadata related to how we computed these files. In particular I'd like to store the restframe wavelength used, since this has a large impact on the measured stats. I'm assuming these were run for (1040 < lr < 1200A?)? Do they ignore LyB region? The stats there should be slightly different, so I guess the safest / easiest thing is to restrict the analysis to LyA region for now? We might need input from @andreicuceu as well, and pinging @cramirezpe to keep him in the loop.

py/picca/delta_extraction/expected_fluxes/true_continuum.py
```diff
@@ -0,0 +1,486 @@
+"""This module defines the class Dr16ExpectedFlux"""
```
andreufont, line 1: @alxogm - we probably need to change this first documentation string

py/picca/delta_extraction/expected_fluxes/true_continuum.py
```diff
@@ -0,0 +1,486 @@
...
+                    "var lss binning"]
+
+defaults = {
+    "iter out prefix": "delta_attributes",
+}
+
+
+class TrueContinuum(ExpectedFlux):
+    """Class to the expected flux calculated used the true continuum unabsorbed contiuum
+    for mocks. It uses var_lss pre-computed from mocks and the mean flux estimated from
```
andreufont, line 28: This doc string seems broken, both in the beginning and ending of sentences.

py/picca/delta_extraction/expected_fluxes/true_continuum.py
```diff
@@ -0,0 +1,486 @@
...
+    compute_expected_flux
+    read_true_continuum
+    read_var_lss
+    populate_los_ids
+
+    Attributes
+    ----------
+    los_ids: dict (from ExpectedFlux)
+    A dictionary to store the mean expected flux fraction, the weights, and
+    the inverse variance for each line of sight. Keys are the identifier for the
```
andreufont, line 47: @iprafols - I guess "inverse variance" here refers to the noise variance from the pipeline directly (no eta, no var_lss, etc.), right? While "weights" include the final pixel variance, including all ingredients? I think it would be good to clarify this here and elsewhere, since we use "variance" with many different meanings in picca.

py/picca/delta_extraction/expected_fluxes/true_continuum.py
```diff
@@ -0,0 +1,486 @@
...
+    get_stack_delta: scipy.interpolate.interp1d or None
+    Interpolation function to compute the mean delta (from stacking all lines of
+    sight). None for no info.
+
+    get_var_lss: scipy.interpolate.interp1d
+    Interpolation function to compute mapping functions var_lss. See equation 4 of
+    du Mas des Bourboux et al. 2020 for details. Data for interpolation is read from a file.
+
+    input_directory: str
+    Directory where true continum data is store
```
andreufont, line 63: I would have thought that the directory with the true continuum is the same directory where the mock spectra are. Do we need another variable?

py/picca/delta_extraction/expected_fluxes/true_continuum.py
```diff
@@ -0,0 +1,486 @@
...
+        iter_out_file = self.iter_out_prefix + ".fits.gz"
+
+        with fitsio.FITS(out_dir + iter_out_file, 'rw',
+                         clobber=True) as results:
+            header = {}
+            header["FITORDER"] = -1
+            if Forest.wave_solution == "log":
+                num_bins = int((Forest.log_lambda_max - Forest.log_lambda_min) /
+                               Forest.delta_log_lambda) + 1
+                stack_log_lambda = (
```
andreufont, line 458: This variable "stack_log_lambda" is not used anywhere, is it?

py/picca/delta_extraction/expected_fluxes/true_continuum.py
```diff
@@ -0,0 +1,486 @@
...
+                    self.get_mean_cont(self.log_lambda_rest_frame),
+                    self.get_mean_cont_weight(self.log_lambda_rest_frame),
+                ],
+                              names=['loglam_rest', 'mean_cont', 'weight'],
+                              extname='CONT')
+
+            elif Forest.wave_solution == "lin":
+                num_bins = int((Forest.lambda_max - Forest.lambda_min) /
+                               Forest.delta_lambda) + 1
+                stack_lambda = (Forest.lambda_min +
```
andreufont, line 473: same here, I believe "stack_lambda" is not used.

py/picca/delta_extraction/expected_fluxes/true_continuum.py
```diff
@@ -198,7 +188,99 @@ def compute_expected_flux(self, forests, out_dir):
...
+        for forest in forests:
+            if forest.bad_continuum_reason is not None:
+                continue
+            bins = ((forest.log_lambda - Forest.log_lambda_min_rest_frame -
+                     np.log10(1 + forest.z)) /
+                    (Forest.log_lambda_max_rest_frame -
+                     Forest.log_lambda_min_rest_frame) * num_bins).astype(int)
+
+            var_lss = self.get_var_lss(forest.log_lambda)
+            var_pipe = 1. / forest.ivar/ forest.continuum**2
```
andreufont, line 263: is the "continuum" here the product of continuum times "mean transmitted flux fraction"? Otherwise it looks like the next equation is not right, one needs to take into account the "mean transmitted flux fraction" somewhere.

andreufont:
By the way @andreicuceu , @alxogm , I noticed that the current code hard-codes the value of <F>(z) 
```
            mean_optical_depth = np.ones(forest.log_lambda.size)
            tau, gamma, lambda_rest_frame = 0.0023, 3.64, 1215.67
```
We should be reading this from the raw skewers instead.

py/picca/delta_extraction/expected_fluxes/true_continuum.py
andreufont: Ignasi confirmed that this is "continuum * mean transmitted flux fraction". We should rename some of the variables in picca, but let's leave for a future issue.

py/picca/delta_extraction/expected_fluxes/true_continuum.py
andreufont: Ignasi has improved the docstring.

py/picca/delta_extraction/expected_fluxes/true_continuum.py
andreufont: Ignasi confirmed that indeed they are the same directory, that we should synchronise at some point (will open issue)

alxogm: I think it could be useful to keep it at least as optional to pass a different directory, I can imagine running some mocks and not used the true continuum flag, but having the same mock with different specifications but same true continuum so I can use it. But as you prefer.

alxogm:
> Thanks @alxogm for putting this together. Quick comment regarding the filenames with mean_flux and var_lss from mocks. Should we specify which mocks do these come from? Maybe something like: colore_v9_lin_2.4.fits.gz? There might be equivalent mocks from Saclay at some point...
> 
Agree. 

> If we wanted to go an extra mile, we could also store there some metadata related to how we computed these files. In particular I'd like to store the restframe wavelength used, since this has a large impact on the measured stats. I'm assuming these were run for (1040 < lr < 1200A?)? Do they ignore LyB region? The stats there should be slightly different, so I guess the safest / easiest thing is to restrict the analysis to LyA region for now? We might need input from @andreicuceu as well, and pinging @cramirezpe to keep him in the loop.

Not sure if the files already have some metadata, but I agree all this info should go into such files metadata. 

Yes, this was only Lya region... good point!

py/picca/delta_extraction/expected_fluxes/true_continuum.py
alxogm: Ah, is a leftover from when I was saving it, thanks for noticing. Will remove it.

py/picca/delta_extraction/expected_fluxes/true_continuum.py
alxogm: > By the way @andreicuceu , @alxogm , I noticed that the current code hard-codes the value of (z)
> 
> ```
>             mean_optical_depth = np.ones(forest.log_lambda.size)
>             tau, gamma, lambda_rest_frame = 0.0023, 3.64, 1215.67
> ```
> 
> We should be reading this from the raw skewers instead.

I've seen you open an issue so I'll mark here as resolved, and fix after the first merge.

alxogm:
> If we wanted to go an extra mile, we could also store there some metadata related to how we computed these files. 

I opened an issue to resolve this later on, I think is important we define well how to include the regions other than lya. Different files? or different hdu's in one file? I think different files is easier, since they will be produced that way.

cramirezpe:
It seems that files under `py/picca/delta_extraction/expected_fluxes/var_lss]` where not merged, and I cannot find them in the repo right now. Are they located somewhere at NERSC?

iprafols:
Probably yes. @alxogm do you know where? Can you add them in a new PR?

cramirezpe:
Sorry for the confusion, the files are there, but I need to change the setup file in order to find them when installing through pip. I will take care of it.

PR #833, opened 2021-12-09
commented out print_fmin from iminuit, it was crashing

andreufont:
iminuit2 seems to have deprecated this function

### DIFF
--- a/py/picca/fitter2/chi2.py
+++ b/py/picca/fitter2/chi2.py
@@ -124,7 +124,7 @@ class chi2:
         self.best_fit = self._minimize()
         if self.hesse:
             self.best_fit.hesse()
-            self.best_fit.print_fmin()
+            #self.best_fit.print_fmin()
 
         values = self.best_fit.values.to_dict()
         values['SB'] = False


andreufont:
@iprafols , @Waelthus - this should also be propagated to the Everest tag, and update it (not sure how to update a tag...)

Waelthus:
can you check if print(self.fmin) works? We can make this a bugfix tag 4.2.1 and I think there's a way to update the release to the new tag then...

iprafols (approved):

andreufont:
@Waelthus - working on it!

iprafols:
I'm happy to either comment this print or checking if Michael's sugestion works

andreufont:
@Waelthus - when I try your suggestion, I get the following error:
Minuit' object has no attribute 'print'

Let's merge the pull request as is, and open an issue to solve this properly at another time.

Waelthus:
> @Waelthus - when I try your suggestion, I get the following error: Minuit' object has no attribute 'print'
> 
> Let's merge the pull request as is, and open an issue to solve this properly at another time.

Fine that this was merged, the error seems weird to me, did you do this:
`print(self.best_fit.fmin)`
or this:
`self.best_fit.print(fmin)`
It looks like some iteration of the latter...

andreufont:
I'm not sure what I tried, but it definitely had self.best_fit.print on it.

Waelthus:
ok, then that's the reason it failed, what I meant was the former. So just using the normal print function on the self.best_fit.fmin. The print_xxx functions inside iminuit2 have been removed. As this was failing during analysis, but not in the regression tests, we should probably update those as well.

PR #834, opened 2021-12-09
bugfixes for linear binning

Waelthus:
This is needed for linear binning. Else floats will be read as strings... WIP

### DIFF
--- a/py/picca/data.py
+++ b/py/picca/data.py
@@ -1053,7 +1053,14 @@ class Delta(QSO):
             delta_name = "DELTA"
 
         delta = hdu[delta_name][:].astype(float)
-        log_lambda = hdu['LOGLAM'][:].astype(float)
+
+        if 'LOGLAM' in hdu.get_colnames():
+            log_lambda = hdu['LOGLAM'][:].astype(float)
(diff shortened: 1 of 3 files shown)


Waelthus:
seems to work now, will need to check results

Waelthus:
added a fix reading the new linearly binned deltas. Atm this is just done by converting LAMBDA to log10(λ) on read-in

py/picca/data.py
```diff
@@ -1053,7 +1053,14 @@ def from_fitsio(cls, hdu, pk1d_type=False):
             delta_name = "DELTA"
 
         delta = hdu[delta_name][:].astype(float)
-        log_lambda = hdu['LOGLAM'][:].astype(float)
+
```
iprafols, line 1056: I guess this is necessary to run picca_cf.py right?

iprafols (approved):

py/picca/data.py
Waelthus: yes, else picca won't be able to read anything for downstream analysis. What I do not know if inside e.g. picca_cf we implicitely assume equally sampled log_lambda. So this is just a placeholder change until we have modified routines to properly deal with linear binning

PR #837, opened 2021-12-20
Add testing for python 3.10

Waelthus:
Lets check if this works fine, then we can at least increase version coverage to newer versions already


Waelthus:
tihs fails as healpy is not yet python 3.10 compatible (or at least does not compile on it in the gh-action)

iprafols (commented):
Most of the changes are the same as PR #838 , just changing iminuit version and adding python 3.10

Waelthus:
yes, there were no other changes, but as said: some required packages (healpy at the very least) do not support python 3.10 yet, so testing with it fails

iprafols (approved):
I see that we are now requiring a specific version of iminuit. I think this is fine (as they tend to break things in newer versions), but we should keep an eye on new versions then.

PR #838, opened 2022-01-07
Redshift errors

SamYoules:
Changed converters.py to allow users to optionally add Gaussian random redshift errors to the drq catalog, to simulate errors in the classifiers.

### DIFF
--- a/py/picca/converters.py
+++ b/py/picca/converters.py
@@ -273,7 +273,8 @@ def desi_from_ztarget_to_drq(in_path,
                              out_path,
                              spec_type='QSO',
                              downsampling_z_cut=None,
-                             downsampling_num=None):
+                             downsampling_num=None,
+                             gauss_sigma_v=None):
     """Transforms a catalog of object in desi format to a catalog in DRQ format
 
     Args:
@@ -288,6 +289,9 @@ def desi_from_ztarget_to_drq(in_path,
         downsampling_num: int
             Target number of object above redshift downsampling-z-cut.
             'None' for no downsampling
+        gauss_sigma_v: int
+            Gaussian random error to be added to redshift (in kms)
+            'None' for no error
     """
 
     ## Info of the primary observation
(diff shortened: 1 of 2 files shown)


py/picca/converters.py
```diff
@@ -320,6 +324,13 @@ def desi_from_ztarget_to_drq(in_path,
     for key in ['RA', 'DEC']:
         cat[key] = cat[key].astype('float64')
 
+    # apply error to z
+    if gauss_sigma_v is not None:
+        c = 299792
```
andreufont, line 329: @SamYoules - we definitely have the speed of light defined somewhere in Picca, it would be better to not hardcode values like this in the code.

andreufont (changes requested):
I only added a small comments on the speed of light constant, but the rest looks good!

py/picca/converters.py
SamYoules: Thanks Andreu. This is done now.

andreufont:
@iprafols , @Waelthus - do you know why the tests above are failing? I belive the PR is ready to be merged otherwise.

Waelthus:
there are import errors for numpy stuff, unlikely to be related to the changes, I'll have a quick look

Waelthus:
the failing tests will be resolved with #839, they've been due to a package inconsistency between compile time of fitsio and runtime...

iprafols (approved):
things look fine, @andreufont , I guess you are happy with the changes? You should update your review so that we can merge. Michae'ls PR should be fixing the errors in the test so I say we are ready to merge (once this is verified)

Waelthus:
I merged the updated test, let's see how this works out in here

SamYoules:
I've changed the name of gauss_sigma_v to gauss_redshift_error as per Jim's request. It should be good to go now.

iprafols:
Great, so let's wait for @andreufont to confirm that the changes were applied and then we can merge

andreufont (approved):

PR #839, opened 2022-01-12
fix the testing by reinstalling fitsio after the correct numpy has been installed

Waelthus:
This PR is just doing changes to requirements by pinning the range of some package versions to debug the failing tests, we'll need to figure out a long-term solution after


Waelthus:
it looks like the issue was the following:
when pip installs packages via `pip install -r requirements.txt` it first downloads and compiles everything. 
On the test system the new numpy==1.22 seems to be available already before we pip install it, resulting fitsio to be compiled against numpy 1.22. This can be reproduced by creating a conda env, installing numpy via anaconda and then running `pip install -r requirements.txt`.
This is problematic behaviour for the picca testing as this new numpy version is not yet compatible to numba, i.e. a lower version will be downloaded when installing requirements.

I fix this for the tests by purging pips package cache and reinstalling fitsio explicitely after everything else, this is not ideal but should prevent the issue to appear.

I also added a way to get an ssh conneciton to the github action servers if something fails, this was actually useful here to figure out why things crashed (if the test jobs crash there should be a line like `ssh JTp4hNXRW33dM36TF2EdXDUbq@nyc1.tmate.io` in the log outputs, just run that and you'll connect to that github action server. The connection will stay open for up to 15mins.

iprafols (approved):
@Waelthus , Maybe it'd be good to add a todo comment specifying what we will need to remove in the future when the incompatibilities are lifted

Waelthus:
I added some comments to the workflow file and cleaned its structure a little:
- I'm not sure that we'd want to remove the additional "reinstall fitsio" step once versions are in sync again as they could fall out of sync anytime in the future. Properly fixing this would require changing how the installation is done i.e. probably installing packages in a certain order to prevent this, there might be semi-automatic ways to do it but I'm not enough of an expert, here.
- I did make the additional stuff the tests did on failures (saving outputs and starting a debug ssh session) optional. They will now only launch if [ci save] or [ci debug] are part of the commit msg (mostly in case that github at some point limits time/storage for these actions and to find the steps that failed more quickly)

PR #843, opened 2022-02-08
fixed bug in coadding forests

iprafols:
The function `coadd` in class `Forest` (py/picca/data.py) is set to return self or other when any of the two have arrays with 0 pixels, but the value of the forest is never updated in `pix_data` in  `io.py`. This implies that if the first forest has 0 pixels the information from the others is lost. 
This PR fixed this.


Waelthus:
Ha, we did have a similar issue earlier when doing the minisv i/o.
I added the equivalent fixes for other data formats.

iprafols:
Great! I indeed missed fixing in the other io functions

PR #845, opened 2022-02-14
Delta extraction

iprafols:
This contains the new way to compute the continuum fits and extract the delta field.


py/picca/data.py
```diff
@@ -512,6 +505,14 @@ def __init__(self,
         if reso is not None:
             reso = rebin_reso[w] / rebin_ivar[w]
 
+        # apply dust extinction correction
```
Waelthus, line 508: I guess this is just a reordering so corrections are more close together?

py/picca/data.py
```diff
@@ -1052,7 +1053,14 @@ def from_fitsio(cls, hdu, pk1d_type=False):
             delta_name = "DELTA"
 
         delta = hdu[delta_name][:].astype(float)
-        log_lambda = hdu['LOGLAM'][:].astype(float)
+
```
Waelthus, line 1056: ok, this is just dealing with the new output files and allows putting both binnings in. So all downstream routines can basically work the same way as they still get log-lambda variables. Probably would be good to store the type of binning used in a class variable as well for tracking this downstream (e.g. when computing P1d we'd want to know later which binning this is in, while one could check this explicitely it would be simpler to have this accessible from here...)

Waelthus (commented):
I did look through things that changed wrt the old version. I do have some comments for my understanding, but generally I think those are fine. I did not check any pure additions by delta_extraction.
I do think that the rename test -> tests is ok, but it made my life harder for this comparison as it created a couple 100 additional modified files. For the future it would be a good idea of having such pure filename changes in seperate PRs.
We should probably announce the merge on Slack and have people try out the notebooks, I did not check how well they work. And it's always good if somebody that didn't write tutorials tries out the tutorials.

Maybe we should rename tests so that the delta_extraction tests are running at a similar stage as the delta tests, so far they run last, i.e. after the fitter...

Please also bump the version to something like 5.0rc1 or so...

bin/picca_deltas.py
```diff
@@ -803,12 +816,17 @@ def main(cmdargs):
             header["NSIDE"] = nside
             header["PIXORDER"] = healpy_pix_ordering
             header["FITORDER"] = args.order
-            results.write([stack_log_lambda, stack_delta, stack_weight],
+            results.write([stack_log_lambda, get_stack_delta(stack_log_lambda),
+                           get_stack_delta_weights(stack_log_lambda)],
                           names=['loglam', 'stack', 'weight'],
                           header=header,
                           extname='STACK')
```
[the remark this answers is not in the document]
iprafols: yes, this was it

bin/picca_deltas.py
```diff
@@ -803,12 +816,17 @@ def main(cmdargs):
...
             results.write(
-                [log_lambda, eta, var_lss, fudge, num_pixels],
+                [log_lambda,
+                 Forest.get_eta(log_lambda),
+                 Forest.get_var_lss(log_lambda),
+                 Forest.get_fudge(log_lambda),
+                 num_pixels],
                 names=['loglam', 'eta', 'var_lss', 'fudge', 'nb_pixels'],
                 extname='WEIGHT')
```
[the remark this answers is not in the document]
iprafols: also yes

Waelthus:
as we're not automatically computing test coverage, these are the results for delta_extraction tests only, which are way better than previous picca_deltas if memory serves correctly... Many of the things missing testing are actually regarding P1d (which partly needs implementation still) or checks for e.g. file sanity, that we currently do not explicitely test.


---------- coverage: platform linux, python 3.8.12-final-0 -----------
Name                                                                 Stmts   Miss  Cover
----------------------------------------------------------------------------------------
bin/picca_delta_extraction.py                                           27      0   100%
py/picca/__init__.py                                                     2      0   100%
py/picca/_version.py                                                     1      0   100%
py/picca/delta_extraction/__init__.py                                    2      0   100%
py/picca/delta_extraction/astronomical_object.py                        37     18    51%
py/picca/delta_extraction/astronomical_objects/__init__.py               2      0   100%
py/picca/delta_extraction/astronomical_objects/forest.py               221     80    64%
py/picca/delta_extraction/astronomical_objects/pk1d_forest.py           73     60    18%
py/picca/delta_extraction/astronomical_objects/sdss_forest.py           34      5    85%
py/picca/delta_extraction/astronomical_objects/sdss_pk1d_forest.py       6      2    67%
py/picca/delta_extraction/config.py                                    246     61    75%
py/picca/delta_extraction/correction.py                                  4      1    75%
py/picca/delta_extraction/corrections/__init__.py                        2      0   100%
py/picca/delta_extraction/corrections/calibration_correction.py         42     18    57%
py/picca/delta_extraction/corrections/dust_correction.py                88     25    72%
py/picca/delta_extraction/corrections/ivar_correction.py                40     18    55%
py/picca/delta_extraction/corrections/optical_depth_correction.py       43     11    74%
py/picca/delta_extraction/data.py                                      123     18    85%
py/picca/delta_extraction/data_catalogues/__init__.py                    2      0   100%
py/picca/delta_extraction/data_catalogues/sdss_data.py                 131     48    63%
py/picca/delta_extraction/errors.py                                     10      1    90%
py/picca/delta_extraction/expected_flux.py                              16      3    81%
py/picca/delta_extraction/expected_fluxes/__init__.py                    2      0   100%
py/picca/delta_extraction/expected_fluxes/dr16_expected_flux.py        461    150    67%
py/picca/delta_extraction/mask.py                                        6      1    83%
py/picca/delta_extraction/masks/__init__.py                              2      0   100%
py/picca/delta_extraction/masks/lines_mask.py                           41     12    71%
py/picca/delta_extraction/quasar_catalogue.py                           24      7    71%
py/picca/delta_extraction/quasar_catalogues/__init__.py                  2      0   100%
py/picca/delta_extraction/quasar_catalogues/drq_catalogue.py           125     28    78%
py/picca/delta_extraction/survey.py                                    100      5    95%
py/picca/delta_extraction/utils.py                                      52      4    92%
py/picca/delta_extraction/utils_pk1d.py                                 55     48    13%
py/picca/tests/__init__.py                                               5      2    60%
py/picca/tests/delta_extraction/__init__.py                              2      0   100%
py/picca/tests/delta_extraction/abstract_test.py                        63     19    70%
py/picca/tests/delta_extraction/scripts_test.py                         58      4    93%
py/picca/tests/delta_extraction/test_utils.py                           74     26    65%
----------------------------------------------------------------------------------------
TOTAL                                                                 2224    675    70%

py/picca/cf.py
```diff
@@ -155,9 +155,16 @@ def compute_xi(healpixs):
 
             for delta2 in delta1.neighbours:
                 ang = delta1.get_angle_between(delta2)
-                same_half_plate = ((delta1.plate == delta2.plate) and (
-                    (delta1.fiberid <= 500 and delta2.fiberid <= 500) or
-                    (delta1.fiberid > 500 and delta2.fiberid > 500)))
+                if remove_same_half_plate_close_pairs:
```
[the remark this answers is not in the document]
iprafols: This is a new addition that is required for the deltas produced with the delta_extraction code to work. The reason behind this need is that now plate, fiberid, and mjd (when reading SDSS data) are strings including the information from the different reobserations, and this was breaking this check. 

The `same_half_plate variable` is only used when `remove_same_half_plate_close_pairs` is set to `True`. This is an option that was used when no coadding of the forests was applied to correct for correlated sky noise resiudals. It does not make sense for it to run when reobservations are coadded so it makes sense that the code crashes then. It would probably be better to improve the error reporting, but I think that we should just remove this option (or at least deprecate it)

Answering your question, `remove_same_half_plate_close_pairs` is a module variable that is set in `picca_cf.py` that is why there are no changes prior to this. It was used in the functions `compute_xi_forest_pairs` and the equivalent for the dmat and the metal dmat

py/picca/data.py
iprafols: Mostly yes, but there are some minor changes to the deltas if this correction is applied before rebinning.

py/picca/data.py
iprafols: Yes, that is correct. Going downstream is the part of the code that we still need to refactor, but this is another huge project that we will need to find someone to do. There is a lot of repeated code between `cf.py` and `xcf.py` and some issues when computing the metal matrixes with different binning. I guess we should add this tracking requirement to the list of pending issues there

setup.py
```diff
@@ -17,11 +19,18 @@
     url="https://github.com/igmhub/picca",
     author="Nicolas Busca, Helion du Mas des Bourboux et al",
     author_email="nbusca@lpnhe.in2p3.fr",
```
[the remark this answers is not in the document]
iprafols: I guess for now I'll add my name before the et al. and maybe we can revisit this on a picca task force telecon. I can also change the email since Nicola's LPNHE email is not working anymore. I'll push a commit with these changes

Waelthus:
Re the test order: It can be more useful to have the test that is time consuming and also probably the least changed part of the code last (and I think this is the fitter). When automatic tests fail, further tests are not run at all, so reordering reduces average test runtime.

I had a look at the coverage again and noticed that missing test coverage in dr16_expected_flux.py is completely due to missing tests of linear binning. Ideally we should run one test that goes through a mock dataset (could be a small part of an actual DESI mock or something randomly generated with linear binning). 
Even though the linear binning is less well studied so far it is helpful to know if things changed.

About the version: rc was for release candidate, which then would become 5.0.0 if we don't find issues within the first week or two (last time we updated the version a day or two after creating the tagged release), we could also call this 5.0.0 and then potentially have 5.0.1 later...

iprafols:
Re: tests. There is an issue with the tests, only 35 are run when there should be 107. Possibly an issue with filenames ending with `test.py` instead of beginning with `test`

iprafols:
Hi @Waelthus ,
I have modified the pytest configuration file so that the tests are recognized. However, I am not sure how to order the tests. The only thing I can come up with is moving all the delta_extraction test files to the test folder and renaming all of them, but I don't think this is a good idea

Waelthus:
I guess the order is unimportant enough to be changed later (in case we at some point find out how to do this...)

Waelthus:
added coveralls with coverage checks to the PR, see https://coveralls.io/github/igmhub/picca

pytest.ini
```diff
@@ -1,3 +1,5 @@
+# content of pytest.ini
 [pytest]
 filterwarnings =
     ignore::DeprecationWarning
+python_files = *tests.py test*
```
Waelthus, line 5: is this what heals the testing in the end?

setup.py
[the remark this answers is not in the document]
Waelthus: I'll also change the email to yours on the github repo settings...

iprafols:
We were discussing with Andreu a way to enclose all the linear/logarithmic binning into the reading functions, which would make the code cleaner, but we figured we will address this issue after the merge, to make sure we are not breaking things with last-minute chenges

py/picca/data.py
Waelthus: yes, makes sense, restructuring everything is a long-term task. But it's good to know that everything can at least still read the data. Is evenly spaced binning in LOGLAM implicitely assumed when running (x)cf? For P1d it is, but I know the changes needed from my other branch, need to port this rather sooner than later

py/picca/data.py
Waelthus: yes, that's what you always talked about when you said that corrections are now a different ordering, but this routine will become deprecated anyway... Maybe we should generally add DeprecationWarnings to picca_deltas and routines that only this uses....

py/picca/cf.py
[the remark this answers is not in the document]
Waelthus: ah, that should be ok, maybe add a deprecation warning when this option is used, so that people get a note if they still use it....

Waelthus:
great! This means truecont -> here -> master? And in that case if the PR on truecont is fine with the ppl assigned there, I'm fine with merging to master

PR #851, opened 2022-02-21
all classes except for reading classes transparent to Forest.wave_solution

iprafols:
@andreufont  mentioned that it was weird that the different classes (Corrections, Masks, ExpectedFluxes) were constantly checking for the value `Forest.wave_solution`. This should be handled by the Reading classes alone. 

This PR intends to fix this issue


andreufont (changes requested):
@iprafols , I think we should review how you create the global wavelength arrays, these try / except are a bit annoying and dangerous. It is also confusing to store a variable named delta_log_lambda that stores log(delta_lambda), something completely different.

Waelthus:
Just skimmed the changes so far. Will go through in more detail later.

I agree with Andreu on that variable, I'd rather have nothing stored in that regard and requiring downstream codes to deduce delta-(log)-lambda than having a garbage number stored in the file, and for specifying the i/o and actual rebinning in delta_extraction I think we need to have log vs linear treated somewhat seperately, there is simply no way to convert a Δlogλ to a Δλ that is valid for all λ. Not sure right now if the changes will lead to problems at some point when masking/rebinning spectra as well, note that at no point we'll want to accidentially convert the binning even if storing LOG_LAMBDA for a linear λ binning is fine.

Else, given that not all scenarios are part of the test suite yet, has there been some comparison that the data produced after these changes are the same (except for storing logλ vs storing λ) also for linear binning? Also for variables only used by P1d? I think already in the current master branch some conversion like this messes with the output values for the resolution when using the P1d mode which is output as several 100000 km/s (needs further investigation and will go to a seperate PR, but demonstrates how things can go wrong with those binnings).

Waelthus:
ah, I missed that. 

My 2ct about if we should do this at all would probably be a lets keep modes seperated if there's no strong reason not to. The reason is mostly that unification works for some parts, but not for others which might make is less clear/readable what actually happens (as you might encounter differently binned data put in logλ [where the binning is only known at runtime] vs just having either clearly spelled out in the code [but than being uniform in that unit which is visible when just looking at code]). While it might make it more complex to write new modules or update existing ones it also forces people to think about both modes at all (and e.g. not assuming regular binning in logλ because that's the only mode the implementer personally uses). And independent on how we'll deal with this, we need to have test cases for both anyway.

py/picca/delta_extraction/astronomical_objects/desi_forest.py
```diff
@@ -22,20 +22,16 @@ class DesiForest(Forest):
...
-    lambda_grid: array of float or None (from Forest)
-    Common grid in lambda_ based on the specified minimum and maximum
-    wavelengths, and delta_lambda.
-
-    lambda_rest_frame_grid: array of float or None (from Forest)
-    Same as lambda_grid but for rest-frame wavelengths
-
     log_lambda_grid: array of float or None (from Forest)
     Common grid in log_lambda based on the specified minimum and maximum
     wavelengths, and delta_log_lambda.
```
andreufont, line 27: this doc string makes it look like log_lambda_grid depends on delta_log_lambda, but when using linear pixels (that should be the default soon) it is delta_lambda instead that is used.

py/picca/delta_extraction/astronomical_objects/desi_forest.py
```diff
@@ -77,9 +73,6 @@ class DesiForest(Forest):
     ivar: array of float (from Forest)
     Inverse variance
 
-    lambda_: array of float or None (from Forest)
-    Wavelength (in Angstroms)
-
     log_lambda: array of float or None (from Forest)
     Logarithm of the wavelength (in Angstroms)
```
andreufont, line 77: Maybe add a few extra words to differentiate log_lambda from log_lambda_grid. I'm guessing that this array is shorter, since it only has some of the values in log_lambda_grid?

py/picca/delta_extraction/astronomical_objects/forest.py
```diff
@@ -289,15 +218,7 @@ def coadd(self, other):
...
-        if Forest.wave_solution == "log":
-            self.log_lambda = np.append(self.log_lambda, other.log_lambda)
-        elif Forest.wave_solution == "lin":
-            self.lambda_ = np.append(self.lambda_, other.lambda_)
-        else:
-            raise AstronomicalObjectError("Error in coadding Forest. "
-                                          "Class variable 'wave_solution' "
-                                          "must be either 'lin' or 'log'. "
-                                          f"Found: {Forest.wave_solution}")
+        self.log_lambda = np.append(self.log_lambda, other.log_lambda)
```
andreufont, line 221: Is numpy.append the right thing to use here? Is it ok if the log_lambda array is not sorted? It might be, just asking.

py/picca/delta_extraction/astronomical_objects/forest.py
```diff
@@ -343,11 +264,11 @@ def get_data(self):
             units += ["log Angstrom"]
             array_size = self.log_lambda.size
         elif Forest.wave_solution == "lin":
```
andreufont, line 266: I understand that log_lambda is the variable used now by the Forest class. Wouldn't it be possible then to store just log_lambda always, regardless of what the wave solution in reading was?

py/picca/delta_extraction/astronomical_objects/forest.py
```diff
@@ -440,63 +361,32 @@ def rebin(self):
...
-            self.transmission_correction = self.transmission_correction[w1]
-
-            bins = find_bins(self.lambda_, Forest.lambda_grid)
-            self.lambda_ = Forest.lambda_grid[0] + bins * delta_lambda
-        else:
-            raise AstronomicalObjectError("Error in rebinning Forest. "
-                                          "Class variable 'wave_solution' "
-                                          "must be either 'lin' or 'log'. "
-                                          f"Found: {Forest.wave_solution}")
+        delta_log_lambda = Forest.log_lambda_grid[1] - Forest.log_lambda_grid[0]
```
andreufont, line 364: When using linear pixels, delta_log_lambda will change with lambda. I guess this is ok, since this is only to have a buffer near the edge of our lambda_grid, but it would be good to add a comment in the code to clarify this.

py/picca/delta_extraction/astronomical_objects/forest.py
```diff
@@ -440,63 +361,32 @@ def rebin(self):
...
+            self.ivar = np.array([])
+            self.transmission_correction = np.array([])
+            return [], [], [], [], []
+        self.log_lambda = self.log_lambda[w1]
+        self.flux = self.flux[w1]
+        self.ivar = self.ivar[w1]
+        self.transmission_correction = self.transmission_correction[w1]
+
+        bins = find_bins(self.log_lambda, Forest.log_lambda_grid)
+        self.log_lambda = Forest.log_lambda_grid[0] + bins * delta_log_lambda
```
andreufont, line 389: Wait, now I'm very confused. Are we using delta_log_lambda to setup the grid, even when running with linear pixels? @iprafols

py/picca/delta_extraction/astronomical_objects/forest.py
```diff
@@ -577,21 +475,21 @@ def set_class_variables(cls, lambda_min, lambda_max,
                 np.log10(lambda_min_rest_frame) + pixel_step/2,
                 np.log10(lambda_max_rest_frame),
                 pixel_step)
-            cls.mask_fields = defaults.get("mask fields log").copy()
         elif wave_solution == "lin":
-            cls.lambda_grid = np.arange(
+            cls.log_lambda_grid = np.log10(np.arange(
                 lambda_min,
                 lambda_max + pixel_step/2,
-                pixel_step)
```
andreufont, line 585: Why is the log_lambda_grid not aligned with the input pixels? 

When running the code without rebinning (our baseline analysis in early-3d paper), this would end up shifting the wavelengths by 0.4 A, isn't it?

py/picca/delta_extraction/corrections/ivar_correction.py
```diff
@@ -43,10 +43,8 @@ def __init__(self, config):
             hdu = fitsio.read(filename, ext="VAR_FUNC")
             if "loglam" in hdu.dtype.names:
```
andreufont, line 44: Related to the comment above, I'm not sure we need to allow for these different HDU names (unless we need it for compatibility with older delta files)

py/picca/delta_extraction/expected_fluxes/dr16_expected_flux.py
```diff
@@ -839,27 +649,14 @@ def compute_var_stats(self, forests):
...
-                # filter the values with a pipeline variance out of range
-                lambda_bins = lambda_bins[w]
-                # compute overall bin
-                bins = var_pipe_bins + num_var_bins * lambda_bins
-            else:
-                raise ExpectedFluxError("Forest.wave_solution must be either "
-                                        "'log' or 'linear'")
+            log_lambda_bins = find_bins(
+                forest.log_lambda,
+                self.log_lambda
```
andreufont, line 654: wow, this line is confusing. Are you sure one of these shouldn't be log_lambda_grid?

andreufont (changes requested):
Thank you very much for putting together all of this, it looks like an important amount of work!

Overall I like the changes a lot, since it makes several pieces of code a lot simpler, hiding the pixelisation from the rest of Picca. However, I did add several comments that I'd like to discuss before merging.

py/picca/delta_extraction/astronomical_objects/forest.py
Waelthus: It does work as expected as the rebin function at the end of coadd will take care of sorting everything into output wavelength bins in the end. So all that is really needed is having aligned arrays with merged self and other quantities

py/picca/delta_extraction/astronomical_objects/forest.py
Waelthus: it is aligned, it starts with `log(lambda_min)` as the first output pixel, and ends with the last `lambda_min+N*pixel_step` that is smaller than `lambda_max+pixel_step/2`. I guess one could make the upper end something only mildly larger than `lambda_max`, but else it's fine.

andreufont: Argh, sorry, I missread the code. I thought lambda_min had half step added. Sorry for the noise!

py/picca/delta_extraction/astronomical_objects/forest.py
andreufont: Great. Thanks!

py/picca/delta_extraction/expected_fluxes/dr16_expected_flux.py
Waelthus: I think this might be right, `self.log_lambda` should be the wavelength bins used for eta/var_lss/fudge whereas forest.log_lambda are the wavelength bins of an individual forest after rebinning to the common grid. But @iprafols should confirm. 
Also according to the docstring self.log_lambda is in restframe, not sure right now about forest.log_lambda. Might be useful to make very clear distinctions in the variable names between rf and obs frame; and between eta-lambda and eg mean_cont lambda

py/picca/delta_extraction/astronomical_objects/forest.py
Waelthus: Indeed it looks like this. I am very strongly against doing this! It will mess up the whole Pk1d computation as the number of pixels (and thus the resulting overall noise) in a wavelength bin will become strongly wavelength dependent and the resolution matrix becomes completely meaningless if we do non-linear regriddings in the analysis ever. For that purpose I'm also suggesting to merge #850 before finishing this merge to be sure we catch all the dependencies on wave solution needed for P1d

py/picca/delta_extraction/astronomical_objects/forest.py
Waelthus: Agreed! But as you noticed further down, this does not seem limited to just having a buffer.

py/picca/delta_extraction/astronomical_objects/forest.py
Waelthus: Maybe wavelength itself is a little nicer as the unit can be put into the file explicitely (loglam is both unitless and could follow different log-definitions in principle), with putting Lamba here there would be a unique definition. Also at least for DESI purposes linear binning should become the norm, so this would be a future vs past choice where I'd opt for future...

iprafols: Yes, we can do it like this. In that case I think we should store `wave solution` (and maybe `delta_lambda` or `delta_log_lambda`) as it might be useful to know how were the files computed

py/picca/delta_extraction/astronomical_objects/desi_forest.py
iprafols: I added a commit changing the docstrings. I updated this description in the Forest docstring. I decided to simplify the docstrings to only contain the elements defined there and leave the inherited attributes on the parent class. This is to simplify maintenance.

py/picca/delta_extraction/astronomical_objects/desi_forest.py
iprafols: See also the commit on updated docstrings

py/picca/delta_extraction/astronomical_objects/forest.py
iprafols: this is a mistake from the merge conflicts. Here we should see the linear/logarithmic pixelisation to ensure that the chosen binning is kept. I'll restore this in a commit

py/picca/delta_extraction/corrections/ivar_correction.py
iprafols: I left this for compatibility with older files. Until this is merged, the picca_delta_extraction saves the attributes files with `lambda` on it. We can add a `DeprecationWarning` and remove after a while

py/picca/delta_extraction/expected_fluxes/dr16_expected_flux.py
Waelthus: I think that would be helpful for the reader. `log_lambda_var_func_grid` would be even more explicit, but also even longer. Similarly I'd recommend renaming the lambda_rest_frame of the mean cont (in case this hasn't a speaking name yet) so that it is always very explicit which lambda exactly is which.

andreufont: I like log_lambda_var_grid

py/picca/delta_extraction/astronomical_objects/forest.py
iprafols: Just to add to what Michael was saying, `lin` is already the default choice. I'm happy to do it either way (always saving `log_lambda` or saving `lambda` depending on the choice for wavelength solution)

py/picca/delta_extraction/astronomical_objects/forest.py
iprafols: I have now restored the rebinning function to what it should be

py/picca/delta_extraction/expected_fluxes/dr16_expected_flux.py
iprafols: I don't think longer names are a problem so I say we go with @Waelthus  suggestion. 

For the mean cont, we use Forest.lambda_rest_frame_grid so I take it we don't need renaming there

py/picca/delta_extraction/astronomical_objects/forest.py
andreufont: I agree with Michael that `lambda` is a bit better than log-lambda. However, it would be great if we could still read older delta files that had `log_lambda`. And once you accept that you'll have to be able to read from both formats, it is not too bad if we also allow the writing to be in two formats. So I'm happy to leave the two options as it is now.

Waelthus: Ah, sorry. Misread this. Agreed that saving in whatever format was used in the computation makes a lot of sense.

py/picca/delta_extraction/corrections/ivar_correction.py
andreufont: Let's leave both options for compatibility.

iprafols:
I think for these changes we can trust the tests, @Waelthus  do you agree?

Waelthus (commented):
Ok, I didn't find anything obviously problematic, just 2 comments left from my side:
1) regarding trust in the tests: It looks like the test outputs have changed, At least the deltas taken for comparison were updated. Were those only variable name changes or did the actual values in the fields change? If the latter, by how much? Has any continuum fitting been tested on linear binning? I don't think it has been (at least this was not evaluated previously https://coveralls.io/builds/47541415/source?filename=py%2Fpicca%2Fdelta_extraction%2Fexpected_fluxes%2Fdr16_expected_flux.py#L478), so without a manual comparison under linear binning we do not know. But maybe I don't see this right?
2) The find_bins function is (at least mostly) run on log_lambda, i.e. it will sort the pixels using a metric living in log_lambda space, Do we want this in both binnings? It might be ok to do so for eg the variance functions, stack or even cont fitting (it is never ok for resampling the spectrum). But we should be clear about what happens there in detail... Or did I misunderstand something?

iprafols:
About 1, the deltas were changed to include more fields, the existing ones were unchanged. About the linear test, I will do a comparison run of a small subset of DESI data between master and this new branch. We can then include it as an automatic test.
About 2, it is true that the metric we are using is log_lambda, I'm not sure what would be the impact of this. In the meantime, we could have find_rebin know what Forest.wave_solution is and change the metric accordingly (by doing 10**x whenever necessary)

Waelthus:
if we do your suggestion on 2) we should be able to achieve the same results as previously. Else those would be expected to change for the linear binning (that assumed a linear metric before and log metric now). So that shows again how useful it would be to auto-test both (as in this case for log nothing should and nothing does change, but for lin there shoud be changes that are not tested, so tests are still ok without us having an estimate of the magnitude of change...)

iprafols:
Let's do this for now and then we can check the magnitude of the change on another PR. I'll leave a todo comment in the code. I'll push a commit with the changes and check that the linear binning is not changing either.

Waelthus (approved):
fine with me, but please put that todo in....

iprafols:
Ok so I restored the behaviour for linear binning and added the todo. I'll merge once I check that there are no changes on the linear binning side

iprafols:
I have tested the linear binning mode. There are some changes with respect to the results from the master. I closely inspected the results and found that there were a couple of minor bugs, which I have corrected in the last commit. Also, the remaining differences can be fully explained and are fixing some issues from the master branch:
- During initialization of `sigma_lss` in master for the linear binning was missing the 0.2 that the logarithmic binning had. In simplify_wave_solution both binning had the 0.2
- `get_valid_fit` in master for the linear binning was initialized with `num_pixels`
- `get_valid_fit` in master for the linear binning was kept without `fill_value=“extrapolate”` it had `fill_value=0` instead
- Small change in the test results for `compute_mean_cont_lin` (5th significant figure) due to interpolation of the results
- Small changes in the number of pixels used to compute `eta`, `var_lss`, and `fudge` due to minor fluctuations in the stored value of the wavelengths resulting in it being closer to one bin than the other (even though wavelengths are equal according to `np.isclose`). This only happens when the wavelength is in the middle point between two wavelengths of the common grid
- The rest of the changes (which is most of the changes) occurs because the master branch uses `A*log_lambda + B` for the continuum fit only for the logarithmic bin and this new branch uses it for both. For the linear binning, the master branch uses `A*lambda + B` instead. We might consider changing this in the future, but it was an unintentional change when the linear binning was split from the logarithmic binning. Before any such change is decided, we should investigate the effect of this change.

I have not included the linear tests here as they require the usage of a small amount DESI data and I have yet to ask for permission to publish it. I will include them in the add_test PR that I will shortly sumbit. 

@Waelthus , @andreufont , are you happy with me merging this branch?

andreufont:
Sounds good!

iprafols:
I would maybe wait until we add the tests to do the new release. I will now ask permission to upload the necessary data which means we should not need much time to add this other PR. This way the tag would be significantly better.

PR #854, opened 2022-02-22
make dla masking independent of wavelength basis

Waelthus:
This should not change anything for logλ analyses, but allows DLA masking for linear λ analyses (which crashed before, but does not crash with this). If I read this correctly the DlaProfile expects λ anyway and this only affects the masking related to DLAs which has not been treated for both cases.


iprafols (approved):

iprafols:
Hi @Waelthus , 
I made some slight changes to the file in order no avoid the operation 10**np.log10(x). If they are ok for you then from my point we can go ahead with the merging

Waelthus:
looks good to me.

PR #856, opened 2022-02-28
fixed args

iprafols:
Very small change fixing a mistake in variable naming. Note that this was not caught by the automatic tests and so they should be extended to also cover this function


Waelthus (approved):
thanks for fixing this here

PR #857, opened 2022-02-28
rebin moved outside of init

iprafols:
This intends to fix an issue related with multiple inheritance. We need all the variables to be properly initialized before rebinning and the best way to deal with this is to run rebinning after initialisation is complete.


Waelthus (approved):

Waelthus:
I'm fine with this change, to elaborate on the need:
before we rebinned during init which depending on the order of the rebins could leave us with some arrays (e.g. Forest quantities like lambda, flux) being already rebinned and others (e.g. Pk1d only quantities like resolution) still having their old size when their rebin function gets called. This would then fail as the lambda array is out-of-sync from the start of rebinning (and thus the rebinning could not be figured out anymore).
Thanks for the change!

PR #858, opened 2022-03-01
Add keep survey

iprafols:
This PR allow us to filter a given catalogue to only use a particular "survey". The reason behind this is that the new catalogues have more than one survey on them (e.g. guadalupe has "special" and "main"). Also this will help us better test the loading of new data (currently there are some potential issues when reading it (see the slides for the KP6 telecon from March 1).

Also, I added some tests for the DesiQuasarCatalogue class


py/picca/delta_extraction/quasar_catalogues/desi_quasar_catalogue.py
```diff
@@ -59,7 +68,11 @@ def __init__(self, config):
         self._parse_config(config)
 
         # read quasar catalogue
-        self.catalogue = self.read_catalogue()
+        self.read_catalogue()
+
+        # if not all surveys are specified, then filter the catalogue
+        if "all" not in self.keep_surveys:
```
Waelthus, line 74: this needs to be documented somewhere, this would also include special tiles, ie use more than the default

Waelthus (approved):
looks good to me, needs documentation so that people know how to use the option (and that it's there at all...)

py/picca/delta_extraction/quasar_catalogues/desi_quasar_catalogue.py
iprafols: I'll change `all` to not include the special tiles.  I'll also add this in the configuration tutorials

iprafols: this should be addressed now

Waelthus (approved):

Issue #847, opened 2022-02-17
true continuum branch should use mean flux measured from transmission skewers


andreufont:
Right now it uses a 2nd order polinomial in effective optical depth, but we should be using the actual <F>(z) measured from the skewers. This should be stored in the same files containing var_lss, so it should be easy to implement.

@andreicuceu , @iprafols , @alxogm , @cramirezpe

andreufont:
addressed in #860

PR #860, opened 2022-03-03
Fix true continuum

cramirezpe:
Opened a pull request to discuss changes in the branch. I tried to address the issues exposed in #847. I made a couple of plots with `delta_attributes.fits.gz` from both version (with and without the changes). But not sure how to interpret its result. 

I changed the code to read the mean continuum from mocks. In the last commit I switch to use this mean flux in the continuum computation. I would like to get help on this last commit (that I think is not correct).


alxogm:
I was looking at the code, but when I came back the plots were not longer displayed ... I think the differences were small and on the edges of the wavelength range, which is good, because is where I was seeing some weird stuff happening. 
I think it looked well. 

But I'm confused about your comment below, do you mean you need to revert back that commit? Or you need someone to check/run the code?

> I changed the code to read the mean continuum from mocks. In the last commit I switch to use this mean flux in the continuum computation. I would like to get help on this last commit (that I think is not correct).

py/picca/delta_extraction/expected_fluxes/true_continuum.py
```diff
@@ -129,7 +129,8 @@ def _parse_config(self, config):
 
         self.num_processors = config.getint("num processors")
 
-        self.var_lss_binning = config.get("var lss binning", 'log')
+        self.raw_statistics_binning = config.get("raw statistics binning", 'log')
```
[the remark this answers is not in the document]
andreufont: Also, why do you need to specify the pixel? Isn't this something that you would get from the raw_statistics_file?

py/picca/delta_extraction/expected_fluxes/true_continuum.py
```diff
@@ -307,76 +307,67 @@ def read_true_continuum(self, forest):
...
-            z = 10.**forest.log_lambda / lambda_rest_frame - 1.
+            forest.continuum *= self.get_mean_flux(10**forest.log_lambda)[0]
 
         elif Forest.wave_solution == "lin":
             forest.continuum = true_continuum(forest.lambda_)[0]
-            mean_optical_depth = np.ones(forest.lambda_.size)
-            tau, gamma, lambda_rest_frame = 0.0023, 3.64, 1215.67
-            w = forest.lambda_ / (1. + forest.z) <= lambda_rest_frame
-            z = forest.lambda_ / lambda_rest_frame - 1.
+            forest.continuum *= self.get_mean_flux(forest.lambda_)
```
andreufont, line 325: the option for log lambda above ends with [0], but not the one for linear lambda. Why?

py/picca/delta_extraction/expected_fluxes/true_continuum.py
```diff
@@ -307,76 +307,67 @@ def read_true_continuum(self, forest):
...
-            var_lss_file = resource_filename('picca', 'delta_extraction')
-            var_lss_file += '/expected_fluxes/var_lss/' + filename
+            filename = resource_filename('picca', 'delta_extraction') + '/expected_fluxes/var_lss/'
+            if self.raw_statistics_binning == 'log':
+                filename += 'colore_v9_lya_log.fits.gz'
+            elif self.raw_statistics_binning == 'lin_2.4':
+                filename += 'colore_v9_lya_lin_2.4.fits.gz'
+            elif self.raw_statistics_binning == 'lin_3.2':
+                filename += 'colore_v9_lya_lin_3.2.fits.gz'
+        self.logger.info(f'Reading raw statistics var_lss and mean_flux from file: {filename}')
```
andreufont, line 347: I guess here you are trying to guess which file to use, if the user has not provided the raw_stats file.

However, why do you ask for an extra variable (raw_stats_binning) instead of checking which binning you are actually using in the true continuum analysis? These should always be synchronised anyway, so no need to specify two different pixelisations.

cramirezpe:
> I was looking at the code, but when I came back the plots were not longer displayed ... I think the differences were small and on the edges of the wavelength range, which is good, because is where I was seeing some weird stuff happening. I think it looked well.
> 
> But I'm confused about your comment below, do you mean you need to revert back that commit? Or you need someone to check/run the code?
> 
> > I changed the code to read the mean continuum from mocks. In the last commit I switch to use this mean flux in the continuum computation. I would like to get help on this last commit (that I think is not correct).

Sorry for the confusing message. I need someone to check if the implementation I made is correct. I reupload the plots I sent here in case they are helpful. 
weight_true_fix
mean_cont_true_fix

py/picca/delta_extraction/expected_fluxes/true_continuum.py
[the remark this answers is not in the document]
cramirezpe: There are three raw_statistics_files that can be read right now. One with log pixelization and two with two linear schemes. I think Alma implemented it in this way because the pixelization you use in your delta_extraction run could not be in any of the raw_statistics_files. In that case you may want to select the one that agrees the most.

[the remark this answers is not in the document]
cramirezpe: Regarding the log pixelization, since that is the default pixelization in pica (until now at least), it is good to preserve it as default also here(?)

py/picca/delta_extraction/expected_fluxes/true_continuum.py
cramirezpe: This is an error on my side. In both cases the [0] should not be there.

py/picca/delta_extraction/expected_fluxes/true_continuum.py
cramirezpe: This is because the pixelization in the true continuum analysis could be different from the ones available in raw_statistics_files. There are two options: 
- providing a raw_statistics_filename in the configuration: will directly read this file
- providing a pixelization scheme: will read the stored raw_statistics_files with the correspondent pixelization.

The second option is to allow for computing true deltas without having to compute a compatible raw analysis.

andreufont: In my opinion, if the pixelisation in the true continuum analysis is not the same that was used in the raw analysis, the code should crash.

andreufont: Yes, having these two options is clearer. It's ok, let's leave it as it is for now, as long as there is a check for consistency between the pixels (and restframe wavelength) used, I'm ok with it.

py/picca/delta_extraction/expected_fluxes/true_continuum.py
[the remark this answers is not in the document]
alxogm: Please keep the log pixelization, I'm using it for other projects with Andrea, we'll eventually move to the linear pixelization but not yet.

py/picca/delta_extraction/expected_fluxes/true_continuum.py
alxogm: Yes @cramirezpe that was always the idea to have to two options, at the moment I just left it as one single argument so I didn't break anything as we wanted to merge. thanks for splitting. I agree we need a check for consistency in any case.

alxogm:
@cramirezpe I think you need to add "raw statistics file" to the accepted options in line 19. Other than that, the computation of the deltas seems right. Here are a couple of examples (from a noiseless eBOSS mock I wanted to check) : 

Screen Shot 2022-03-03 at 19 07 16

It is still not perfect, but is better than the previous version.

py/picca/delta_extraction/expected_fluxes/true_continuum.py
andreufont: Ok, let's merge as is and let's open an issue to improve this later on. Thanks both!

py/picca/delta_extraction/expected_fluxes/true_continuum.py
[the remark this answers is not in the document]
andreufont: Ok, let's leave it for now.

andreufont (approved):
Looks good, and the plots shared by Alma are convincing. Please add the small comment requested by Alma, but happy to merge!

py/picca/delta_extraction/expected_fluxes/true_continuum.py
```diff
@@ -95,7 +95,7 @@ def __init__(self, config):
 
         # read large scale structure variance
         self.get_var_lss = None
-        self.read_var_lss()
+        self.read_raw_statistics()
```
iprafols, line 98: Before this line I'm missing a `self.get_mean_flux = None`. All members should be declared in the __init__ function

iprafols (changes requested):
Just a minor thing. Also it would be nice to update the data model with the new members (`get_mean_flux`). @cramirezpe you can check the tutorials to see how to updated it. Let me know if something is unclear or if you need help

py/picca/delta_extraction/expected_fluxes/true_continuum.py
```diff
@@ -351,6 +351,34 @@ def read_raw_statistics(self):
...
+                or not np.isclose(header['L_MIN'], Forest.lambda_min , rtol=1e-3)
+                or not np.isclose(header['L_MAX'], Forest.lambda_max , rtol=1e-3)
+                or not np.isclose(header['LR_MIN'], Forest.lambda_min_rest_frame, rtol=1e-3)
+                or not np.isclose(header['LR_MAX'], Forest.lambda_max_rest_frame, rtol=1e-3)
+                or not np.isclose(header['DEL_L'], Forest.delta_lambda, rtol=1e-3)
+            ):
+                raise ExpectedFluxError(f'''raw statistics file pixelization scheme does not match input pixelization scheme. 
+                \tL_MIN\tL_MAX\tLR_MIN\tLR_MAX\tDEL_LL
+                raw\t{header['L_MIN']}\t{header['L_MAX']}\t{header['LR_MIN']}\t{header['LR_MAX']}\t{header['DEL_LL']}
+                input\t{Forest.lambda_min}\t{Forest.lambda_max}\t{Forest.lambda_min_rest_frame}\t{Forest.lambda_max_rest_frame}\t{Forest.delta_lambda}''')
```
andreufont, line 380: Thanks for adding this consistency test. Have you checked that it works, when you run it with the wrong setting?

cramirezpe: What do you mean by it works? I thought we wanted it to fail if different pixelization schemes were given for raw_stats_file and input

cramirezpe:
I have added a couple of things, and after adding the changes in the data model this should be ready to merge.

New changes are:
- I've removed completely any pixelization scheme as input for the raw statistics file to use. The code will search for a compatible file, if there are no compatible files it will crash.
- The only option now is to provide a custom file, searching for compatible files is a fallback in case no file is provided.
- I've made changes in the raw statistics files because the field 'LR_MAX' in the header was called 'LR_MAN' instead (this also causes a change in the io.py file).

py/picca/delta_extraction/expected_fluxes/true_continuum.py
andreufont: @cramirezpe - Don't you need to declare get_var_lss as well? I believe you need both, isn't it?

andreufont:
The changes look good, thanks! I believe you also need to add self.get_var_lss=None in the init function, but the rest looks ready to merge.

py/picca/delta_extraction/expected_fluxes/true_continuum.py
cramirezpe: Yes, it has to be there as well.

py/picca/delta_extraction/expected_fluxes/true_continuum.py
iprafols: This should be loaded via the `defaults` variable. This means adding between lines 22 and 23 a line with `"raw statistics binning" = "log"` and then replace this line with `self.raw_statistics_binning = config.get("raw statistics binning")`. This is necessary to include the default value in the .config.ini file that is  kept in the output folder

py/picca/delta_extraction/expected_fluxes/true_continuum.py
iprafols: I would add a test using the wrong settings and expecting the ExpectedFluxError

iprafols (changes requested):

iprafols (approved):

iprafols:
@andreufont I think we are ready to merge. Do you think this closes issue #847? Should we close it as well?

andreufont:
Sounds good!

PR #861, opened 2022-03-04
Add num pixels in logs

iprafols:
These are some modifications to inlcude the number of pixels used to compute each of the values of eta, var_lss and fudge. To aovid confusion, another array, `valid_fit`, is also added. Tests suites are updated as the previous files were missing the two fields. I have cheked that they are the same (except for the new fields) before updating


py/picca/delta_extraction/expected_fluxes/dr16_expected_flux.py
```diff
@@ -1103,6 +1159,14 @@ def chi2(eta, var_lss, fudge):
                                       fudge[w],
                                       fill_value="extrapolate",
                                       kind="nearest")
+            self.get_num_pixels = interp1d(self.lambda_[w],
+                                      num_pixels[w],
+                                      fill_value="extrapolate",
+                                      kind="nearest")
+            self.get_valid_fit = interp1d(self.lambda_[w],
```
andreufont, line 1166: You are interpolating an array of booleans here... Is this what you wanted to do? Where is this function get_valid_fit used?

andreufont (approved):
I found it a bit strange to setup an interpolating function of boolean variables, but I'm not sure where it is used (if at all). The rest looks good I think!

iprafols:
It is not used anywhere. It is just set as an interpolation just to have the same structure as get_eta, get_var_lss and get_fudge, and get_num_pairs so that they are llinked together.

PR #863, opened 2022-03-10
copy existing logs to prevent accidental deletion

Waelthus:
currently, if picca_delta_extraction is run in overwrite mode (not sure if this is True even without), any preexisting log is overwritten early on. This PR will rename previous logs based on their modification timestamp instead.

### DIFF
--- a/py/picca/delta_extraction/config.py
+++ b/py/picca/delta_extraction/config.py
@@ -525,6 +525,10 @@ class Config:
         usages. The file is saved under the name .config.ini and in
         the self.out_dir folder
         """
-        config_file = open(f"{self.out_dir}/.config.ini", 'w')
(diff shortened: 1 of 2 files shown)


iprafols (commented):
Would it not be better to append the time to the end of the log file automatically?

Waelthus:
I'd guess it's nice to have a name without any weird additional numbers. This is just meant to be a saveguard against accidental overwriting, which you'd normally not want to do anyway (but which can happen and for the log even killing the job after doesn't bring it back).

iprafols:
Then I guess we should do the same for the .config.ini?

iprafols (approved):

iprafols:
This way we can fully revert an accidental overwrite

PR #867, opened 2022-03-14
Modular compute continuum

cramirezpe:
Ignasi and I developed this branch in order to improve the reusability of the class Dr16ExpectedFlux for future analyses with DESI early data.

Since the tests pass and changes are minimal and only internal, we consider that we can already merge this into master.


iprafols:
there is no change per se, the results are exactly the same as before. This is just a reordering of some functions so that we can better reuse them to test alternative algorithms. Basically, the `chi2` and `get_cont_model` functions that were defined inside `compute_continuum` are defined outside that function. For the chi2 we created another class so that iminuit can handle some extra parameters, for the `get_cont_model` we simply defined it as a member of `Dr16ExpectedFlux`. Finally, we added another function to compute the weights used in the continuum fitting. 

Doing this will allow us to test several things by simply inheriting from Dr16ExpectedFlux and modifying a couple of methods (which means less debugging)

Waelthus:
yes, that seems like a reasonable thing to do. It definitely makes sense to not define those things within the cont fitting routine only and having easier debugging is always good. Will have a look tomorrow

iprafols (approved):

Waelthus (approved):
Looks good to me, will interact with the changes in #865, but those can be solved in there...

PR #870, opened 2022-03-29
add option for SNR cut (default 0 for BAO, 1 for Pk)

Waelthus:
This PR turns a hard-coded setting for PK1d into an optional parameter in the ini file. It's set-up to the current default values of SNR>0 for BAO and SNR>1 for Pk1d, but then applied to both, i.e. allowing a) to change the Pk1d cuts and b) to apply cuts on BAO if ever needed.


py/picca/delta_extraction/data.py
```diff
@@ -30,6 +31,7 @@
     "lambda min rest frame": 1040.0,
     "minimum number pixels in forest": 50,
     "rejection log file": "rejection_log.fits.gz",
+    "minimal snr": None,
```
iprafols, line 34: I would not add `None` as a value here. Also, it's not trivial to figure out the default value. How about adding `minimal snr bao3d` and `minimal snr pk1d` with the actual defaults here? I would then replace the block in line 209 by
```
if self.analysis_type == "BAO 3D":
    self.min_snr = config.getfloat("minimal snr bao3d")
elif self.analysis_type == "PK 1D":
    self.min_snr = config.getfloat("minimal snr pk1d")
if self.min_snr is None:
    raise  DataError(
                "Missing arguments 'minimal snr bao3d' (if 'analysis type' = 'BAO 3D') or ' minimal snr pk1d' (if 'analysis type' = 'Pk1d') required by Data")
```

iprafols: I'm also missing a `self.min_snr = None` between lines 95 and 96. Variables should always be declared in the `__init__` method

iprafols (changes requested):
I would add the defaults in a slightly different way so that all the actual default values are hard-coded in the same place (easier to maintain). Before we merge, we should also include these new options to the configuration tutorial and to the data model

py/picca/delta_extraction/data.py
Waelthus: hm, it's a bit weird to have two parameters dealing with that, would it be fine having a None default for this and defining 2 defaults that are loaded in the respective case?

Waelthus: ok, I think I get why the None is bad... It will screw up the testing environment.

Waelthus: changed it according to your suggestion

Waelthus:
yes, we will need this in the data model, tests and docs....

py/picca/delta_extraction/data.py
Waelthus: or would you think that having 2 default values defined for a joint config parameter might work? My previous testing issue was actually something else

iprafols (approved):

py/picca/delta_extraction/data.py
iprafols: having two default values for a single parameter should work as well, as you wish

iprafols:
The changes look fine, I would just update the tutorials and data model before merging though

iprafols:
that is tricky, that is why I suggested using the two variables in the first place. The config class do not actually read the arguments as they might not be there for some classes

PR #871, opened 2022-03-29
Add option to interpolate the blinding template for non-standard binning

andreicuceu:
I added a new option in picca_export to interpolate the blinding template from a precomputed regular grid (now available on NERSC). This option is automatically activated for non-standard binning (i.e. 50 bins of 4 Mpc/h in rp and rt for the auto and double the rp bins for the cross) of the correlation. I have not tested this yet.

I also updated the default blinding template from v1 to v1.2. The new file has the same seed for blinding ap/at, but uses the new default version of vega where we do not allow metal correlations to have ap/at different from 1. This has a negligible impact on the blinding template, but it gets picked up in the total chisq which is used as a safety check in the blinding script. Therefore, an update was in order.

### DIFF
--- a/bin/picca_export.py
+++ b/bin/picca_export.py
@@ -4,6 +4,7 @@ import sys
 import argparse
 import fitsio
 import numpy as np
+import scipy.interpolate
 import scipy.linalg
 import h5py
 import os.path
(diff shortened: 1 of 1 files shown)


bin/picca_export.py
```diff
@@ -267,15 +268,45 @@ def main(cmdargs):
...
+        else:
+            # Read the regular grid blinding file and get the right template
+            blinding_filename = ('/global/cfs/projectdirs/desi/science/lya/y1-kp6/'
+                                 'blinding/y1_blinding_v1.2_regular_grid_29_03_2022.h5')
+            if not os.path.isfile(blinding_filename):
+                raise RuntimeError("Missing blinding file. Make sure you are running at"
+                                   " NERSC or contact picca developers")
+            blinding_file = h5py.File(blinding_filename, 'r')
+            hex_diff = np.array(blinding_file['blinding'][args.blind_corr_type]).astype(str)
+            diff_grid = np.array([float.fromhex(x) for x in hex_diff])
```
iprafols, line 302: Up until this line, the two blocks inside the if/else are the same, so they could be moved outside the if/else bloc.

iprafols (changes requested):
I suggest doing some small changes to simplify code

bin/picca_export.py
andreicuceu: Done. Still have to test it before merging. Will update you when I'm done.

iprafols (changes requested):
Changes look fine, from my side we can merge after the testing is done. I keep the request changes for the moment in case something needs to be changed due to the tests results

andreicuceu:
@iprafols We tested the branch, and it's ready to be merged.

PR #875, opened 2022-04-01
Move num processors to general, add parallel healpix I/O

Waelthus:
Moved the number of processors to the general section to allow running mock I/O also with just one CPU. Setting `num processors` to 0 will use half the node.

Also includes a crude implementation for parallel reading of DESI healpix spectra copied over from desisim_mocks mode.


py/picca/delta_extraction/data_catalogues/desisim_mocks.py
```diff
@@ -60,6 +60,10 @@ def __init__(self, config):
 
         self.logger = logging.getLogger(__name__)
         super().__init__(config)
+        if self.use_non_coadded_spectra:
+            self.logger.Warning('the "use_non_coadded_spectra" option was set, '
```
iprafols, line 64: This should be `self.logger.warning` note the capital W that is currently in

py/picca/delta_extraction/expected_flux.py
```diff
@@ -35,13 +36,21 @@ class ExpectedFlux:
     def __init__(self, config):
         """Initialize class instance"""
         self.los_ids = {}
+        self.num_processors = None
```
iprafols, line 39: This should be added in the docstring, under Attributes

py/picca/tests/delta_extraction/data/.config.ini
```diff
@@ -49,11 +50,12 @@ limit eta = (0.5, 1.5)
 limit var lss = (0.0, 0.3)
 num bins variance = 20
 num iterations = 5
-num processors = 1
 order = 1
 use constant weight = False
 use ivar as weight = False
 out dir = /Users/iperezra/software/picca/py/picca/tests/delta_extraction/results/config_tests/
+num processors = 1
```
iprafols, line 57: This should not be neeeded, right? This value should be on the general section only. Not sure why you remove it in line 52 and re-add it here

tutorials/delta_extraction/picca_delta_extraction_configuration_tutorial.ipynb
```diff
@@ -284,6 +284,7 @@
     "This section controls the extraction of deltas. It specifies the class that will be used to compute the quasar continua and the mean transmission. The main arguments are\n",
     "- `type`: Contains the name of the class used to extract the deltas. For example \"Dr16ExpectedFlux\". Basically can be any data type stored in the folder `py/picca/delta_extraction/expected_fluxes/`. **Type: str, Required: yes**\n",
     "- `module_name`: Name of the file containing the expected flux type. If the file is named following the convention this shoul normally not be necessary. **Type: str, Required: no**\n",
+    "- `num processors`: Number of processors to be used to compute the mean continua and to read in data for DesiHealpix based modes. 0 will use half the number of available cores.  **Type: int, Required: no, Default: 0**\n",
```
iprafols, line 287: This should not be here, this should be in the general section

iprafols (changes requested):
Some minor things need tweaking, but otherwise looks fine.

py/picca/tests/delta_extraction/data/.config.ini
Waelthus: it is, I tried out the tests (without that num processors which I originally deleted and then re-added at possible a different position) and as we're copying the value into the structure it's written for general and all children (ie data and expected flux for this case).

Waelthus: if you mean that changing the position should not be needed: that is true...

iprafols (approved):

PR #878, opened 2022-04-08
Add tests

iprafols:
This commit includes more testing to improve the coverage. In particular it includes tests for the linear binning case


py/picca/tests/delta_extraction/data_tests.py
```diff
@@ -126,26 +545,726 @@ def test_data_filter_forests(self):
...
+        self.assertTrue(len(data.forests) == 63)
+
+        # run with 0 processors; case: only sv data
+        config = ConfigParser()
+        config.read_dict({"data": {
+            "catalogue": f"{THIS_DIR}/data/QSO_cat_fuji_dark_healpix.fits.gz",
+            "keep surveys": "all",
+            "input directory": f"{THIS_DIR}/data/",
+            "out dir": f"{THIS_DIR}/results/",
+            "num processors": 0,
```
Waelthus, line 907: to have more control over this case, maybe just use 2 processors here (as 0 gets you half of what is available)

pro: guaranteed run in multiprocessing with known number of cpus (i.e. cannot accidentally be using 1 cpu when the system only has 2), e.g. same number will be used when test is run on a notebook vs github action, minimal possible demand in terms of number of cpus
con: figuring out the exact number to use would not be tested in this case (but that's not a huge issue)

I see that some test below is using 2 instead, I'd still consider using 0 here a potentially dangerous choice... We could give the if condition that sets up the right number of cpus a pragma to not be included in coverage if that's the reason for 0 here...

py/picca/tests/delta_extraction/data_tests.py
```diff
@@ -126,26 +545,726 @@ def test_data_filter_forests(self):
...
+        self.assertTrue(len(data.forests) == 63)
+
+        # run with 0 processors; case: main data present
+        config = ConfigParser()
+        config.read_dict({"data": {
+            "catalogue": f"{THIS_DIR}/data/QSO_cat_fuji_dark_healpix_with_main.fits.gz",
+            "keep surveys": "all",
+            "input directory": f"{THIS_DIR}/data/",
+            "out dir": f"{THIS_DIR}/results/",
+            "num processors": 0,
```
Waelthus, line 976: see above

py/picca/tests/delta_extraction/data_tests.py
```diff
@@ -126,26 +545,726 @@ def test_data_filter_forests(self):
...
+        self.assertTrue(len(data.forests) == 194)
+
+        # run with 0 processors
+        config = ConfigParser()
+        config.read_dict({"data": {
+            "catalogue": f"{THIS_DIR}/data/desi_mock_test_catalogue.fits",
+            "keep surveys": "all",
+            "input directory": f"{THIS_DIR}/data/",
+            "out dir": f"{THIS_DIR}/results/",
+            "num processors": 0,
```
Waelthus, line 1222: see above

py/picca/tests/delta_extraction/expected_flux_tests.py
```diff
@@ -381,69 +572,318 @@ def test_dr16_expected_flux_compute_var_stats(self):
...
-        # setup Forest variables; case: logarithmic wavelength solution
-        setup_forest("log", rebin=3)
+        out_file = f"{THIS_DIR}/results/var_stats_lin.txt"
+        test_file = f"{THIS_DIR}/data/var_stats_lin.txt"
 
         # initialize Data and Dr16ExpectedFlux instances
         config = ConfigParser()
         config.read_dict({
-            "data": sdss_data_kwargs,
+            "data": desi_healpix_kwargs,
```
Waelthus, line 612: wouldn't it make more sense to do this test on SDSS data (or on both)? Given that in the end for DESI ppl will probably use the lin mode more? But testing DESI with log is useful for cross checking results of lin vs log quickly...

Waelthus (approved):
To me those changes look fine, will take a look at a test run and at the coverage as well. I see that also the test tiles/pix are included now, so this merge will be a huge step forward wrt testing

Waelthus:
atm it looks like we're not testing exposure differences (which might be simple to include for DESI data by adding one run using "use non-coadded spectra" in a DESI-Pk1d mode), some eBOSS correction for resolution (only used for Pk1d, only for eBOSS), some masks (esp. BAL which iirc isn't used in the mainline analysis anymore), corrections (mostly some non-typical combinations of options that are not tested yet), eBOSS DRQ (checks for old data formats) and DESI tile mode (e.g. tests on non-cumulative tiles, but also Pk1d from tiles, non-coadded spectra...) have lower code coverage as well. I'd say all of those currently non-tested parts are ok to not be tested atm (regarding e.g. the DESI tile mode, if at all, we'd probably not want more testing, but ideally more unification of the tile/healpix modes => less duplicate code => tests covering both of them in most things)

py/picca/delta_extraction/astronomical_objects/forest.py
```diff
@@ -136,49 +136,60 @@ def __init__(self, **kwargs):
...
-        self.mean_snr = sum(snr) / float(len(snr))
-
         # call parent constructor
         super().__init__(**kwargs)
 
         self.consistency_check()
 
+        # compute mean quality variables
+        snr = self.flux * np.sqrt(self.ivar)
+        self.mean_snr = sum(snr) / float(len(snr))
```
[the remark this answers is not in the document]
iprafols: changed in a commit

py/picca/delta_extraction/astronomical_objects/forest.py
```diff
@@ -136,49 +136,60 @@ def __init__(self, **kwargs):
...
+                "Forest.set_class_variables"
+            )
         if cls.mask_fields is None:
+            raise AstronomicalObjectError(
+                "Error constructing Forest. Class variable "
+                "'mask_fields' must be set prior to initialize "
+                "instances of this type. This probably means you did not run "
+                "Forest.set_class_variables"
+            )
+        if not isinstance(cls.mask_fields, list):
```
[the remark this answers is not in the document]
iprafols: No, this is just a check that was missing

py/picca/delta_extraction/data.py
```diff
@@ -150,9 +148,11 @@ def __parse_config(self, config):
                 pixel_step_rest_frame = pixel_step
                 self.logger.info("'delta lambda rest frame' not set, using "
                                  f"the same value as for 'delta lambda' ({pixel_step_rest_frame})")
-        else:
-            raise DataError("Forest.wave_solution must be either "
-                            "'log' or 'lin'")
+        # this should not be reached as wave_solution is either "lin" or "log"
+        # added here only in case we add another wave_solution in the future
```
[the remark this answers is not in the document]
iprafols: No, the value of `wave_solution` here is the one used to initialize the `Forest` class variable. Initially, I thought to remove it (as you can see in the commit you're signaling) since I did not know how to have `coverall` ignore certain branches. Then I figured it would be better to have this check to mark this line if we decide to include something else in the future. Otherwise, we risk changing the `assert` above (line 127) and unknowingly add a bug.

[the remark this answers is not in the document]
iprafols: I can imagine a different mode if e.g., at some point, we decide to directly load data from multiple surveys

py/picca/tests/delta_extraction/data/bad_format/missing_z_color.fits
```diff
@@ -0,0 +1 @@
...                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                
```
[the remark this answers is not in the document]
iprafols: yes, this is a fits file. If you do inspect a fits file using the  `less`  command, then you always see the header as plain text and then you see the binary bits as unreadable characters. Here you seem not to see it as there is a very small array stored in this file

py/picca/tests/delta_extraction/data_tests.py
iprafols: missed adding the run with 2 processors in this particular test. Thanks for pointing it out. Added a commit to fix it.

py/picca/tests/delta_extraction/data_tests.py
iprafols: right after this is the equivalent test with `num processors = 2`

py/picca/tests/delta_extraction/data_tests.py
iprafols: right after this is the equivalent test with `num processors = 2`

iprafols:
Yes, there are still some tests missing. Apart from what you mention, there is also the problem of fits for eta, var_lss and fudge never converging on the tests we are performing. In general, corrections and masks are not very well tested. 

At some point we will need to add the remaining missing tests.

About the code duplication, I plan to add an automatic linting check (and do actual linting when uploading it), that is also why I think merging this branch now is useful, so that we can lint more "safely" (in particular when dealing with code duplication).

PR #879, opened 2022-04-11
Fix z resolution coadd

Waelthus:
This PR will allow coadding of resolution matrices of different shape (i.e. R- and Z- in current ohio-mocks which wouldn't run anymore without this fix).
To do so the reso matrix with less diagonals is padded with zeros on both sides to match the one with more diagonals.
@corentinravoux will do some tests.

### DIFF
--- a/py/picca/delta_extraction/data_catalogues/desi_healpix.py
+++ b/py/picca/delta_extraction/data_catalogues/desi_healpix.py
@@ -160,6 +160,8 @@ class DesiHealpix(DesiData):
             with context.Pool(processes=self.num_processors) as pool:
 
                 pool.starmap(self.read_file, arguments)
+            for key,forest in forests_by_targetid.items():
+                forest.consistency_check()
         else:
             forests_by_targetid = {}
             for (index,
@@ -309,7 +311,7 @@ class DesiHealpix(DesiData):
                     "z": row['Z'],
                 }
                 args["log_lambda"] = np.log10(spec['WAVELENGTH'])
-                
+
                 if self.analysis_type == "BAO 3D":
                     forest = DesiForest(**args)
                 elif self.analysis_type == "PK 1D":
--- a/py/picca/delta_extraction/data_catalogues/desi_healpix.py
+++ b/py/picca/delta_extraction/data_catalogues/desi_healpix.py
@@ -161,6 +161,7 @@ class DesiHealpix(DesiData):
 
                 pool.starmap(self.read_file, arguments)
             for key,forest in forests_by_targetid.items():
+                #TODO: the following just does the consistency checking again, to avoid mask_fields not being populated. In the long run an alternative way of running the multiprocessing is envisioned which would be more stable, see discussion in PRs 879 and 883
                 forest.consistency_check()
         else:
             forests_by_targetid = {}
--- a/py/picca/delta_extraction/data_catalogues/desisim_mocks.py
+++ b/py/picca/delta_extraction/data_catalogues/desisim_mocks.py
@@ -116,6 +116,7 @@ class DesisimMocks(DesiHealpix):
 
                 pool.starmap(self.read_file, arguments)
             for key,forest in forests_by_targetid.items():
+                #TODO: the following just does the consistency checking again, to avoid mask_fields not being populated. In the long run an alternative way of running the multiprocessing is envisioned which would be more stable, see discussion in PRs 879 and 883
                 forest.consistency_check()
         else:
             forests_by_targetid = {}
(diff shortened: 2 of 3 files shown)


py/picca/delta_extraction/data_catalogues/desi_healpix.py
```diff
@@ -160,6 +160,8 @@ def read_data(self):
             with context.Pool(processes=self.num_processors) as pool:
 
                 pool.starmap(self.read_file, arguments)
+            for key,forest in forests_by_targetid.items():
```
iprafols, line 163: This check is already done when the forest is constructed, why are we calling it again here? Is it to do the consistency check after coadding given that the resolution matrixes might change? If so, then I would suggest to add it in DesiPk1dForest.coadd and remove it from here and from DesisimMocks.

Or am I missing something?

py/picca/delta_extraction/data_catalogues/desisim_mocks.py
```diff
@@ -115,6 +115,8 @@ def read_data(self):
             with context.Pool(processes=self.num_processors) as pool:
 
                 pool.starmap(self.read_file, arguments)
+            for key,forest in forests_by_targetid.items():
```
iprafols, line 118: see my other comment

iprafols (changes requested):
Changes seem fine. However, there are some things I would like to clarify

py/picca/delta_extraction/data_catalogues/desi_healpix.py
Waelthus: If I understood Corentin correctly the issue is that some properties are not properly synchronized across multiple processes right now leading to errors downstream. This is fixed by the additional check. But I think we'll need to investigate this further and then fix this at the roots instead of readding those properties (also in case something else got lost in between due to non-synchronized processes)...

Waelthus: Without being an expert on multiprocessing either, I think this should be right. 
We're modifying the class instead of the instantiated objects that are passed back, so the parent process cannot know about those properties. 
@iprafols: Do you remember the reason for modifying this on the class level and not within the actual forest objects as @p-slash suggested? I'm not sure if this fix would help or cause other trouble...

An alternative way to fix this might be to have the `read_file` function actually return the output forests instead of appending them to the `forests_by_targetid` dict. Then the `pool.starmap` would return a list of the dicts generated from each file which we would need to merge (and e.g. coadd for duplicate targetids etc). This would be safer in a way because communication between processes working on the different files would be unnessessary (currently they need to talk to each other via the dict and are syncronized as the dict is created from a `multiprocessing.SyncManager`, but I don't know how well this works and if it guarantees synchronization before every read/write operation on the dict or the objects contained inside)

Waelthus:
It might be useful to merge this as a preliminary bugfix, but we'll need to come back to it later...

py/picca/delta_extraction/data_catalogues/desi_healpix.py
iprafols: The field `mask_fields` is intended to be the same for all instances of a particular run and is thus a class variable. 

A simple check would be to set Forest.mask_fields in `desi_healpix.read_data` before the forests are initialized (@p-slash, can you check if this works?). This would help us verify that this is indeed the problem. However, I do not envision this as a permanent solution as this is coupling Forest and DesiHealpix in a way that can be potentially troublesome in the future.

I think the solution @Waelthus suggests about returning the forests as opposed to appending them to the dictionary is the right path forward. It would make our life easier as we would not be modifying objects "behind the scene" and it will save us some of the race conditions. We could even run the co-adding in parallel if we keep the forests as lists as in 
```
if targetid in forests_by_targetid:
    forests_by_targetid[targetid].append(forest)
else:
    forests_by_targetid[targetid] = [forest]
```
We can then merge the lists from the different dictionaries in a single process (which should be super fast) and then co-add them using a parallelized function such as 
```
def coadd(forest_list):
    forest = forest_list[0]
    for forest2 in forest_list[1:]:
        forest.coadd(forest2)
    return forest
```
that would run for each `targetid`. 

How do you guys see this?

p-slash: Just a heads up. I wasn't able read the comments in detail (busy day), but I'll see if I can modify the code with your suggestions tomorrow

Waelthus: @iprafols: I'd say this would be a good step forward potentially making our lives way easier. No time to implement/test this right now, but if @p-slash wants to give it a go, I'd be happy to review.

Waelthus:
I would opt for merging this now as is and performing tests/modifications leading to a proper fix in a followup-PR to be opened today... Note that this PR originally wasn't even about fixing the multiprocessing, but about fixing I/O for current mocks in case of mismatches in the reso matrix shape...

iprafols (approved):

PR #880, opened 2022-04-19
fixed bugs in BAL masking

iprafols:
@LaurEnnesser this PR is to fix the bugs you reported on the BAL masking, can you make sure they are correct?


LaurEnnesser:
Yes, these all appear to be correct. Thanks!

LaurEnnesser (approved):

PR #883, opened 2022-04-28
Fix multiprocess

p-slash:
I wanted initiate this even though not fully tested. It also contains my requirement changes to make installation work in NERSC.
Basically two changes:
1) `Forest.mask_fields` is modified before reading files.
2) A small revamp of parallel reading using `imap` and an intermediate reader class. Each process reads and saves forest into dictionary. Then these dictionaries are merged one by one to the master dict.
Both these changes worked (ran without an error) for Pk1D + BalMask. 

Edit apr 29: I have some more changes, but don’t want to push since testing fails. Sorry didn’t know there was an automatic test!


iprafols:
@corentinravoux, given that this PR is still not ready for review (I think) I marked it as draft PR for the time being. If it's ready (or when it's ready) please press the ready for review and request a reviewer

Waelthus:
given that we'd want to merge to master and that no comments are currently here, I'm changing the base version to master for this PR

p-slash:
I performed the data_tests on NERSC and passed it. The problem seems to be with coveralls. Not sure why, but it could be due to authentication since this from a fork. Otherwise I think the code is ready. mask_fields are set in constructor of DesiHealpix, which is inherited by DesiMocks; and strictly speaking this fix happens after all the data is read.

Waelthus:
I re-ran the tests in case that was just a coveralls hickup, else I'll merge the fork to an igmhub branch and re-open the PR for merging that into master...

py/picca/delta_extraction/data_catalogues/desi_healpix.py
```diff
@@ -346,8 +207,181 @@ def read_file(self, filename, catalogue, forests_by_targetid):
...
+            filename = (
+                f"{input_directory}/{healpix//100}/{healpix}/{coadd_name}-{survey}-"
+                f"dark-{healpix}.fits")
+
+            arguments.append((filename,group))
+
+        self.logger.info(f"reading data from {len(arguments)} files")
+
+        if self.num_processors>1:
+            with multiprocessing.Pool(processes=self.num_processors) as pool:
```
Waelthus, line 363: I think the only part that really needs to be seperated for single/multi-core processing is the `imap_it=` vs the `reader=` line, the merging could be outside the if clause (minimizing code duplication)

Waelthus: i.e. maybe just do `imap_it=(reader(arg) for arg in arguments)` in the sequential version and have the merging for loop outside

Waelthus (approved):

Waelthus:
I did add a coulple comments here, but will merge this to an igmhub branch for tests now

py/picca/delta_extraction/data_catalogues/desi_healpix.py
```diff
@@ -346,8 +207,181 @@ def read_file(self, filename, catalogue, forests_by_targetid):
 
                 # keep the forest
                 if targetid in forests_by_targetid:
-                    existing_forest = forests_by_targetid[targetid]
-                    existing_forest.coadd(forest)
-                    forests_by_targetid[targetid] = existing_forest
+                    forests_by_targetid[targetid].coadd(forest)
```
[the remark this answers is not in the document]
p-slash: My understanding is that coadding function doesn't return a forest, but modifies the parent forest. That's how I read the initial piece of code. Is it not true?

py/picca/delta_extraction/data_catalogues/desi_healpix.py
p-slash: We can also not check for number of processors. It should behave like a serial code when `num_processors=1`. I don't know if that would create some weird behavior

Waelthus: the issue here is that it still would be run through multiprocessing stuff which makes debugging painful even if only on a single CPU, so having the serial case multiprocessing-free is just a convenience to allow debugging issues more easily

p-slash: Ok that makes sense. I'm worried about reducing the code duplication though. You can't take imap outside of `with`. For sequential version, keeping all in a list as you suggest might be memory intense. Let's postpone this to later testing if that's ok?

PR #884, opened 2022-05-03
Add pylint

iprafols:
This PR adds pylint as an automatic check (only for delta_extraction)
In order to pass the test, some basic linting is performed.  Before the PR is ready for review, some bits and pieces must be reorganized in order to avoid code duplication


iprafols:
Duplicated code is now removed. Overall I think the code is much simpler now. Before the merge is done, I need to update the data model with those changes, but this does not affect the code and it can be reviewed

py/picca/delta_extraction/data_catalogues/desi_healpix.py
```diff
@@ -4,25 +4,16 @@
...
 import fitsio
-import healpy
 import numpy as np
 
-from picca.delta_extraction.astronomical_objects.desi_forest import DesiForest
-from picca.delta_extraction.astronomical_objects.desi_pk1d_forest import DesiPk1dForest
-from picca.delta_extraction.astronomical_objects.forest import Forest
-from picca.delta_extraction.data_catalogues.desi_data import DesiData, defaults, accepted_options
+from picca.delta_extraction.data_catalogues.desi_data import DesiData
+from picca.delta_extraction.data_catalogues.desi_data import (# pylint: disable=unused-import
```
Waelthus, line 11: the split is just for line length? And the pragma is for pylint not to complain there because the import is multi-line?

py/picca/delta_extraction/data_catalogues/desi_healpix.py
```diff
@@ -124,77 +121,73 @@ def read_data(self):
...
 
-                input_directory = f'{self.input_directory}/{survey}/dark'
-                coadd_name = "spectra" if self.use_non_coadded_spectra else "coadd"
-                filename = (
-                    f"{input_directory}/{healpix//100}/{healpix}/{coadd_name}-{survey}-"
-                    f"dark-{healpix}.fits")
+                filename, is_mock_aux = self.get_filename(survey, healpix)
+                if is_mock_aux:
+                    is_mock = True
+                #input_directory = f'{self.input_directory}/{survey}/dark'
```
Waelthus, line 143: please remove comments if this is running well...

py/picca/delta_extraction/data_catalogues/desi_healpix.py
```diff
@@ -124,77 +121,73 @@ def read_data(self):
...
-                    f"{input_directory}/{healpix//100}/{healpix}/{coadd_name}-{survey}-"
-                    f"dark-{healpix}.fits")
-
-                #TODO: not sure if we want the dark survey to be hard coded in here, probably won't run on anything else, but still
-
+                filename = self.get_filename(survey, healpix)
+                filename, is_mock_aux = self.get_filename(survey, healpix)
+                if is_mock_aux:
+                    is_mock = True
+                #input_directory = f'{self.input_directory}/{survey}/dark'
```
Waelthus, line 174: see above

py/picca/delta_extraction/expected_fluxes/dr16_expected_flux.py
```diff
@@ -471,17 +484,32 @@ def compute_delta_stack(self, forests, stack_from_deltas=False):
...
+            kind="nearest",
+            fill_value="extrapolate")
         self.get_stack_delta_weights = interp1d(
             Forest.log_lambda_grid[stack_weight > 0.],
             stack_weight[stack_weight > 0.],
             kind="nearest",
             fill_value=0.0,
             bounds_error=False)
 
+    # TODO: We should check if we can directly compute the mean continuum
```
Waelthus, line 499: this would mean getting mean_cont instead of iterative refinement of previous mean cont estimate?

py/picca/delta_extraction/expected_fluxes/dr16_expected_flux.py
```diff
@@ -803,21 +829,23 @@ def chi2(eta, var_lss, fudge):
                                       fill_value="extrapolate",
                                       kind="nearest")
 
-    # pylint: dissable=no-self-use
+    # pylint: disable=no-self-use
     # We expect this function to be changed by some child classes
-    def get_continuum_model(self, forest, aq, bq, **kwargs):
+    def get_continuum_model(self, forest, zero_point, slope, **kwargs):
```
Waelthus, line 834: sounds like a good change

py/picca/delta_extraction/expected_fluxes/dr16_expected_flux.py
```diff
@@ -903,31 +932,27 @@ def populate_los_ids(self, forests):
...
 
             mean_expected_flux = forest.continuum * stack_delta
-            var_pipe = 1. / forest.ivar / mean_expected_flux**2
-            variance = eta * var_pipe + var_lss + fudge / var_pipe
-            weights = 1. / variance
-
+            weights = self.get_continuum_weights(forest, mean_expected_flux)
+            # this is needed as the weights from get_continuum_weights are
+            # divided by the continuum model squared, in this case mean_expected_flux
+            # TODO: check that we indeed need this or if the weights without it
```
Waelthus, line 941: Ok, so this does give the same result as before, right? It might be reasonable to have some kind of Fbar weights as an alternative, I guess... But this should be optional, not replacing the current setup

pylintrc
```diff
@@ -11,7 +11,7 @@ ignore=CVS
 
 # Add files or directories matching the regex patterns to the blacklist. The
```
Waelthus, line 12: should we maybe relax a little on e..g. line lengths? here and for yapf? sometimes it's just simpler to allow slightly longer lines and not have as much of trouble with the wrapped lines which can become annoying (e.g. for imports, comments, deeply nested loops)

Waelthus (approved):
Sounds good to me! Some comments attached.

Waelthus:
Please note that this will partially clash with #887, so might be good to have a look which merge order might be less troublesome

README.md
```diff
@@ -118,3 +118,15 @@ picca_export.py
...
+
+### For Developers
+Before submitting a PR please make sure to:
+1. Check the tutorials. Update them if necessary (typically the tutorial `picca_delta_extraction_configuration_tutorial` will need to be updated.
+2. Update the data model
+3. For every file you have modified run
+```yapf --style google file.py > aux
+mv aux file.py
+```
+to ensure the coding styles are maintained.
```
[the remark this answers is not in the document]
Waelthus: yes, might also be useful to not fully automatize this to ingest better coding standards into people...

py/picca/delta_extraction/data_catalogues/desi_healpix.py
iprafols: The split is because one is used in the module and the others are simply loaded (they are necessary imports but are not used inside the module). The pragma is for pylint to not complain about the fact the two imported variables are not used inside the module

py/picca/delta_extraction/data_catalogues/desi_healpix.py
iprafols: done

py/picca/delta_extraction/data_catalogues/desi_healpix.py
Waelthus: that's because they are only used for child classes? Or why would we elsewise import them here if they aren't used?

pylintrc
iprafols: If I recall properly, though, the values we are using are already a bit more relaxed than the default values for pylint

iprafols:
I'll review #887 before merging this one. I'll check locally which is the easiest merge order

py/picca/delta_extraction/data_catalogues/desi_healpix.py
iprafols: They are (or could be) used by child classes. In any case, they are also used by `Config` in order to enforce correct parameters in the config file and give early errors

py/picca/delta_extraction/expected_fluxes/dr16_expected_flux.py
Waelthus: While that would be nice, depending on the type of downstream analysis that can give unexpected effects. For Pk1d for instance, using the standard weights would be relatively simple to do when estimating the power spectrum from real space data, but when doing straight-forward FFTs each pixel of a spectrum will have the same weight in the output P(k), i.e. we cannot account for the different weighting of the pixels in a straightforward way.
This does not mean we couldn't in principle do the cont fitting with some weighting scheme, but effects this has need to be studied.
Agreed that within a run, i.e. after specifying which kind of weights one wants, things should be treated self-consistently.

iprafols: Maybe we should open an issue to discuss how to proceed with this

pylintrc
Waelthus: Ah, didn't read the whole pylintrc. We're using 100 chars, which is ok (the default is ~80 which is a little narrow often enough and going way larger leads to trouble when having 2 files next to each other e.g. in a diff). And indeed if I parse the regexp correctly we're allowing comments and web-addresses to go beyond the line limit (and could add other stuff to there if we figure out there's issues).

Is yapf using the same setup that the linter uses? Or will that produce the default settings or has a seperate config? Iirc there was some way to unify all of this, but I don't remember right now how it worked...

py/picca/delta_extraction/expected_fluxes/dr16_expected_flux.py
Waelthus: I think it's fine as long as we keep the current major modes (i.e. standard or constant weights, not sure if anyone uses ivar-only weights) in working shape which allows us to do analyses similar to previous works.

We could have a discussion if other weightings might be useful, then implement those, and then if they have been tested and lead to improvement I'd be fine with dropping previous modes. In principle, already the change in binning from log to lin can be seen as a λ-dependent change in weighting, at least for e.g. constant weights.

pylintrc
iprafols: I guess yapf is using google coding standards with their defaults (as I have not created a config file for that)

Waelthus: ok, let's just keep it like that for the moment.

PR #888, opened 2022-05-06
Fix multiprocessing from linted code

iprafols:
This PR includes the changes in branch fix-multiprocessing with the changes of branch add-pylint. Since the merge conflicts were bothersome, I added the changes manually (thus the new independent branch).


iprafols:
This PR should replace #871

iprafols:
data model needs to be updated before the merge

py/picca/delta_extraction/data_catalogues/desi_data.py
```diff
@@ -32,6 +32,30 @@
...
+    """Merge forests_by_pe and forests_by_targetid as Forest instances.
+
+    Arguments
+    ---------
+    forests_by_targetid: dict
+    Dictionary were forests are stored. Its content is modified by this
+    function with the new forests.
+
+    forests_by_pe: str
+    Name of the file to read
```
p-slash, line 45: docstring error: forests_by_pe is a dictionary.

py/picca/delta_extraction/data_catalogues/desi_data.py
```diff
@@ -139,20 +278,21 @@ def format_data(self, catalogue, spectrographs_data, targetid_spec,
...
-
         reso_from_truth: bool - Default: False
         Specifies whether resolution matrixes are read from truth files (True)
         or directly from data (False)
 
         Return
         ------
+        forests_by_targetid: dict
+        Dictionary were forests are stored. Its content is modified by this
+        function with the new forests.
```
p-slash, line 289: docstring forests_by_targetid. The contents are not modified I guess, it just returns. I am also assuming format_data function is just a copy of the old code.

py/picca/delta_extraction/data_catalogues/desi_data.py
iprafols: Well spotted, addressed in new commit.

py/picca/delta_extraction/data_catalogues/desi_data.py
iprafols: You're right, fixed it in a commit. 
Regarding the second part of you comment yes and no. It is part of the old code but not just a copy. It was also part of `read_file` method in `DesiTile` so it was convenient to split it to avoid the duplication

py/picca/delta_extraction/data_catalogues/desi_healpix.py
```diff
@@ -125,58 +129,69 @@ def read_data(self):
 
         is_sv = True
         is_mock = False
+        forests_by_targetid = {}
+
+        arguments = []
+        for (healpix, survey), group in zip(grouped_catalogue.groups.keys,
```
p-slash, line 135: We don't need to zip grouped_catalogue, but if that's what you prefer that's fine. Not sure if there is an overhead with zipping.

py/picca/delta_extraction/data_catalogues/desi_healpix.py
```diff
@@ -125,58 +129,69 @@ def read_data(self):
...
-                    is_sv = False
-
-                filename, is_mock_aux = self.get_filename(survey, healpix)
-                if is_mock_aux:
-                    is_mock = True
-                arguments.append((filename, group, forests_by_targetid))
-
-                self.logger.info(f"reading data from {len(arguments)} files")
             with context.Pool(processes=self.num_processors) as pool:
+                imap_it = pool.imap(
```
p-slash, line 151: I also recently changed this to imap_unordered. Shouldn't be a big deal either way

p-slash (commented):
I didn't catch a mistake. It looks nice and clean! You removed read_file from DesiHealpix, so just make sure it is not called in test scripts. I would also keep some of my comments regarding imap and multiprocessing (and logger problem), so that someone does not trip over same things.

py/picca/delta_extraction/data_catalogues/desi_healpix.py
iprafols: How do imap and imap_unordered differ? Is it a matter of speed? If it's not significant I'd rather keep the ordered version to ensure that it runs the same way in different machines. This would make things simpler to debug. For example, it took me a while to figure out that the delta_extraction code was not giving me the exact same results because the spectra were ordered in a different way and the DLA masking had some random numbers involved, meaning that the order mattered

p-slash: I think the difference is imap preserves the order of the arguments, and imap_unordered doesn't, which might speed up a bit. But yours is a valid reason to keep it ordered. There might be subtleties when data gets bigger, but no need to prematurely worry about it.

py/picca/delta_extraction/data_catalogues/desi_healpix.py
iprafols: I added them with the `zip` as I didn't like accessing the first element in the group to get them, but then there will always be at least an element on each group (otherwise the `(healpix, survey)` pair does now appear). I changed it back to how you did it in your version.

iprafols:
Can you include these comments via a commit? After this, fi you're happy that I properly addressed the few issues you raised, then accept the revision and I'll merge this PR with master and close the other one.

p-slash (approved):

PR #889, opened 2022-05-16
parallelisation added to DesiTile

iprafols:
We could read DESI data in parallel with DesiHealpix but this was missing for DesiTile. This PR fix this


py/picca/delta_extraction/data_catalogues/desi_tile.py
```diff
@@ -229,9 +245,14 @@ def read_file(self, filename, catalogue):
 
         hdul.close()
 
-        select = ((catalogue['TILEID'] == tile_spec) &
-                  (catalogue['PETAL_LOC'] == petal_spec) &
-                  (catalogue['NIGHT'] == night_spec))
+        if "cumulative" in self.input_directory:
+            select = ((catalogue['TILEID'] == tile_spec) &
+                      (catalogue['PETAL_LOC'] == petal_spec) &
+                      (catalogue['LASTNIGHT'] == night_spec))
```
Waelthus, line 251: wasn't it LAST_NIGHT? Just from memory, didn't check...

iprafols: No, it's LASTNIGHT, I just run it on NERSC and it works

PR #890, opened 2022-05-17
Fix true continuum p1d

corentinravoux:
A typo and a fix to run true continuum with constant weights in p1d mode.


py/picca/delta_extraction/expected_fluxes/true_continuum.py
```diff
@@ -293,11 +293,6 @@ def read_raw_statistics(self):
         # files are only for lya so far, this will need to be updated so that
         # regions other than Lya are available
 
-        if self.use_constant_weight:
-            logger.info(
```
iprafols, line 297: Added a commit removing an unused import (caused by the removal of this line)

iprafols (commented):
Not sure why the file `bin/picca_convert_transmission.py` appears to be empty. Can you double-check that this is not an error?

corentinravoux:
I think it is only the diff which is empty. It was not in a +x mod, I just changed it.

iprafols:
I see, then I'm fine with merging

iprafols (approved):

PR #891, opened 2022-05-24
Fixed extension name for quasar catalogue

iprafols:
Minor fix so that we can work with different versions of the DESI catalogs. We will probably need to change this once we have settled on a naming of the catalogue HDUs


Waelthus:
can we use this PR as well to fix issues regarding DLA cats, i.e. this question by @p-slash: https://desisurvey.slack.com/archives/CTJ0GRSR3/p1653503871639789 and @corentinravoux's answer

iprafols:
I have added a commit to address also the DLA catalog issues. Now it reads Z_DLA or Z if Z_DLA is not present

Waelthus (approved):

PR #893, opened 2022-06-01
Fix continuum fitting tests

iprafols:
This PR fixes an issue with the continuum fitting tests, in particular for the functions eta, var_lss, and fudge. The current way fits work in the master branch, we require more than 100 quasars to be in a given bin in `log_lambda` *and* `var_pipe`. This causes all the fits in the automatic tests to fail making it impossible to automatically check for changes in the fitting behaviour. 

To fix this, I changed the requirement to be at least 1 quasar. The tests should be passing as I updated the test suite accordingly. I was expecting there not to be any change when running with large sets of spectra but apparently there are minor changes (due to the distribution in pipeline variance). However, based on an analysis of 50k DESI spectra, I think that the results with the modified code are superior to the master branch.


.github/workflows/pylint.yml
```diff
@@ -18,7 +18,7 @@ jobs:
         sudo apt-get -y install libbz2-dev
         python -m pip install --upgrade pip
         if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
-        pip install pylint
+        pip install pylint==2.12.2
```
Waelthus, line 21: why do we need to fix the pylint version? Is it already to prevent installing early 3.0 versions (that are partly backwards incompatible I guess), why then not fix it to 2.14.0

iprafols: there seems to be a change in the way new pylint version works that made it fail the test on the commit `update second half of the test suite`

py/picca/delta_extraction/expected_fluxes/dr16_expected_flux.py
```diff
@@ -760,8 +761,8 @@ def chi2(eta, var_lss, fudge):
                     variance)
                 weights = var2_delta[index * num_var_bins:(index + 1) *
                                      num_var_bins]
-                w = num_qso[index * num_var_bins:(index + 1) *
-                            num_var_bins] > 100
+                w = count_qso[index * num_var_bins:(index + 1) *
+                            num_var_bins] > 0
```
Waelthus, line 765: can we make the min_num_qso a variable of the cont fitter? So that previous results are still reproducable if the variable is set to 100. And tests will still give the same result as before if 100 is used.

Can we then have the actual change to 0 in a seperate PR that does only this and a change of tests? I think that would make clearer which exact change produced changes in test results.

Waelthus:
What I also don't get is why tests fail even with all expected result files having been changed

iprafols:
Yes, that is weird, they do not fail locally.

py/picca/delta_extraction/expected_fluxes/dr16_expected_flux.py
iprafols: sure, I'll do another PR with changing it to a variable and then will update this one with the change to 0

.github/workflows/pylint.yml
Waelthus: it might be that they changed that version to 3.0 alpha (at least they mention on their github that all incompatible changes were put in 3.0a), but that's anyway not the point of this PR, so can be adapted later

py/picca/delta_extraction/expected_fluxes/dr16_expected_flux.py
Waelthus: I merged that PR to master and here, what is missing is only the new default value or value in the test config...

Waelthus:
This commit should now only adapt tests and log statements (i.e. qso counting), so I'm fine with it

Waelthus (approved):

iprafols:
Addressed by PR #897

PR #894, opened 2022-06-02
Var stats fitter change

iprafols:
This PR basically converts the minimum number of quasars participating in a bin of wavelength and pipeline variance for it to be considered in the fit of eta, var_lss and fudge


iprafols:
This is the intermediate step to PR #893 that @Waelthus  asked for

Waelthus (approved):

PR #897, opened 2022-06-03
Fix continuum fitting tests, attempt2 

iprafols:
A fresh attempt at PR #893 after the recent commits to the master branch. This **only** updates the value of `min num qso in fit` to 0. The rest of the changes are just updates on the test suite due to this change


iprafols:
ok so it seems we have the same issue as before, the tests pass locally on my laptop but not here. @Waelthus, can you check if they also fail on yours as with the other PR?

Waelthus:
they fail: logs attached
[test.log](https://github.com/igmhub/picca/files/8859497/test.log)
[test.out.log](https://github.com/igmhub/picca/files/8859500/test.out.log)

Waelthus:
Maybe the right approach to fix the tests would be to run them on the github action system, store the results from there and replace the tests by those (then they would fail on your notebook, but should work on the auto tests)

iprafols:
There is something weird going on here... I'm not sure we should disable the multiprocessing (some of the tests are precisely testing that). There must be something going on that affects the results. Maybe an ordering problem?

Waelthus:
yes, I know we did multiprocessing tests to test the multiprocessing analysis. But multiprocessing is also a way to change e.g. order of execution, so if that plays a role for whatever reason we'd have more stable results with single core analyses.
And then we might want to add a multiprocessed test but with relaxed requirements or doing multiprocessing of specific parts seperately to figure out what exactly is going on (i.e. if the issue comes from i/o or cont fitting)

iprafols:
I found an issue in the computation of the quasar continua when run multiple cores. I believe this is what is causing this problem. I opened PR #902  to solve this. To keep things clean, I will close this PR and make another attempt at fixing the tests once the PR #902 is merged into master

PR #899, opened 2022-06-13
fixed installation instructions

iprafols:
Removes the `--user` from the picca installation instructions. Use `pip install -e .` instead

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,19 +13,21 @@ This package contains tools used for the analysis of the Lyman-alpha forest samp
 The current reference is du Mas des Bourboux et al. 2020 (https://arxiv.org/abs/2007.08995).
 
 ## Installation
-
-download
-```bash
+We recommend to create a clean environment:
+```
+conda create -n my_picca_env python==version
+conda activate my_picca_env
+```
+then download
+```
 git clone https://github.com/igmhub/picca.git
 ```
-
 add to your bashrc
-```bash
+```
 export PICCA_BASE=<path to your picca>
 ```
-
 and finally run
-```bash
+```
 pip install -e .
 ```
 
--- a/README.md
+++ b/README.md
@@ -24,16 +24,10 @@ add to your bashrc
 export PICCA_BASE=<path to your picca>
 ```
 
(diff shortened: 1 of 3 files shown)


Waelthus:
maybe add that we suggest installing picca in a clean environment, e.g. by 
```
conda create -n my_picca_env python==version
conda activate my_picca_env
```

iprafols:
added conda instructions as suggested by @Waelthus

andreicuceu:
@iprafols I think this looks good. Let's have a chat about the requirements files at the telecon today. I think it might be useful to restructure them and add a sentence in the readme on what parts of the code need them. This may be outside the scope of this pull request, but I think at least with the main `requirements.txt` file, we should say something in the readme on how/when to use it.

Waelthus:
Note that currently we rely on the req-file for the auto-tests as well. I guess this can all be changed, but at least for me that's the main reason to have it lying around (in addition to having documentation regarding the versions used in tests)

andreicuceu:
I'm not sure that has to be changed because I don't think we should get rid of the `requirements.txt` file. We need to find a balance between ease of use and reproducibility. For the vast majority of use cases, we don't need to specify exact package versions (ie calling only `pip install -e .` is fine). However, we should have well-defined environments when running the main DESI analyses. For example, when producing the final deltas/correlations/fits for Y1, we need to make sure the exact versions are saved. This could be done with the `requirements.txt` file, or maybe in some other way. I'm open to ideas.

Waelthus:
Oh, yes. I did not see that we still recommend using the req-file at installation. I personally never used the file during install except if I wanted to pin down some version...
All dependencies are downloaded and installed automatically anyway.
So personally I'd say we can remove that line from the instructions.

andreicuceu:
@iprafols and @Waelthus, I updated the instructions and removed the superfluous requirements files. Please have a look, and let me know if I missed something, or you would want to add/modify something.

andreicuceu (approved):

iprafols:
looks fine to me

Waelthus (approved):

PR #900, opened 2022-06-15
Optimize forest

iprafols:
This PR adds numba versions for functions find_bins and rebin


iprafols:
See the speed report:

iprafols:
We should probably do some clean-up before we merge, but for now I leave the extra functions in so the we can rerun the speed tests if necessary

py/picca/delta_extraction/utils.py
```diff
@@ -144,6 +148,95 @@ def find_bins(original_array, grid_array, wave_solution):
...
+        grid_array = 10**grid_array
+    else:  # pragma: no cover
+        raise DeltaExtractionError(
+            "Error in function find_bins from py/picca/delta_extraction/utils.py"
+            "expected wavelength solution to be either 'log' or 'lin'. ")
+    original_array_size = original_array.size
+    grid_array_size = grid_array.size
+    found_bin = np.zeros(original_array_size, dtype=np.int16)
+    for index1 in range(original_array_size):
+        min_dist = 1000 * (grid_array[1] - grid_array[0])
```
Waelthus, line 170: having 1000 pixels as the initial min_dist seems arbitrary, why not just set it to np.inf at the beginning and then whatever the first value will be gets set...
Also: I'm pretty sure there's a faster way to do this closest bin search, probably something like bisecting the grid_array until only the closest candidate is there instead of looping through all of them.
If we can assume that both arrays are monotonic in lambda we could also break the inner loop for any dist>min_dist if we're starting with min_dist=np.inf, saving ~50% or so of the calls on average.

py/picca/tests/delta_extraction/speed_tests.py
```diff
@@ -0,0 +1,155 @@
+"""This module contains speed tests that will be removed as deprecated
+functions are erased"""
+from time import time
+import timeit
+
```
Waelthus, line 5: I guess this file will only stay inside the repo for this PR and be removed afterwards together with the slow options for find_bins/rebin

Waelthus (changes requested):
Looks great as a start. I guess you can have additional speed gain when changing the logic of the numba code slightly.
Also not sure why certain dtypes have been chosen here...

py/picca/delta_extraction/data_catalogues/sdss_data.py
```diff
@@ -136,9 +136,10 @@ def read_from_spec(self, catalogue):
                 continue
             self.logger.progress(f"Read {filename}")
 
-            log_lambda = hdul[1]["loglam"][:]
-            flux = hdul[1]["flux"][:]
-            ivar = hdul[1]["ivar"][:] * (hdul[1]["and_mask"][:] == 0)
+            log_lambda = np.array(hdul[1]["loglam"][:], dtype=np.float32)
```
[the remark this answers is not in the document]
iprafols: If I don't add it I get this error: `numba.core.errors.NumbaValueError: Unsupported array dtype: >f4`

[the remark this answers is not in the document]
iprafols: I'll change it to np.float64

py/picca/delta_extraction/utils.py
iprafols: I tried using `np.finfo(np.float64).max` but there was a numba error as some ob the objects were np.float32. Now it seems to work so I'll change it back. I also added a `else: break` to save some of the loops. I'm not sure it's trivial to add the bisecting given the non monotonically behaviour of the `abs` function

py/picca/tests/delta_extraction/speed_tests.py
iprafols: I was planning on removing it before merging, yes. Just added in case you wanted to do some tests

py/picca/delta_extraction/utils.py
```diff
@@ -144,6 +148,95 @@ def find_bins(original_array, grid_array, wave_solution):
...
+        for index2 in range(grid_array_size):
+            dist = np.abs(grid_array[index2] - original_array[index1])
+            if dist < min_dist:
+                min_dist = dist
+                found_bin[index1] = index2
+    return found_bin
+
+
+@njit(parallel=True)
```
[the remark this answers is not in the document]
Waelthus: maybe one test run on nersc?

py/picca/delta_extraction/utils.py
Waelthus: I think this can still be done, we still know that in this case we have monotonic decrease followed by monotonic increase, just needs a little more syntax... we can use the current approach for now and potentially update later

iprafols: Yes let's do that. I think that with these changes the bottleneck will be elsewhere. Probably in function `merge_new_forest` (but still have not run the profiling)

Waelthus (approved):
please remove unneeded stuff, else this is fine with me atm. Should have some possibility to further speedup

py/picca/delta_extraction/utils.py
iprafols: Here are the speed test results at NERSC: 
```
#####################################
# Speed tests on function find_bins #
#####################################
find_bins
   Compilation: 0.868795s
   1000 calls: 0.000711s/call
   10000 calls: 0.000711s/call
   100000 calls: 0.000712s/call
find_bins_nonumba
   Compilation: 0.000065s
   1000 calls: 0.003837s/call
   10000 calls: 0.003840s/call
   100000 calls: 0.003839s/call
find_bins_numbaparallel
   Compilation: 0.905982s
   1000 calls: 0.000688s/call
   10000 calls: 0.000615s/call
   100000 calls: 0.000619s/call
##########################################
# Speed tests on function rebin (Forest) #
##########################################
Note: all runs here use the numbaized version of find_bins
Note: all runs here initialize a forest and call the rebin function
Rebin with numba
   1 call + Compilation: 4.373353s
   1000 calls: 0.000278s/call
   10000 calls: 0.000274s/call
   100000 calls: 0.000273s/call
Rebin without numba
   1 call + Compilation: 0.000681s
   1000 calls: 0.000547s/call
   10000 calls: 0.000545s/call
   100000 calls: 0.000545s/call
##############################################
# Speed tests on function rebin (DesiForest) #
##############################################
Note: all runs here use the numbaized version of find_bins
Note: all runs here initialize a forest and call the rebin function
Rebin with numba
   1 call + Compilation: 0.000380s
   1000 calls: 0.000277s/call
   10000 calls: 0.000277s/call
   100000 calls: 0.000277s/call
Rebin without numba
   1 call + Compilation: 0.000600s
   1000 calls: 0.000551s/call
   10000 calls: 0.000551s/call
   100000 calls: 0.000551s/call
```

iprafols: and with OMP_NUM_THREADS=16: 
```
#####################################
# Speed tests on function find_bins #
#####################################
find_bins
   Compilation: 2.941702s
   1000 calls: 0.000723s/call
   10000 calls: 0.000723s/call
   100000 calls: 0.000723s/call
find_bins_nonumba
   Compilation: 0.000060s
   1000 calls: 0.003933s/call
   10000 calls: 0.003919s/call
   100000 calls: 0.003903s/call
find_bins_numbaparallel
   Compilation: 0.959301s
   1000 calls: 0.000670s/call
   10000 calls: 0.000626s/call
   100000 calls: 0.000624s/call
##########################################
# Speed tests on function rebin (Forest) #
##########################################
Note: all runs here use the numbaized version of find_bins
Note: all runs here initialize a forest and call the rebin function
Rebin with numba
   1 call + Compilation: 4.269881s
   1000 calls: 0.000273s/call
   10000 calls: 0.000271s/call
   100000 calls: 0.000270s/call
Rebin without numba
   1 call + Compilation: 0.000673s
   1000 calls: 0.000533s/call
   10000 calls: 0.000537s/call
   100000 calls: 0.000535s/call
##############################################
# Speed tests on function rebin (DesiForest) #
##############################################
Note: all runs here use the numbaized version of find_bins
Note: all runs here initialize a forest and call the rebin function
Rebin with numba
   1 call + Compilation: 0.000359s
   1000 calls: 0.000276s/call
   10000 calls: 0.000276s/call
   100000 calls: 0.000276s/call
Rebin without numba
   1 call + Compilation: 0.000584s
   1000 calls: 0.000547s/call
   10000 calls: 0.000540s/call
   100000 calls: 0.000542s/call
```

PR #902, opened 2022-06-24
Fixed parallelisation problem on compute continuum

iprafols:
This PR fixes a problem found on compute continuum where the different processors reset the continuum fitting parameters, leading to unstable and incorrect behavior. 

This also includes a minor change in the configuration so that `num processors` can be set independently for the data reading and for the expected flux computation. Note that this should not make its way to the tutorials as this is a feature intended for debug purposes only


andreicuceu:
Parallelization looks safe to me now.

andreicuceu (approved):

PR #907, opened 2022-06-28
Faster true cont

p-slash:
True continuum analysis reads files one healpix at a time, which should reduce io friction and speed up delta_extraction

### DIFF
--- a/py/picca/delta_extraction/expected_fluxes/true_continuum.py
+++ b/py/picca/delta_extraction/expected_fluxes/true_continuum.py
@@ -24,6 +24,7 @@ defaults = {
     "raw statistics file": "",
     "use constant weight": False,
 }
+IN_NSIDE = 16
 
 
 class TrueContinuum(ExpectedFlux):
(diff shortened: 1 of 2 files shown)


py/picca/delta_extraction/expected_fluxes/true_continuum.py
```diff
@@ -152,13 +152,30 @@ def compute_expected_flux(self, forests):
         -----
         ExpectedFluxError if Forest.wave_solution is not 'lin' or 'log'
         """
-        context = multiprocessing.get_context('fork')
-        pool = context.Pool(processes=self.num_processors)
+        in_nside = 16
+        healpixes = np.array([healpy.ang2pix(in_nside,
```
iprafols, line 156: This is also computed in function `read_true_continuum` so maybe we don't need to compute it twice. How about we compute it only here and pass the healpix as an argument there?

py/picca/delta_extraction/expected_fluxes/true_continuum.py
```diff
@@ -152,13 +152,30 @@ def compute_expected_flux(self, forests):
         -----
         ExpectedFluxError if Forest.wave_solution is not 'lin' or 'log'
         """
-        context = multiprocessing.get_context('fork')
-        pool = context.Pool(processes=self.num_processors)
+        in_nside = 16
```
iprafols, line 155: I'm thinking we should hard-code `in_nside` at the beginning of the module so that it is made more clear. I would add
`IN_NSIDE = 16` at line 27 and then here do `in_nside = IN_NSIDE` (or use `IN_NSIDE` directly)

py/picca/tests/delta_extraction/expected_flux_tests.py
```diff
@@ -1667,7 +1667,7 @@ def test_true_continuum_populate_los_ids(self):
 
         # compute the forest continua
         for forest in data.forests:
-            expected_flux.read_true_continuum(forest)
+            expected_flux.read_true_continuum([forest])
```
iprafols, line 1670: can we add a test for reading a group of forests with the same healpix? The way we now do it on the code?

iprafols (changes requested):

py/picca/delta_extraction/expected_fluxes/true_continuum.py
p-slash: Done

py/picca/tests/delta_extraction/expected_flux_tests.py
p-slash: Not sure if I did it what you wanted here. I put all "reading a group of forests" lines into another function. Now I call that function in tests.

Waelthus:
Nice fix to an oversight of the earlier parallelization changes. If I read the discussion right, nothing really to do here anymore except for maybe test changes... Should we merge @iprafols?

iprafols (approved):

PR #908, opened 2022-06-28
Update bal_mask.py

p-slash:
Adding Si IV lines to mask in BAL quasars. This affects the first side band region, i.e. not Lya forest.


iprafols (commented):
Looks fine, but really @LaurEnnesser should confirm as she is the expert on BAL masking

LaurEnnesser (approved):

PR #911, opened 2022-07-01
Parallel delta saving

p-slash:
One more... This parallelizes the delta saving.


py/picca/delta_extraction/data.py
```diff
@@ -400,34 +442,31 @@ def find_nside(self):
         for forest, healpix in zip(self.forests, healpixs):
             forest.healpix = healpix
 
-    def save_deltas(self):
+    def save_deltas(self, num_processors=1):
```
Waelthus, line 445: can't we actually use `self.num_processors` inside the function? We'd just need to add a `self.num_processors=1` to the `__init__` here (as DESI data has it and SDSS does not)

Waelthus (approved):
looks good, maybe change how the num_processors is passed, but fine elsewise (if it works)

py/picca/delta_extraction/data.py
iprafols: I agree, I'd say we move the definition of `self.num_processors` from `DesiData` to `Data`. This way, we will be implementing the parallelization of the saving also for `SdssData`. This change will not otherwise affect `SdssData` as this variable would not be used elsewhere

py/picca/delta_extraction/survey.py
```diff
@@ -258,6 +258,6 @@ def save_deltas(self):
         """Save the deltas."""
         t0 = time.time()
         self.logger.info("Saving deltas")
-        self.data.save_deltas()
+        self.data.save_deltas(self.num_processors)
```
iprafols, line 261: with the change suggested above, we would not need this change here

py/picca/delta_extraction/data.py
iprafols: We will need to update the data model, though

iprafols (changes requested):

py/picca/delta_extraction/data.py
p-slash: I tried it, but it got very complicated. This option is entangled in configuration settings. Maybe just merge this, and we figure out how to move design later?

iprafols: I fixed it in PR #921. That PR is merging to this branch so it should be resolved first

p-slash:
There seems to be conflicts with master. I'm not familiar with data_model pictures. Who can help fix these problems? @Waelthus @iprafols

Waelthus:
I guess this is because some changes to the data model draws were done in another (now merged) PR while this was open... So I'll merge this back into here and then re-add the changes from here

Waelthus:
ok, the data model stuff is fixed. please check if anything got modified in an unwanted way during the merge (esp data.py/desi_data.py)

Waelthus (approved):
from my perspective this should be good to go, automatic tests seem to be fine as well, maybe give it a try in an actual parallelized run to make sure

iprafols (approved):

PR #913, opened 2022-07-22
Optional stack delta correction

p-slash:
These changes add an option for `Dr16ExpectedFlux` that controls stack delta normalization in continuum fitting. The option is named `force stack delta to zero` and its default value (`True`) preserves the current behavior of continuum fitting. This flexibility will enable us to decouple and test calibration and numerical artifacts.


iprafols (approved):
Changes look fine. I'm happy to merge this as long as the automatic tests pass

PR #914, opened 2022-07-25
Fixed defaults and accepted options

iprafols:
Minor fix of the default values and accepted options. Note that this was not picked up by the automatic tests so some more tests are required

### DIFF
--- a/py/picca/delta_extraction/masks/absorber_mask.py
+++ b/py/picca/delta_extraction/masks/absorber_mask.py
@@ -11,10 +11,10 @@ from picca.delta_extraction.mask import Mask
 
 defaults = {
     "absorber mask width": 2.5,
+    "los_id name": "THING_ID",
 }
 
-accepted_options = ["absorber mask width", "filename"]
-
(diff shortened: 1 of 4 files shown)


py/picca/delta_extraction/masks/absorber_mask.py
```diff
@@ -11,10 +11,10 @@
 
 defaults = {
     "absorber mask width": 2.5,
+    "los_id name": "THING_ID",
```
Waelthus, line 14: this is mainly to allow additional flexibility later when having other surveys/cats? Instead of guessing through all known values for this? Where was this set before if we didn't allow it as an option and didn't have a default value?

Waelthus (approved):

py/picca/delta_extraction/masks/absorber_mask.py
iprafols: This was not working before. I called it a minor fix as it is somehow still not working (though this step is required to make this class work). When I run delta_extraction using this mask there seems to be no effect. I added an issue to track it, but I plan to address it after the holidays

PR #915, opened 2022-07-25
Added extra log messages to monitor progress

iprafols:
This PR adds some extra logging info reporting times spent in each of the steps of the Dr16ExpectedFlux computation. The number of processors used is also reported. This will useful for debugging purposes


Waelthus (approved):
Looks ok to me. Having the timings in the logs definitely makes sense.

If we'd want an actual progress bar, we could have a look at tqdm and similar things as well (but not sure how well those would work in multiprocessing or on logfiles instead of an interactive stdout)...

PR #916, opened 2022-07-25
method compute_continuum moved to function

iprafols:
The changes here take method `compute_continuum` from `Dr16ExpectedFlux` and convert it to a function outside the class. This is done as this function can be called in parallel and separating it from the class ensures that there are no race problems. Thus, it will prevent future bugs.

In order to achieve this, I had to reorder a bit some of the functions here and there. Still, I believe this makes class `Dr16ExpectedFlux` simpler while maintaining a reasonable simplicity in the other involved classes.


andreicuceu (approved):
Parallelization looks safe now. The lone compute_continuum function may be better off in a separate file (to keep with the paradigm of one class per file), but it's not a big deal if it stays there either.

iprafols:
@andreicuceu  I moved the function to a separate file (`delta_extraction/expected_fluxes/utils.py`)

andreicuceu (approved):

PR #919, opened 2022-08-01
Keep masked pixels

p-slash:
This branch enables to keep all pixels in delta files. Masked pixels are noted with ivar=0 and weight=0. These changes are substantial. It is getting really hard for me to keep these changes separate. I did a huge merge fix for the recent changes. Let's merge them before the code gets rearranged again.


py/picca/delta_extraction/astronomical_objects/desi_pk1d_forest.py
```diff
@@ -197,22 +197,33 @@ def rebin(self):
...
+        # Find non-empty bins
+        binned_arr_size = bins.max() + 1
+        bincounts = np.bincount(bins, minlength=binned_arr_size)
+        wnonempty_bins = bincounts != 0
+
+        # Do a simple average when ivar=0
+        orig_ivar_2 = orig_ivar[w1]
+        w__ = orig_ivar_2>0
+        orig_ivar_2[~w__] = 1
+        rebin_reso_ivar = np.bincount(bins, weights=orig_ivar_2, minlength=binned_arr_size)
```
Waelthus, line 216: so this gives masked pixels a weight of 1 in the rebinned reso matrix whereas for the non-masked pixels ivar weighting is used, leading to potentially lower weight than for a masked pixel (i.e. whenever ivar<1)?

Waelthus: I guess the whole resolution matrix rebinning is anyway nonsense when actually rebinning pixels, as something would need to happen in the other dimension (i.e. the position difference wrt the central pixel) of the resolution matrix as well, so it's not really an issue, but just wanted to point this out
I guess in principle we'd also want to somehow flag the resolution matrix elements that are affected by a masked pixel in any way, i.e. in both the central pixel dimension or the difference dimension, but that might not be worth the effort

py/picca/delta_extraction/astronomical_objects/pk1d_forest.py
```diff
@@ -238,44 +238,52 @@ def rebin(self):
...
-        self.reso = rebin_reso[w2] / rebin_ivar[w2]
-        self.reso_pix = rebin_reso_pix[w2] / rebin_ivar[w2]
+        self.exposures_diff[w2] = rebin_exposures_diff[w2_] / rebin_ivar[w2_]
+        self.reso[w2] = rebin_reso[w2_] / rebin_ivar[w2_]
+        self.reso_pix[w2] = rebin_reso_pix[w2_] / rebin_ivar[w2_]
 
         # finally update control variables
-        self.mean_reso = self.reso.mean()
+        self.mean_reso = self.reso[w2].mean()
         self.mean_z = (
```
Waelthus, line 278: I guess here we'd need to decide if we want to use the mean redshift of pixels that actually make it into the masked spectrum or the central redshift of the spectrum before masking (which is the behaviour now and has also been previously except if the edge of the spectrum has been masked)

Waelthus (commented):
Looks good to me overall, some comments above regarding some design decisions.

Please ideally also update the class diagram. The config tutorial is updated already, right?

Waelthus:
also note that the linting check fails because of duplicated code and an indentation by 4 spaces too much, i.e.:

> ************* Module picca.delta_extraction.masks.bal_mask
> py/picca/delta_extraction/masks/bal_mask.py:240:0: W0311: Bad indentation. Found 16 spaces, expected 12 (bad-indentation)
> ************* Module picca.delta_extraction.quasar_catalogues.drq_catalogue
> py/picca/delta_extraction/quasar_catalogues/drq_catalogue.py:1:0: R0801: Similar lines in 2 files
> ==picca.delta_extraction.expected_fluxes.dr16_expected_flux:[468:508]
> ==picca.delta_extraction.expected_fluxes.true_continuum:[173:196]

py/picca/delta_extraction/astronomical_objects/forest.py
```diff
@@ -122,73 +121,78 @@ def rebin(log_lambda, flux, ivar, transmission_correction, z, wave_solution,
...
-        bins = np.zeros(log_lambda.size, dtype=np.int64)
-        rebin_ivar = np.zeros(log_lambda.size)
-        w1 = np.zeros(log_lambda.size, dtype=bool_)
-        w2 = np.zeros(log_lambda.size, dtype=bool_)
-        return (log_lambda, flux, ivar, transmission_correction, mean_snr, bins,
-                rebin_ivar, orig_ivar, w1, w2)
 
     bins = find_bins(log_lambda, log_lambda_grid, wave_solution)
+    binned_arr_size = bins.max() + 1
     log_lambda = log_lambda_grid[0] + bins * pixel_step
```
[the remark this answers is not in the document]
p-slash: Yes, I think only out-of-boundary pixels are removed from spectra at this points. That is there are separate `w1` and `ivar>0` masks.

py/picca/delta_extraction/astronomical_objects/forest.py
```diff
@@ -122,73 +121,78 @@ def rebin(log_lambda, flux, ivar, transmission_correction, z, wave_solution,
...
 
+    # Find non-empty bins
+    bincounts = np.bincount(bins, minlength=binned_arr_size)
+    wnonempty_bins = bincounts != 0
```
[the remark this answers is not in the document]
p-slash: I can't remember well. I think there are empty bins on the lower end by construction?

py/picca/delta_extraction/astronomical_objects/forest.py
```diff
@@ -122,73 +121,78 @@ def rebin(log_lambda, flux, ivar, transmission_correction, z, wave_solution,
...
-    w2 = (rebin_ivar > 0.)
-    flux = rebin_flux[w2] / rebin_ivar[w2]
-    transmission_correction = rebin_transmission_correction[w2] / rebin_ivar[w2]
-    ivar = rebin_ivar[w2]
+    w2_ = (rebin_ivar > 0.) & wnonempty_bins
+    w2  = w2_[wnonempty_bins]
+    flux = np.zeros(final_arr_size)
+    transmission_correction = np.zeros(final_arr_size)
+    ivar = np.zeros(final_arr_size)
+
```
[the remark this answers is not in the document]
p-slash: I think this essentially removes pixels at the lower boundary. There shouldn't be gaps between pixels still

py/picca/delta_extraction/masks/lines_mask.py
```diff
@@ -32,15 +32,15 @@ class LinesMask(Mask):
     Table with the observed-frame wavelength of the lines to mask. This usually
     contains the sky lines.
     """
-    def __init__(self, config):
+    def __init__(self, config, keep_masked_pixels=False):
```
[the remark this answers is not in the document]
iprafols: I would also read it from the config

py/picca/delta_extraction/expected_flux.py
```diff
@@ -268,6 +267,24 @@ def compute_forest_variance(self, forest, continuum):  # pylint: disable=no-self
         raise ExpectedFluxError("Function 'compute_forest_variance' was not "
                                 "overloaded by child class")
 
+    def compute_forest_weight(self, forest, continuum):
```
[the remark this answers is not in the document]
iprafols: Also, if we default variance as `np.inf` for masked pixels as Michael suggested, then this function would take care of itself.

py/picca/delta_extraction/least_squares/least_squares_var_stats.py
```diff
@@ -151,8 +151,13 @@ def initialize_delta_arrays(self, forests):
             # ignore forest if continuum could not be computed
             if forest.continuum is None:
                 continue
-            var_pipe = 1 / forest.ivar / forest.continuum**2
-            w = ((np.log10(var_pipe) > VAR_PIPE_MIN) &
+
+            w =  forest.ivar > 0
+            var_pipe = np.empty_like(forest.log_lambda)
+            var_pipe[w]  = 1 / forest.ivar[w] / forest.continuum[w]**2
+            var_pipe[~w] = VAR_PIPE_MAX+1
```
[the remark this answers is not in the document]
iprafols: In my opinion, it makes more sense to default to np.inf, if only for clarity

py/picca/delta_extraction/least_squares/least_squares_var_stats.py
```diff
@@ -169,21 +174,14 @@ def initialize_delta_arrays(self, forests):
...
             # add contributions to delta statistics
-            rebin = np.bincount(bins, weights=delta)
-            mean_delta[:len(rebin)] += rebin
-
-            rebin = np.bincount(bins, weights=delta**2)
-            var_delta[:len(rebin)] += rebin
-
-            rebin = np.bincount(bins, weights=delta**4)
-            var2_delta[:len(rebin)] += rebin
+            mean_delta += np.bincount(bins, weights=delta, minlength=mean_delta.size)
```
iprafols, line 180: Not sure these changes will not crash in some cases. Why can we now drop the `[:len(rebin)]`?

py/picca/delta_extraction/survey.py
```diff
@@ -248,7 +248,7 @@ def read_masks(self):
         self.logger.info(f"Reading masks. There are {num_masks} masks")
 
         for MaskType, mask_arguments in self.config.masks:
-            mask = MaskType(mask_arguments)
+            mask = MaskType(mask_arguments, self.config.keep_masked_pixels)
```
iprafols: see my comments on config.py

py/picca/tests/delta_extraction/abstract_test.py
```diff
@@ -145,6 +145,8 @@ def compare_fits(self, orig_file, new_file):
                 new_header = new_hdul[hdu_name].header
                 for key in orig_header:
                     self.assertTrue(key in new_header)
+                    if key == 'NAXIS2':
```
iprafols, line 148: why do we expect NAXIS2 to have issues? Why do we need to treat it in a special way?

py/picca/tests/delta_extraction/abstract_test.py
```diff
@@ -175,20 +178,28 @@ def compare_fits(self, orig_file, new_file):
                             print(f"Column {col} in HDU {orig_header['EXTNAME']} "
                                   "missing in new file")
                         self.assertTrue(col in new_data.dtype.names)
+                        
+                        if 'IVAR' in new_data.dtype.names:
+                            w = new_data['IVAR']>0
+                        elif 'WEIGHT' in new_data.dtype.names:
+                             w = new_data['WEIGHT']>0
```
iprafols, line 185: there seems to be identation issues here

py/picca/tests/delta_extraction/abstract_test.py
```diff
@@ -175,20 +178,28 @@ def compare_fits(self, orig_file, new_file):
                             print(f"Column {col} in HDU {orig_header['EXTNAME']} "
                                   "missing in new file")
                         self.assertTrue(col in new_data.dtype.names)
+                        
+                        if 'IVAR' in new_data.dtype.names:
```
iprafols, line 182: why is this necessary? Even if masked pixels are kept, why would they be different in different runs of the tests?

iprafols: This applies to the rest of the changes in this file from here onwards

py/picca/tests/delta_extraction/expected_flux_tests.py
```diff
@@ -170,14 +170,14 @@ def test_dr16_expected_flux_compute_continuum_lin(self):
             elif continua.get(forest.los_id) is None:
                 self.assertTrue(forest.continuum is None)
             else:
-                if not np.allclose(forest.continuum, continua.get(forest.los_id)):
+                if not np.allclose(forest.continuum[forest.ivar>0], continua.get(forest.los_id)):
```
iprafols, line 173: I'd rather update the test suite with the correct changes than to keep this potentially faulty check

iprafols (commented):
Changes in general look fine, but I have some concerns that should be addressed before merging

py/picca/delta_extraction/astronomical_objects/desi_pk1d_forest.py
```diff
@@ -197,22 +197,33 @@ def rebin(self):
         AstronomicalObjectError if Forest.wave_solution is not 'lin' or 'log'
         """
         bins, rebin_ivar, orig_ivar, w1, w2 = super().rebin()
-        if len(rebin_ivar) == 0:
+        if len(rebin_ivar) == 0 or np.sum(w2) == 0:
             self.resolution_matrix = np.array([[]])
-            return [], [], [], [], []
+            return [], [], [], [], np.array([])
```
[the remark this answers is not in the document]
p-slash: I think something complained (pylint) or using `array.size` somewhere.

py/picca/tests/delta_extraction/expected_flux_tests.py
p-slash: These test code changes were to make sure masking behaved the same as before. Since new files have more pixels, the array sizes are different. But these codes would be removed if baseline files are replaced with new ones. I think the changes are safe at this points, so I'll replace test files and use them going forward with the comments

py/picca/tests/delta_extraction/abstract_test.py
p-slash: array size is different. I'm replacing test files with the new ones, so it will be removed

py/picca/delta_extraction/least_squares/least_squares_var_stats.py
p-slash: It shouldn't crash. minlength argument makes sure the array `np.bincount` returns is the same size as `mean_delta` (since there are no overflows). Without this argument `np.bincount` removes the array at the right end, so you needed `[:len(rebin)]`.

py/picca/delta_extraction/config.py
```diff
@@ -34,7 +34,7 @@
     "overwrite", "logging level console", "logging level file", "log",
     "out dir", "num processors"
 ]
-accepted_masks_options = ["num masks", "type {int}", "module name {int}"]
+accepted_masks_options = ["num masks", "keep pixels", "type {int}", "module name {int}"]
```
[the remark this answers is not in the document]
p-slash: I now read the option from config. `Mask` class in now initialized with config argument.

p-slash:
Great comments and nice catches. I did a bunch of updates:

- changed `variance=0` to `np.inf`. Got rid of `compute_weights` or whatever.
- `keep pixels` is now read in config.
- Changed the test files, so no hack is needed in scripts.

py/picca/delta_extraction/astronomical_objects/desi_pk1d_forest.py
p-slash: Interesting point. The resolution matrix values are still needed even when `ivar=0`, so this was my solution to keep them. We can go back to this later

Waelthus: Agreed, for the moment the resolution matrix values are still needed for central pixel with `ivar!=0` and offset pixel `ivar==0`, not sure if there's a way around this anyway... Regarding central pixels with `ivar==0` I guess we don't need anything, right? 

And regarding the weighting at rebinning stage I'd probably ignore masked pixels, but as said when we're actually rebinning the concept of using the resolution matrix doesn't really make sense (without at least a "rebinning" of the resolution matrix in both its dimensions, that is, which would then produce the kind of issues we saw in the quickquasars reso matrix).

Waelthus:
> > also note that the linting check fails because of duplicated code and an indentation by 4 spaces too much, i.e.:
> > > ************* Module picca.delta_extraction.masks.bal_mask
> > > py/picca/delta_extraction/masks/bal_mask.py:240:0: W0311: Bad indentation. Found 16 spaces, expected 12 (bad-indentation)
> > > ************* Module picca.delta_extraction.quasar_catalogues.drq_catalogue
> > > py/picca/delta_extraction/quasar_catalogues/drq_catalogue.py:1:0: R0801: Similar lines in 2 files
> > > ==picca.delta_extraction.expected_fluxes.dr16_expected_flux:[468:508]
> > > ==picca.delta_extraction.expected_fluxes.true_continuum:[173:196]
> 
> I moved the duplicated code to `expected_flux.py`. Now child classes inherit this method.

and this indeed fixed the linting errors.

py/picca/delta_extraction/astronomical_objects/forest.py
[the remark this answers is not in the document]
Waelthus: hm, I guess this could happen depending on how the wavelength grid is chosen in restframe and observed frame. But ideally we should at least prevent this from happening in the default setup. Maybe I'm ignoring some detail, @iprafols?

py/picca/delta_extraction/astronomical_objects/forest.py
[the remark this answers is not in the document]
Waelthus: Can you double check what you get when e.g. using a Delta_lambda=0.7 on DESI data with the normal spacing, will this produce an even grid with that spacing or have gaps in between. I'd guess the latter, but it's not super relevant right now (might be when adding other datasets at a later time?). If the former it would be interesting to know how the gaps are filled...

py/picca/delta_extraction/config.py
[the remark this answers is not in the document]
Waelthus: shouldn't this be a general option as well and be passed to all masks?

py/picca/delta_extraction/least_squares/least_squares_var_stats.py
Waelthus: isn't minlength just ensuring it's at least that size? But agreed, it might be that now that we're not dropping pixels, but keeping them some of those [:...] statements can be removed. I'd need to think about this in more depth, to fully grasp where this is now possible and where not.

Waelthus (commented):
Changes look mostly fine to me.
The only thing that really bothers me is that the test results have changed (which is expected as no pixels are dropped anymore). Can you check that the new test outputs masked to [IVAR>0] are identical to the old ones (in the same sense that the comparison routines do, i.e. np.isclose or whatever it was called).
Just to make sure we're not accidentally introducing a change of results here...

p-slash:
> Changes look mostly fine to me.
> The only thing that really bothers me is that the test results have changed (which is expected as no pixels are dropped anymore). Can you check that the new test outputs masked to [IVAR>0] are identical to the old ones (in the same sense that the comparison routines do, i.e. np.isclose or whatever it was called).
> Just to make sure we're not accidentally introducing a change of results here...

Tests have passed before I introduced changes to the test files. So It was recovering the previous behavior

p-slash:
Alright, I rewrote the necessary parts in `astronomical_objects_tests.py`. It was hard to disentangle, so I transferred kwargs things to functions. Then, I was able to track what was happening in each function.
The main difference is truth is now calculated for equal spacing. If we agree on that part, this PR should be good to go.
I am using this branch in my p1d reductions, and it works. BUT, I'd understand if we create a version before merging this. Maybe it is better if we do that. Publish 6.0, then merge, then call this 6.1. What do you think?

Waelthus:
New tests needed for the last change

Waelthus:
Please merge when those are added, the new version has been allowing us to easily go back...

p-slash:
I made changes had to change some nans to zeros in test files (see last commit). I will merge if it passes

p-slash:
I think I need one of you to approve these changes @Waelthus  @iprafols

Waelthus:
note that the code testing coverage decreased with the changes, so that's the unsuccessful test

Waelthus (approved):
I already gave my ok earlier, but I think we didn't approve yet because of the not yet available intermediate release (which we now have)

iprafols:
I'd say we can go ahead and merge it even if the coveralls test is not passing. @Waelthus, do you agree?

Waelthus:
In principle I'd be fine with it.
Do we know where the decrease in coverage comes from? I.e. is it new functionality that is just not tested yet, or is it something else that changed the ratio of untested lines, or even just some setup change leading to definitions of what counts as tested/untested being different?

iprafols:
It's mostly a few lines that were added that are not checked in the tests. e.g. method `get_inner_region_slice` in forest.py. It'd be better to add some new tests, but I think we risk delaying and adding complexity (=more bugs) if we wait longer to merge this

PR #923, opened 2022-08-05
Resomat nan fix

p-slash:
This patch fixes non-numerical errors in spectral_resolution_desi when handling resolution matrix with negative elements etc.


py/picca/delta_extraction/utils_pk1d.py
```diff
@@ -213,44 +214,105 @@ def spectral_resolution(wdisp,
...
+    # using an arbitrary epsilon is not stable
+    # use nonzero absolute minimum of the row
+    nonzero_abs_min_per_row = _find_nonzero_abs_min_per_row(reso_matrix, num_rows)
+
+    # mask out rows that are all zeros
+    w = nonzero_abs_min_per_row>0
+    if not np.any(w):
+        return np.zeros_like(lambda_), np.zeros_like(lambda_)
+
+    shift = reso_matrix.min(axis=0) - nonzero_abs_min_per_row
```
Waelthus, line 286: so this is to shift up the reso_matrix elements in case there are negaftive/zero ones? To avoid NANs later in the process? I guess it's never strongly negative and this is fine and just catching a numerical effect.
Or is it becoming strongly negative? In which case a more sophisticated solution might be necessary...

py/picca/delta_extraction/utils_pk1d.py
```diff
@@ -213,44 +214,105 @@ def spectral_resolution(wdisp,
...
-             reso[reso.shape[0] // 2, :] / reso[reso.shape[0] // 2 + 2, :])))
-        / 4.0) #this is rms
+    indices = np.array([-2, -1, 1, 2], dtype=int)
+    ratios = reso[num_offdiags, :]/reso[num_offdiags+indices, :]
+    ratios = np.log(ratios)
+    w2 = ratios > 0
+    norm = np.sum(w2, axis=0)
+    new_ratios = np.zeros_like(ratios)
+    new_ratios[w2] = 1./np.sqrt(ratios[w2])
+    # ratios = 1./np.sqrt(np.log(ratios))
```
Waelthus, line 300: thanks for the overhaul of this part. Note that those calculations are just to get a Gaussian kernel based resolution estimate for each pixel which is mostly used as a cross-check for the FFT based analysis at least

py/picca/delta_extraction/utils_pk1d.py
```diff
@@ -213,44 +214,105 @@ def spectral_resolution(wdisp,
...
+    # use nonzero absolute minimum of the row
+    nonzero_abs_min_per_row = _find_nonzero_abs_min_per_row(reso_matrix, num_rows)
+
+    # mask out rows that are all zeros
+    w = nonzero_abs_min_per_row>0
+    if not np.any(w):
+        return np.zeros_like(lambda_), np.zeros_like(lambda_)
+
+    shift = reso_matrix.min(axis=0) - nonzero_abs_min_per_row
+    reso = reso_matrix[:, w] - shift[w]
```
Waelthus, line 287: is this shift needed only for these simplified calculations of resolution? Or do we need something similar when computing corrections based on the matrix itself? I guess not, as this kind of shifts would only affect the mean, i.e. FFT(R) (k=0) and not the higher modes...

Waelthus (approved):
Looks good to me, see comments above

p-slash:
Resolution matrix can become strongly negative. In fact it deviates quite a bit from Gaussianity for some rows such that off diagonals are larger than the central value. Hence all the hoops to jump. I would make minimal assumptions when correcting p1d; and it sounds like that's the case.
I mean... We might as well remove this value from files & calculations if it is not used :)

Waelthus:
Indeed we probably could. The main reason to have it is that it was an output in the SDSS based versions and it was simplifying cross checking between old and new versions when doing the changes for pk1d delta_extraction.

Are you using the values from here for a correction in the QE analysis? Or are you just using the (modified? masked?) entries of the resolution matrix itself as input? If no-one is using these values, there's not really a reason to try getting those right and I would veto against this change (but maybe have a nanmean when getting mean resolution outputs). Then one could use it to debug spectra with weird resolution matrices...

p-slash:
No I don't use these values at all. I don't want to see a flood of warnings though. So we can get rid of this function completely, or merge these changes for now. OR keep it separate and I use it in my own branch, which is a little risky when something new comes along.

Waelthus:
shouldn't we then just suppress the warnings for the part of the code that creates them? Given that we know the source? Or are those created downstream at a later point (or even in another code)?
If this is complicated then yes, please merge...

p-slash:
I'd be fine with suppressing. That didn't occur to me. Don't know how though. Up to you, we can keep these or revert and suppress. You're more familiar with this part of the code.

Waelthus:
what kind of warning do you get? I'd probably just comment the shift and add the filter for that warning in here

iprafols:
I'm confused about this PR, why are the tests not being triggered?

Waelthus:
now that the tests have been running: I didn't see any of the warnings there that @p-slash got annoyed by. If this is actually fixing your issues regarding warnings spam without actually changing things a lot, I'd be happy to clean it up and merge. 
The failing tests are because currently and updated version of delta-1.fits is checked in that would need to be reverted...

p-slash:
> now that the tests have been running: I didn't see any of the warnings there that @p-slash got annoyed by. If this is actually fixing your issues regarding warnings spam without actually changing things a lot, I'd be happy to clean it up and merge. The failing tests are because currently and updated version of delta-1.fits is checked in that would need to be reverted...

That's fine by me. It might be easy to discard this pull request and add the warning filter to the master in a separate branch. Thanks for doing this!

Waelthus:
I guess at the current time we'd only need to revert the test file (which I just did) and maybe clean up currently commented lines. There were some improvements in here that might be worth keeping around...
Let's see if the tests run through fine.

Waelthus:
I'll merge this whenever tests have run through again (the last comment only removed comments, so should still be fine)...

PR #925, opened 2022-08-18
edit picca to accept fugu catalogues

Waelthus:
this is a PR to merge Calum's changes for fugu analyses into master


py/picca/delta_extraction/data_catalogues/desi_healpix.py
```diff
@@ -73,11 +73,16 @@ def get_filename(self, survey, healpix):
         is_mock: bool
         False, as we are reading DESI data
         """
-        input_directory = f'{self.input_directory}/{survey}/dark'
+        if survey.lower().startswith("sv"):
```
Waelthus, line 76: ok, so if using a fugu catalog this would be run using a guadalupe input directory and switches to  fuji files for everything sv related? 
This is not really transparent and I guess we'd need to have some description about how this is run, and ideally we should have a proper fugu mode

p-slash: I think @corentinravoux's solution is better than changing the code tbh. We can simply create a folder with links to surveys and call it fujilupe. I have a minimal one in desi/users/naimgk.

Waelthus: if we could have some directory linking to files in fuji or guadalupe that should work as well (at least from my perspective), would just need to have a directory containing the links for all the data and is available to everyone...
any thoughts @andreufont

andreufont: I'm not sure I understand the proposed alternative by Naim / Corentin. You would create a folder (names fujilupe?) that would have two sub-folders (fuji and guadalupe) that would be links to the actual folders of Fuji and Guadalupe? That would still require different code (since you would have the sub-folders). 

Anyway, I'd be happy with any implementation that allows us to use the master branch for SV only analyses (reading Fuji), DA02 analyses (reading Guadalupe) or combined analyses (reading both).

andreufont: How about this alternative: for any Picca run with DESI data, the `input directory` option should point to the lowest level directory where all data reductions leave. At NERSC this would be:
`/global/cfs/cdirs/desi/spectro/redux/`
 
Then, using the value of `survey` that we find in the quasar catalog, we can reconstruct the full path to that file. 

The only problem with this is that there might be more than one path with different releases. For instance, soon there will be the same quasar under `/redux/himalaya/healpix/` and under `/redux/iron/healpix/`. 

Maybe we should specify not only `input directory`, but also `release`. If the user passes `release=iron`, then it would look for the spectrum under `/redux/iron/healpix/`, but if `release=himalaya` or `release=guadalaya` (Guadalupe + Himalaya), then it would look for the spectrum under `/redux/himalaya/healpix/`.

andreufont: The alternative is to deprecate the use of Himalaya as soon as we have Iron, just like we could have deprecated Everest as soon as we had Fuji and Guadalupe. I'd prefer to keep the option to run on older releases if not too much headache.

andreufont: @alxogm - what do you think?

Waelthus: I'm not completely sure about @corentinravoux's way of doing this, but would've guessed it's like having a fugu directory that contains all the healpix directories (all levels of the hierarchy) and then links to the fuji/guadalupe files in there, i.e. same structure as fuji or guadalupe, but files of both. With the disadvantage that every link can only point to one of the files...

Waelthus: ah, actually it's different from what I thought, with a directory structure like `fujilupe/healpix/{main,sv1,sv2,sv3}/{bright,dark}/`which I would've expected to require code changes as well

Waelthus: In principle we could also modify the code to just accept multiple input dirs, not sure how much work that would be right now, but this would be a clean solution where just every input directory is listed in the config

andreufont: And how do you know which one of the input dirs to use for each spectrum? Try all of them, and check that only one contains a file with such a name? Or use the "survey" entry to decide which input dir to use? 

I kind of prefer the alternative to specify the release in the config file, and then point to the base directory (`/global/cfs/cdirs/desi/spectro/redux/`)

Waelthus: true, I guess checking all files for duplicate entries would be troublesome, we could use the survey entries, but that's an issue with guadelaya (where survey will be main always, I guess).

Specifying a release in the config file wouldn't really work either for multi-release setups like fugu or guadelaya, I guess except if we're telling picca somewhere which release to use for which target or let picca read in everything and coadd across all releases...

p-slash: I suppose it was easier for fujilupe since fuji doesn't have main survey and guadalupe doesn't have SV. Looking briefly at the code, one option would be rename `main` folders to `main-{release}` when linking and pass that as `SURVEY` in catalog. No code changes, but a redefinition of survey in catalogs.

p-slash: Picca coadds spectra with same `TARGETID`, so no problem in duplicates.

Waelthus: But if we do not want to select a particular release and just coadd them all, we could actually just go through multiple input directories and fetch all files that have entries in the cat (or alternatively specify multiple releases to do the same). We'd probably get some objects from each healpix file in each release we use and thus need to open all files anyway. So overhead shouldn't be large.

The complications only arise if we want to select a release to be used and not coadd across releases...

Waelthus: ok, It would be the most flexible solution as one can decide at cat creation which releases to use and when publishing the cats it's also clear which version of the reductions has been used for each target. I like this idea.

Note, this is a task for both picca modifications (parsing that release column and doing the code update for generating path from release) and catalog modifications. So we'll need to synchronize both tasks. And it might be best if we're not rushing this while many people are on holidays.

If we'd like to keep this branch as a intermediate solution until we have a final fix, I'd be happy to merge master back into here now, so that the analyses are identical except for the fugu usage...

iprafols:
Hi, @Waelthus @calumgordon , 
I implemented the solution by @p-slash  discussed  in the last telecon. If I did things correctly, then the main branch should work for fugu using `/global/cfs/cdirs/desi/spectro/redux/fugu/healpix` as `input directory`. @calumgordon, can you check that this is correct? If so, then we'll close the PR

andreufont:
Wait, have you created a folder there? I would be surprised that the data team gave you permission to do that... Let's discuss on Slack

Waelthus:
I think in terms of file system permissions, this is possible for all members of the `desi` group, but would be good to check if that's ok. I would've expected someone's desi/users directory for this purpose.
Also: should we check that this branch and the fugu directory on master are giving the same answer?

iprafols:
I have moved the folder to `/global/cfs/projectdirs/desi/users/iprafols/desi` in the meantime while we decide what to do. This way we can move forward in testing this.

iprafols:
the symlink folder `/global/cfs/cdirs/desi/science/lya/fugu_healpix/healpix` produces the same results as the fugu branch, so there is no longer any need for this PR to be merged. Will close it in a couple days unless someone has a reason not to

Waelthus:
I completely agree. We should maybe have a "how to do fugu analysis" as part of the readme or sth, but this PR can be closed

Waelthus:
Closing this after having decided to deal with things differently.

PR #928, opened 2022-09-08
Raw io ohio 2

p-slash:
These are some old changes I resurrected (mostly following raw-io-ohio). I cleared up the error messages in TrueContinuum, increased tolerance for wavelength choices (was driving me crazy) and added compatibility for Ohio P1D mocks.
pylint was being silly on NERSC. Or maybe I'm just not seeing it.


py/picca/delta_extraction/expected_fluxes/true_continuum.py
```diff
@@ -396,65 +397,57 @@ def read_raw_statistics(self):
...
-        hdul.close()
+            pixel_step = 10**Forest.log_lambda_grid[1] - 10**Forest.log_lambda_grid[0]
+            pixel_step_key = 'DEL_L'
+            log_lambda = np.log10(fits_data['LAMBDA'])
+
+            if not header['LINEAR']:
+                is_rawfile_consistent = False
+                err_msg += "File wave solution is not linear.\n"
+
+        def _check_header_consistency(_key, _test_val, atol):
```
Waelthus, line 424: This function is the source of the pylint errors.
The reason is that you're using variables that are not defined within the scope of the function (err_msg) or are potentially just defined in the scope of the function (is_rawfile_consistent), i.e. cannot be accessed outside.
For this to work the variables would either need to be declared as global (not sure if that works and also bad style) or would need to be passed to the function by reference (I'm never sure how to guarantee this in python for stuff like a single bool) or by value and then returned in the end...

p-slash: I thought locally declared functions can access variables outside. We shouldn't declare global. I will rewrite to return values, but this just looked cleaner.

PR #930, opened 2022-09-16
forest continuum is multiplied by transmission_correction in true

p-slash:
forest continuum is multiplied by transmission_correction in true continuum. this correction is needed for DLA masking + true continuum analysis


Waelthus (approved):
Right, for the standard cont fitting the transmission correction is applied on each forest during cont fitting, while here it was ignored so far. I.e. damping wings of DLAs would stay inside the spectra, messing up analyses that e.g. try to disentangle impacts of DLAs and cont fitting.
Not sure if anyone ever would want to do true continuum analyses with the optical depth correction, but there this would trigger as well...

PR #931, opened 2022-09-16
Dla cat colname fix

p-slash:
There was a bug when DLA redshift column is "Z_DLA". Strictly speaking only one line of code had to be changed, but I made some aesthetic changes.


Waelthus (approved):
good for me, looks like we only had a partial fix for this case in place before

PR #933, opened 2022-09-21
Speed up for P1d mock analysis

p-slash:
speed up when doing Pk1D analysis using mocks that reads the resolution matrix from truth files by opening each truth file once. hopefully, it is significant enough to be effective on y5 mock analysis.

### DIFF
--- a/py/picca/delta_extraction/data_catalogues/desi_healpix.py
+++ b/py/picca/delta_extraction/data_catalogues/desi_healpix.py
@@ -205,7 +205,21 @@ class DesiHealpixFileHandler(DesiDataFileHandler):
             self.logger.warning(
                 f"Missing Z band from {filename}. Ignoring color.")
 
+        hdul_truth = None
         reso_from_truth = False
+        all_reso_present = all(x in hdul for x in (f"{color}_RESOLUTION" for color in colors))
+        if self.analysis_type == "PK 1D" and not all_reso_present:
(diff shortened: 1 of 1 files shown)


py/picca/delta_extraction/data_catalogues/desi_healpix.py
```diff
@@ -216,35 +230,29 @@ def read_file(self, filename, catalogue):
...
-                                spec["RESO"] = hdul_truth[
-                                    f"{color}_RESOLUTION"].read()
-                        else:
-                            raise DataError(
-                                f"Error while reading {color} band from "
-                                f"{filename}. Analysis type is 'PK 1D', "
-                                "but file does not contain HDU "
-                                f"'{color}_RESOLUTION'")
+
+                if self.analysis_type != "PK 1D":
```
iprafols, line 234: This if feels weird. It makes line 235 be duplicated unnecessarily. How about just checking for `self.analysis_type == "PK 1D"` and then have the block 238 to 247 inside that if?

iprafols (changes requested):
I have a few concerns/improvements that I think could improve the code's clarity. @p-slash can you take a look at them and tell me what you think?

p-slash:
@iprafols great points. changed the code and pushed a commit.

iprafols (approved):

PR #934, opened 2022-10-21
Raw_pk1d

Waelthus:
This is a fix that allows picca_pk1d to run on raw transmission delta outputs by deactivating noise and reso correction


bin/picca_Pk1D.py
```diff
@@ -70,7 +70,24 @@ def process_all_files(index_file_args):
...
+                delta.mean_snr=1e5
+                delta.mean_reso=1e-3
+                delta.mean_reso_pix=1e-3
+                delta.exposures_diff = np.zeros(delta.delta.shape)
+                #if args.linear_binning:
+                #    delta.dlambda = np.median(sp.diff(10**d.ll))  #(d.ll[-1]-d.ll[0])/(len(d.ll)-1) #both of those should give the same result, but the first is more explicite, second one should be faster, but this shouldn't be a dominant effect
+                #else:
+                #    delta.dll = np.mean(np.diff(d.ll))  #(d.ll[-1]-d.ll[0])/(len(d.ll)-1) #both of those should give the same result, but the first is more explicite, second one should be faster, but this shouldn't be a dominant effect
+                #delta.mean_resolution_matrix=???
+            running_on_raw_transmission = True
```
corentinravoux, line 85: for the values you used by default, do you think it will have an impact on the result? why not taking larger ones just to be sure?

bin/picca_Pk1D.py
```diff
@@ -70,7 +70,24 @@ def process_all_files(index_file_args):
     # read fits or ascii file
     if args.in_format == 'fits':
         hdul = fitsio.FITS(file)
-        deltas = [Delta.from_fitsio(hdu, pk1d_type=True) for hdu in hdul[1:]]
+        try:
+            deltas = [Delta.from_fitsio(hdu, pk1d_type=True) for hdu in hdul[1:]]
+            running_on_raw_transmission = False
+        except ValueError:
+            print("\nPk1d_type=True didn't work on read in, maybe perfect model? Trying without!")
+            deltas = [delta.from_fitsio(hdu,Pk1D_type=False) for hdu in hdul[1:]]
```
corentinravoux, line 78: I'm testing this and I have the following error at this line:
NameError: free variable 'delta' referenced before assignment in enclosing scope
Maybe you do not need a from_fitsio function here ?

corentinravoux (changes requested):
There is an error for             deltas = [delta.from_fitsio(hdu,Pk1D_type=False) for hdu in hdul[1:]],
I do not think we need to use a from_fitsio function

bin/picca_Pk1D.py
Waelthus: the values are just put there so they can be reported in the output files (and so that intermediate steps trying to select parts of those arrays do not crash), they are used nowhere in the actual calculation.

Waelthus:
ah, yes. There has been a change in capitalization convention. Good catch, thanks for the fix.

bin/picca_Pk1D.py
corentinravoux: ok makes sense

corentinravoux (approved):

Waelthus:
this still needs an update to the tests to be merged (i.e. testing a "raw-like" delta file for pk1d) so that coverage tests are happy

corentinravoux:
> this still needs an update to the tests to be merged (i.e. testing a "raw-like" delta file for pk1d) so that coverage tests are happy

Can you point me to the testing part, I'm not really used to the testing, but I will gladly do it for practice

Waelthus:
In general you'd need to have a small file of deltas in the right format, need to have the outputs as well for a run on just those deltas and add another function to the test.

Note that outputs of those tests are created by the test in `Products/...` and compared to stored outputs in `py/picca/test/test_Pk1D/`.

The test that would need to be added should be sth like `test_pk1d_raw` and would be part of the  `TestPk1d` class in `https://github.com/igmhub/picca/blob/master/py/picca/tests/test_2_pk1d.py#L53` (essentially the same one as existing just with modified file names). 

The simplest might be running on an existing output delta file made for BAO analyses. If that runs the lines from the raw analysis you could use it as input and only need to save a new output into picca (without thinking about which exact deltas to use as input).

iprafols:
@corentinravoux  any progress on updating the tests?

corentinravoux:
Sorry I forgot about this. It should be ok now

iprafols:
Great, I take it that all the issues above are resolved and that we can merge this?

corentinravoux:
For me yes, it know works on raw data and I think Marie-Lynn tested it again.

PR #935, opened 2022-11-10
Add delta rebinning and optional fix for true continuum mean deltas

andreicuceu:
The main update is an option to rebin deltas before computing correlations / distortion matrices. Needed for noiseless analyses, and for reducing the impact of biased weights in noisy analyses.

The second update is adding the "force stack delta to zero" option to the true continuum analysis to fix the mean delta different from zero problem. However, unlike with continuum fitting, this is turned off by default here.

### DIFF
--- a/py/picca/delta_extraction/expected_fluxes/true_continuum.py
+++ b/py/picca/delta_extraction/expected_fluxes/true_continuum.py
@@ -17,7 +17,7 @@ from picca.delta_extraction.utils import (find_bins, update_accepted_options,
 
 accepted_options = update_accepted_options(accepted_options, [
     "input directory", "raw statistics file", "use constant weight",
-    "num bins variance"
+    "num bins variance", "force stack delta to zero"
 ])
 
 defaults = update_default_options(defaults, {
--- a/py/picca/delta_extraction/expected_fluxes/true_continuum.py
+++ b/py/picca/delta_extraction/expected_fluxes/true_continuum.py
@@ -23,6 +23,7 @@ accepted_options = update_accepted_options(accepted_options, [
 defaults = update_default_options(defaults, {
     "raw statistics file": "",
     "use constant weight": False,
+    "force stack delta to zero": False
 })
 
 IN_NSIDE = 16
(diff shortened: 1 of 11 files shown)


py/picca/delta_extraction/expected_fluxes/true_continuum.py
```diff
@@ -224,13 +227,16 @@ def populate_los_ids(self, forests):
...
                 weights[w] = 1
-                weights[~w]= 0
-                mean_expected_flux = forest.continuum
+                weights[~w] = 0
             else:
-                mean_expected_flux = forest.continuum
                 weights = 1. / self.compute_forest_variance(
                     forest, forest.continuum)
 
+            mean_expected_flux = forest.continuum
```
p-slash, line 235: We may want to use `mean_expected_flux = np.copy(forest.continuum)` to have same behavior as fitting. Explicit copying keeps continuum unmodified, which is the default behavior for fitting.

py/picca/io.py
```diff
@@ -1415,7 +1415,7 @@ def read_blinding(in_dir):
     return blinding
 
 
-def read_delta_file(filename, from_image=None):
+def read_delta_file(filename, from_image=None, rebin_factor=None):
```
p-slash, line 1418: Update doctoring with resin_factor

py/picca/io.py
```diff
@@ -1447,7 +1452,8 @@ def read_deltas(in_dir,
                 max_num_spec=None,
                 no_project=False,
                 from_image=None,
-                nproc=None):
+                nproc=None,
+                rebin_factor=None):
```
p-slash, line 1456: Update doctoring

p-slash (commented):
Some minor comments

py/picca/delta_extraction/expected_fluxes/true_continuum.py
iprafols: Makes sense to me

iprafols (commented):
There are some comments by @p-slash  that should be addressed before merging. Also, unless there is some reason not to, I think the rebinning should also be added to the metal distortion matrices.

iprafols (approved):

PR #939, opened 2022-12-06
Raw analysis updates

andreicuceu:
I added the option to specify the tracer hdu name so the raw analysis works with the Lyman-beta region and with metals. Also added the option to use spline interpolated mean flux and var_lss functions, which helps with stability, especially when running on metals.


andreicuceu:
@alxogm have a look and see if this fixes the problems you were having. For metals you just pass "tracer='F_METALS'" and "use_splines=True"..

alxogm:
thanks @andreicuceu , I have tested it and seems to be working fine. I just added a couple of things that were missing in order to use the new functionality. If you agree with such additions then I think it can be merged.

alxogm (approved):

PR #940, opened 2022-12-08
Update setup to add rejection logs folder

cramirezpe:
rejection logs were not reachable by the package if installed through pip install. They should be declared in setup.py


andreufont:
I have no idea about how these setup work, but the fix seems to work.

andreicuceu (approved):

Issue #941, opened 2022-12-08
Improve setup file


andreicuceu:
The current setup.py files declares a bunch of submodules in delta_extraction as modules, but all of these are internal things in the code and will most likely never be used independently. So we should hide all of them from the setup. This could be achieved through using the proper import paths, or locally exporting them in the __init__.py files so other parts of the code can easily see them.

PR #943, opened 2022-12-10
fix issue 941

iprafols:
I think this fixes issue #941


cramirezpe (approved):
I was able to import all the modules listed in the old file using the new setup.py

PR #944, opened 2022-12-15
New blinding

iprafols:
This branch contains a fix to enforce the blinding strategy using column LASTNIGHT. The new fix also automatically picks different blinding seeds for early data, year 1 data and year 3 data


iprafols:
Some things to do before merging:

- [ ] Fix tests
- [ ] Remove patch to read catalogue from hdu 1

py/picca/delta_extraction/utils.py
```diff
@@ -82,8 +82,10 @@
     "LY10": 919.3514,
 }
 
-ACCEPTED_BLINDING_STRATEGIES = ["none", "minimal", "corr_yshift"]
-
+ACCEPTED_BLINDING_STRATEGIES = [
+    "none", "minimal", "corr_yshift", "dasi_m2", "desi_y1", "desi_y3"]
```
andreicuceu, line 86: Should we remove corr_yshift from this list?

andreicuceu:
Hi @iprafols, I updated picca_export to support m2, y1 and y3 blinding. Please have a look. This code still needs to be tested once we have the other parts working. 

Note that I created a blinding template for Y3 as well, but this is the same BAO-only blinding as before. As full-shape is part of the key project in Y3, we should be careful with this until we actually have a proper Y3 blinding scheme in place. 

So for now I just added a line in fitter2 to break if anyone tries to run fits on files with the desi_y3 flag. Vega has some capabilities to blind full-shape or making sure no one tries to fit fsigma8/full-shape AP). But it remains to be seen what we're actually gonna do with the blinding. So until we figure that out, I'm just directing people to use Vega if they need to do any fits on Y3 data, and in Vega I'll just add a check to make sure no-one is fitting any cosmology if run on files with the desi_y3 flag.

andreicuceu:
Also, the m2 blinding templates store at NERSC are not actually called m2, but their old names. So this code would not work right now. We'll need to rename them or duplicate with new names right before pushing this to master. 

If we rename, the old files would not exist, so old versions of picca would always break and anyone would have to rerun deltas and correlations with the new flags to be able to export. 
If we duplicate and rename, old and new versions of picca would both work, but this is also dangerous as people would still be able to run on Y1 data with EDR blinding.

Which do you think we should do?

py/picca/delta_extraction/data_catalogues/desi_data.py
```diff
@@ -185,30 +183,37 @@ def set_blinding(self, is_mock, is_sv):
...
-
-        is_sv: boolean
-        True if reading SV data only, False otherwise
         """
+        if all(self.catalogue["LASTNIGHT"] < 20210801):
+            blinding_strategy = "desi_m2"
+        elif all(self.catalogue["LASTNIGHT"] < 20220801):
+            blinding_strategy = "desi_y1"
+        else:
+            blinding_strategy = "desi-y3"
```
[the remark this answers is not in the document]
iprafols: changed now

py/picca/delta_extraction/utils.py
iprafols: can do, but then we won't be able to run with old deltas (see the other thread)

iprafols:
Hi @andreicuceu 

I have reviewed your changes and they make sense. Also makes sense to crush fitter2 for desi_y3 and have it managed by Vega. We still have a bit of time until we need to think what is the best blinding strategy for desi_y3 and this makes sure that we don't screw up before deciding on it.

About the template names, I'd advocate for renaming. Especially if we remove `corr_yshift` from the accepted blinding strategies. This will force people to update their version of picca. Let's discuss this on the picca task force telecon today.

iprafols:
All tests are passing now except for 3. These are related to the different behaviour of the code: for SV data, we are now adding the `desi-m2` blinding tag whereas before we had `none`. There are two ways to solve this. One option is to add another date in `set_blinding` saying that if LASTNIGHT is early enough (which date do we choose?) then we engage the blinding strategy `none`. The other option is to update the test suite. 

I don't think there is much impact irrespective of what we choose, since we will be unblinding EDR data soon enough, but still, I think we need to discuss which path to follow

py/picca/delta_extraction/data_catalogues/desi_data.py
```diff
@@ -27,7 +28,7 @@
 defaults = update_default_options(defaults, {
     "delta lambda": 0.8,
     "delta log lambda": 3e-4,
-    "blinding": "corr_yshift",
+    "blinding": "none",
```
andreufont, line 31: is it dangerous to have "none" as default here?

andreufont: nevermind, I think this is fine

py/picca/delta_extraction/data_catalogues/desi_data.py
```diff
@@ -196,19 +191,32 @@ def set_blinding(self, is_mock, is_sv):
...
-                                    "blinded. 'none' blinding engaged")
-                self.blinding = "none"
-        # TODO: remove this when we are ready to unblind
         else:
-            if self.blinding != "corr_yshift":
-                self.logger.warning(f"Selected blinding, {self.blinding} is "
-                                    "being ignored as data should be blinded. "
-                                    "'corr_yshift' blinding engaged")
-                self.blinding = "corr_yshift"
+            if all(self.catalogue["LASTNIGHT"] < 20210520):
```
andreufont, line 195: I just checked that the "main" survey started in May 14th 2021, so you could use here `20210514`

py/picca/delta_extraction/data_catalogues/desi_data.py
iprafols: we can add another one, but then which one? I added this one in order not to have to keep changing it. In any case, I'm considering removing this option and just figuring out which blinding to set directly

py/picca/delta_extraction/data_catalogues/desi_tile.py
```diff
@@ -248,7 +239,7 @@ def read_file(self, filename, catalogue):
         if "cumulative" in self.input_directory:
             select = ((catalogue['TILEID'] == tile_spec) &
                       (catalogue['PETAL_LOC'] == petal_spec) &
-                      (catalogue['LAST_NIGHT'] == night_spec))
+                      (catalogue['LASTNIGHT'] == night_spec))
         else:
             select = ((catalogue['TILEID'] == tile_spec) &
                       (catalogue['PETAL_LOC'] == petal_spec) &
```
andreufont, line 245: What would happen if "cumulative" was not declared? It looks like "LASTNIGHT" would not be defined and the code would crash later? 

Do the "daily catalogues" have "cumulative" folders? If so, why do we need the second option?

py/picca/delta_extraction/quasar_catalogues/desi_quasar_catalogue.py
```diff
@@ -126,29 +126,49 @@ def add_healpix(self):
...
+                raise QuasarCatalogueError(
+                    f"Missing required column {col} in quasar catalogue")
+
+        # optional columns
         if 'TILEID' in catalogue.colnames:
             keep_columns += ['TILEID', 'PETAL_LOC']
+            if 'PETAL_LOC' not in catalogue.colnames:
+                raise QuasarCatalogueError(
+                    "When TILEID is in the catalogue, PETAL_LOC is also "
+                    "expected to be present but it is not.")
```
andreufont, line 167: When does this happen? No harm to have the check, but I'd be surprise that this ever happens.

py/picca/delta_extraction/utils.py
```diff
@@ -82,8 +82,10 @@
     "LY10": 919.3514,
 }
 
-ACCEPTED_BLINDING_STRATEGIES = ["none", "minimal", "corr_yshift"]
-
+ACCEPTED_BLINDING_STRATEGIES = [
+    "none", "minimal", "desi_m2", "desi_y1", "desi_y3"]
```
andreufont, line 86: When/where is the "minimal" strategy used? Should we remove it?

andreufont (commented):
Looks good, I also have minor comments on desi_tile and desi_data. Let's discuss them at the telecon.

py/picca/delta_extraction/data_catalogues/desi_data.py
andreufont: Right, I think it might be better to not have "blinding" as an option, but rather as something that is set internally

iprafols: blinding option was removed

py/picca/delta_extraction/data_catalogues/desi_data.py
```diff
@@ -196,19 +191,32 @@ def set_blinding(self, is_mock, is_sv):
...
+                        f"Selected blinding, {self.blinding} is being ignored. "
+                        f"Loaded data should use '{blinding_strategy}' instead."
+                        f"'{blinding_strategy}' blinding engaged")
+                    self.blinding = blinding_strategy
+                elif blinding_strategy not in UNBLINDABLE_STRATEGIES:
+                    self.logger.warning(
+                        f"Selected blinding, {self.blinding} is being ignored "
+                        f"as loaded data should be blinded."
+                        f"'{blinding_strategy}' blinding engaged")
+                    self.blinding = blinding_strategy
```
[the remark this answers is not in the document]
iprafols: this was simplified now that blinding option is not present

py/picca/delta_extraction/data_catalogues/desi_data.py
iprafols: changed the date

py/picca/delta_extraction/quasar_catalogues/desi_quasar_catalogue.py
iprafols: I'm not sure if this will ever happen. When I changed LAST_NIGHT to LASTNIGHT I saw that we were not checking that the selected columns were actually present so I added the checks

py/picca/delta_extraction/utils.py
iprafols: I guess we can remove it, given that we also removed the corr_yshift and that we are not allowing the user to choose

py/picca/delta_extraction/data_catalogues/desi_data.py
```diff
@@ -138,14 +140,9 @@ def __parse_config(self, config):
...
-        if self.blinding is None:
-            raise DataError("Missing argument 'blinding' required by DesiData")
-        if self.blinding not in ACCEPTED_BLINDING_STRATEGIES:
-            raise DataError(
-                "Unrecognized blinding strategy. Accepted strategies "
-                f"are {ACCEPTED_BLINDING_STRATEGIES}. "
-                f"Found '{self.blinding}'")
+        self.unblind = config.getboolean("unblind")
+        if self.unblind is None:
+            raise DataError("Missing argument 'unblind' required by DesiData")
```
andreufont, line 145: @iprafols - Why do we need to specify unblind? I'd imagine it is ok to run with `unblind=None` for now, isn't it?

iprafols: When run with the scripts its value is loaded to False by default, the check is there in case the class is used in a notebook or similar to make sure that the default value is properly loaded when the variable is not specified

py/picca/delta_extraction/data_catalogues/desi_tile.py
andreufont: @iprafols - this is still pending

iprafols: Here the usage of LASTNIGHT is unrelated to that of the blinding. 
Here we are not defining column LASTNIGHT. Instead, we are using it to figure out in which file the spectra are saved. If "cumulative" is not declared, then this is not needed as the spectra are not coadded and thus they are saved in the nights they were observed. 
Thus, I think there is nothing to fix here.

andreufont (approved):
I think we are good to go, although I have not checked this very last version of the branch.

andreufont:
@julienguy , @Andrea-MG  - could you do "git pull" on this new-blinding branch, reinstall, and try to run again picca_delta_extraction? There have been minor changes in the last couple of hours.

julienguy:
I am already up to date in the branch (at Tue Dec 20 09:43:09 PST 2022). I am running `picca_delta_extraction.py`.

andreufont:
Great! Maybe @theabbybault , @HiramHerrera or @cramirezpe can also try the new-blinding branch on their FuGu analyses?

julienguy:
Successful run of `picca_delta_extraction.py` with this new-blinding branch.

andreufont (approved):
Looks good!

PR #953, opened 2022-12-21
Pylint update

iprafols:
From the discussion on the telecon on Monday, we realized there was some issue in the way we were running pylint. The problem was that base files were not being analysed. I fixed that issue (and updated pylint version while at it) and linted the code.


iprafols:
Once this is merged, we will be able to fix the current pylint problem of the pk1d_post_process branch

Waelthus (approved):
Ok with me, this will enable stricter linting and a more recent pylint version. Hope that does not interfere negatively with other PRs as well.

All changes except `pylint.yml` and `pylintrc` are just reformatting due to updated linting, right?

iprafols:
This should not affect with the other PR except for the Pk1d, where it should solve some of the testing issues

Issue #802, opened 2021-09-29
Skip blinding when working with MgII and CIV forests


andreufont:
Right now the blinding strategy is applied to any deltas coming from main-survey data. However, we shouldn't need to blind non-Lya absorption, and I guess soon we might start to look at metal cross-correlations.

I'd suggest to only blind in picca_deltas (in the sense of triggering the blind='minimal' or blind='corr_yshift' labels) when io.read_data returns a non-zero blinding *AND* (lambda_rest_min < lambda_rest_lya).

PR #954, opened 2022-12-21
metal forests not blinded

iprafols:
This PR addresses issue #802  and makes the code not blind metal forests

### DIFF
--- a/py/picca/delta_extraction/data_catalogues/desi_data.py
+++ b/py/picca/delta_extraction/data_catalogues/desi_data.py
@@ -184,7 +184,7 @@ class DesiData(Data):
         if is_mock:
             self.blinding = "none"
         # do not blind metal forests (not lya)
-        elif Forest.log_lambda_rest_frame_grid[0] < ABSORBER_IGM["LYA"]:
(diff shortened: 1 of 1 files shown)


py/picca/delta_extraction/data_catalogues/desi_data.py
```diff
@@ -180,13 +180,13 @@ def set_blinding(self, is_mock):
...
         if is_mock:
-            if self.blinding != "none":  # pragma: no branch
-                self.logger.warning(f"Selected blinding, {self.blinding} is "
-                                    "being ignored as mocks should not be "
-                                    "blinded. 'none' blinding engaged")
-                self.blinding = "none"
+            self.blinding = "none"
+        # do not blind metal forests (not lya)
+        elif Forest.log_lambda_rest_frame_grid[0] < ABSORBER_IGM["LYA"]:
+            self.blinding = "none"
```
andreufont, line 188: Wait, I think this should be the opposite! We should blind rest-frame grids < 1215.67

iprafols: ups, you're right. I have fixed it

andreufont (approved):
This looks good.

PR #955, opened 2022-12-22
ImageHDU as default format

cramirezpe:
Had to change tests to force old format on them.


Waelthus (approved):
This looks ok to me and changes most defaults to the new format. Maybe update the tutorial files as well.
Tests seem to run through fine, so those have been added in earlier PRs for the new delta format?

cramirezpe:
I've updated the tutorial.

Tests that were made using the old method are still there, I've just changed the config file so they continue using the old BinTableHDU method. For ImageHDU, we generated a specific test. 

So, most of the tests are still using BinTableHDU. If we are going to keep it indefinitely this should work.

Waelthus:
Yes, this should be ok as long as we have at least one test checking that the default format works ok...
Elsewise, I think the "examples" files need an update as well (i.e. the Pk1d versions there should use the old format, the BAO versions the new one)...

iprafols (approved):

PR #958, opened 2022-12-22
Use autoyapf to automatically format code on push

Waelthus:
This is a test if autoyapf could be of use to us


Waelthus:
the restart of actions after the yapfing commits might be complicated, see: https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#restrictions-on-repository-forks

Waelthus:
One further note regarding this: When autoyapf does not change anything, no commit is done and thus tests are done correctly in that case. If anything is committed after the autoyapfing, they will be run again, so the issue is only if the last commit is a change by autoyapf (in which case there are the workarounds from the link above...)

.github/workflows/autoyapf.yml
```diff
@@ -1,8 +1,33 @@
...
+          python -m pip install --upgrade pip
+          if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
+          pip install yapf toml
+      - name: autoyapf
+        id: autoyapf
+        run: |
+          yapf --style google --recursive --in-place .
+      - name: Check for modified files
+        id: git-check
+        run: echo "modified=$(if git diff-index --quiet HEAD --; then echo "false"; else echo "true"; fi)" >> $GITHUB_ENV
```
iprafols, line 25: I'm wondering if it might be better to just fail (instead of creating commits we are unaware of). I'm thinking of the case where yapft or pylint update in a way yapft creates problems for pylint and thus we end up in a test-failing bug without being aware

iprafols (dismissed):

Waelthus:
I'd say we use this PR as a testbed for strategies regarding code formatters for now and open a new one with actual changes (and a seperate one with the initial application of the formatter) once we know what to do

PR #965, opened 2023-01-09
fix so that we can run on eBOSS mocks from quickquasrs

iprafols:
This PR comes from an interaction via email with Jim Rich. 
He reported an issue when running picca_dmat on Saclay mocks. These mocks are computed with quickquasars and thus have the file structure of DESI mocks (though they are eBOSS mocks and so they have THING_ID instead of TARGETID). This requires one to use `--mode desi_mocks` which raises an issue later on. This PR fixes the issue by using TARGETID when present (and not when mode is set to desi)

### DIFF
--- a/py/picca/io.py
+++ b/py/picca/io.py
@@ -1419,7 +1419,7 @@ def read_delta_file(filename, rebin_factor=None):
     """Extracts deltas from a single file.
     Args:
         filename: str
-            Path to the file to read        
+            Path to the file to read
         rebin_factor: int - default: None
             Factor to rebin the lambda grid by. If None, no rebinning is done.
     Returns:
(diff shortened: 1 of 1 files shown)


Waelthus:
I think this is very similar to what we did in the early times before fully switching to DESI conventions. Can be merged if it doesn't break anything. I guess we'll figure that out easiest after merging (and if there's issues most of the main analyses should be run with a release now, so small risk if things go wrong)

Waelthus (approved):

Waelthus:
also: is there any overlap between the problem here and #841 ?

iprafols:
Possibly, though I don't remember exactly what was the motivation for that draft PR. We should ask @alxogm to be sure

Waelthus:
Overall, I'd say that ideally we'd fix the convention used in the Saclay mocks, or create a mode for that setup that is correctly traced through the code.
But if this change does not destroy any analysis (and does not have side effects such as accidental unblinding), I'm fine with the change

iprafols:
let's do this then

PR #967, opened 2023-01-17
Fix deltas rebin

cramirezpe:
Previous implementation would fail if one of the first two pixels in the deltas field is masked.

Now, dwave (width of deltas) is read from the header of the deltas file, avoiding this problem. 

It works with both ImageHDU and TableHDU formats, but only for linear deltas (as before).

### DIFF
--- a/py/picca/data.py
+++ b/py/picca/data.py
@@ -958,6 +958,7 @@ class Delta(QSO):
             Neighbouring deltas/quasars
         fname: string or None
             String identifying Delta as part of a group
+        delta_log_lambda
 
     Methods:
         __init__: Initializes class instances.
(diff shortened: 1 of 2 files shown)


py/picca/data.py
```diff
@@ -958,6 +958,7 @@ class Delta(QSO):
             Neighbouring deltas/quasars
         fname: string or None
             String identifying Delta as part of a group
+        delta_log_lambda
```
iprafols, line 961: add explanation and type

iprafols:
just a minor change, otherwise it's fine

py/picca/data.py
cramirezpe: Sorry I copied this line when it was already there.

cramirezpe:
I modified the hdul.close() position as well.

iprafols (approved):

PR #968, opened 2023-01-18
Pk1d zchunks

armengau:
New option in pk1d_postprocess: --apply_z_weights (default False).
If set, each FFT chunck at redshift z contributes 2 nearby z bins, with a triangle-shape weight function identical to that used by the QMLE.
Tests with the fujilupe mocks show the impact on the measured P1D is less than 1%.
- The zmin and zmax redshift bins must necessarily be handled differently
- This PR also includes a minor bug fix (header in output P1D FITS file).

### DIFF
--- a/py/picca/pk1d/postproc_pk1d.py
+++ b/py/picca/pk1d/postproc_pk1d.py
@@ -244,13 +244,13 @@ def compute_mean_pk1d(p1d_table,
             if apply_z_weights:  # special chunk selection in that case
                 delta_z = zbin_centers[1:]-zbin_centers[:-1]
                 if not np.allclose(delta_z, delta_z[0], atol=1.e-3):
-                    raise ValueError("Option apply_z_weights is set: redshift bins should all have equal widths.")
+                    raise ValueError("z bins should have equal widths with apply_z_weights.")
                 delta_z = delta_z[0]
 
                 select = (p1d_table['k'] < kbin_edges[ikbin + 1]) & (
                             p1d_table['k'] > kbin_edges[ikbin]
                         )
-                if (izbin==0) or (izbin==nbins_z-1):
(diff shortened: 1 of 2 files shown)


py/picca/pk1d/postproc_pk1d.py
```diff
@@ -349,6 +382,9 @@ def variance_function(snr, amp, zero_point):
                     mean = np.mean(p1d_table[col][select])
```
iprafols, line 382: shouldn't this line be in an else after line 386 in the new version?

py/picca/pk1d/postproc_pk1d.py
```diff
@@ -349,6 +382,9 @@ def variance_function(snr, amp, zero_point):
...
                     # unbiased estimate: num_chunks-1
                     error = np.std(p1d_table[col][select]) / np.sqrt(num_chunks - 1)
```
iprafols, line 384: shouldn't this also be computed differently when apply_z_weights is on?

iprafols:
The changes look fine. I have spotted a couple of lines that potentially will need changes. @armengau  can you review them? Otherwise, I think the PR is ready to merge

py/picca/pk1d/postproc_pk1d.py
armengau: yes

py/picca/pk1d/postproc_pk1d.py
armengau: Yes, I computed and implemented the expression now.
Also fixed a bug (missing sqrt) in the other expression when SNR weighting is on

armengau:
I think it's ok now

iprafols (approved):

PR #969, opened 2023-01-20
Couple of very minor fixes

iprafols:


Waelthus (approved):
the new version looks fine for me, were there installation problems with the previous setup file? If yes, was the current release affected? Or has the git version in the description (or the line break) only been introduced after?

iprafols:
Here I just added the missing f in the f-string so that the commit hash appears in the description of the code. Before it just printed the code

Waelthus:
ah, right. That makes sense!

PR #971, opened 2023-01-30
added H0 argument with default Planck2015 value for the cfs, dmats, metal matrices.

calumgordon:
I've added an H0 arg in picca_cf, picca_xcf, picca_dmat, picca_xdmat, picca_metal_dmat and picca_metal_xdmat and set the default value to Planck2015. Previously we were not passing this as an argument and it was stored in picca.constants.Cosmo as H0=100.

### DIFF
--- a/bin/picca_cf.py
+++ b/bin/picca_cf.py
@@ -165,7 +165,13 @@ def main(cmdargs):
                         required=False,
                         help=('Equation of state of dark energy of fiducial '
                               'LambdaCDM cosmology'))
-
+    
+    parser.add_argument('--fid-H0',
+                        type=float,
+                        default=67.31,
+                        required=False,
+                        help=('Hubble constant of fiducial LambdaCDM cosmology'))
+    
     parser.add_argument('--no-project',
                         action='store_true',
                         required=False,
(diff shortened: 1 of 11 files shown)


iprafols (approved):

iprafols:
I just added a few "s" that were missing in the `args` variable

iprafols:
Hmmm, good point, so then maybe this change is not necessary. @andreicuceu, in Vega there was an option to work with Mpc and not Mpc/h, wasn't there? If so, how are we dealing with this in picca? Or are we not? Cause if we are not, then maybe we should do a similar change to fix that

andreicuceu:
Vega does have an option like this but not in master (it's on a separate branch), but it's a hacky after the fact way of doing it. That's why with the new lya_2pt code I'm planning to have a flag allowing both options.

andreicuceu:
I'd say if you want to make the change, then it's fine to do it, just make sure you can reproduce the previous correlations before merging

Waelthus:
I'm all for allowing analyses to use actual com. Mpc instead of Mpc/h, but agree with the above comments that
a) one should double-check that the change is done at all necessary positions and we're not ending up using a mix of unit-systems accidentally,
b) it should be clear which system is used, i.e. make sure a comment is written to the logs that tells which unit system is used and where relevant which value of H0 was assumed, and
c) we should allow using either system by a simple flag that creates the right log entry and e.g. sets H0 to 100 and throws an error if a different value has been set before

PR #972, opened 2023-01-31
True cont update

alxogm:
Added two options: 

`-- use splines ` To use mean continuum and var_lss from a spline fit saved in the raw statistics file.  
`--recompute var lss ` To use the var lss provided by the raw statistics file instead of the re-computed one. Set as True by default. 

Both additions motivated by the work with Andrea on CIV forest, as we were having issues getting a good true continuum CF.


py/picca/delta_extraction/expected_fluxes/true_continuum.py
```diff
@@ -141,10 +145,11 @@ def compute_expected_flux(self, forests):
         # the might be some small changes in the var_lss compared to the read
         # values due to some smoothing of the forests
         # thus, we recompute it from the actual deltas
-        self.compute_var_lss(forests)
-        # note that this does not change the output deltas but might slightly
-        # affect the mean continuum so we have to compute it after updating
-        # var_lss
+        if self.recompute_varlss :
```
iprafols, line 148: remove space before :

py/picca/delta_extraction/expected_fluxes/true_continuum.py
```diff
@@ -123,7 +126,8 @@ def __parse_config(self, config):
         self.raw_statistics_filename = config.get("raw statistics file")
 
         self.force_stack_delta_to_zero = config.getboolean("force stack delta to zero")
-
+        self.use_splines=config.getboolean("use splines")
```
iprafols, line 129: We should raise ExpectedFluxError when the different options are `None` if only to check that the default values are properly digested (it could not be the case if, for example, one is running from a notebook. I will add a commit to fix this

iprafols (approved):

iprafols:
It seems github actions have updated their version of pylint so I needed to fix a few minor things in some files for the pylint test to pass

PR #973, opened 2023-02-01
Change LASTNIGHT for COADD_LASTNIGHT

alxogm:
The final name for the Last night column in the per healpix catalog is COADD_LASTNIGHT. so I'm changing from LASTNIGHT to COADD_LASTNIGHT, this will require that other catalogs than the Himalayas one are updated, we can coordinate that with Abby and Lucas.  

@iprafols @julienguy do you agree or do you prefer that we test for LASTNIGHT as well?


iprafols:
Hi @alxogm with your changes we are also checking for LASTNIGHT, what we would not be checking for is LAST_NIGHT. If some of the (old) catalogues will still have LASTNIGHT, then I'd keep it. Otherwise, instead of keeping LASTNIGHT, I would go straight to replacing it there (and in the other places in the code where it appears) by COADD_LASTNIGHT

iprafols (approved):

iprafols:
Fine with me

PR #974, opened 2023-02-07
Change default redshift evolution to match QSO

andreicuceu:
When computing the cross-correlation (and xdmat, metal_xdmat, xwick, etc.) the redshift evolution parameter for discrete objects was set to 1. This PR changes it to 1.44 to match the value used for quasars in DR16. This is in line with what we do for the delta field where the default is 2.9 (value of alpha_lya).

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,10 +154,12 @@ Before submitting a PR please make sure to:
 2. Update the data model
 3. For every file you have modified run
    ```
-   yapf --style google file.py -i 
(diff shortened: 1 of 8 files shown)


iprafols (changes requested):
changes look fine but we should set up the test to keep using the old value or else update the test results so that we can pass the tests

iprafols (approved):

PR #976, opened 2023-02-16
Update picca_Pk1D.py

alxogm:
Add simple line necessary when reading deltas from eBOSS in  ascii format.

### DIFF
--- a/bin/picca_Pk1D.py
+++ b/bin/picca_Pk1D.py
@@ -86,6 +86,7 @@ def process_all_files(index_file_args):
     elif args.in_format == 'ascii':
         ascii_file = open(file, 'r')
         deltas = [Delta.from_ascii(line) for line in ascii_file]
+        running_on_raw_transmission = False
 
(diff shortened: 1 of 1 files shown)


Waelthus (approved):
Is this actually all that was missing for running on eBOSS ascii files again? We didn't really use that mode in years, so it might well be that there's undesired behaviour.
But if it works fine, I'm ok with merging.

alxogm:
The code runs, but we still have to understand the outputs. There is a new MsC student in my group with whom we wanted to try reproduce eBOSS results. 

So I think we can merge and if something arises we'll report it.

Waelthus:
Not sure if there have been any fits outputs of the eBOSS Pk1d deltas as well so that one could use the default I/O routines (if not I guess one could generate those by re-running delta_extraction on eBOSS or even the old picca_deltas on eBOSS with the fits output mode).
Merging this anyway to allow runing on ascii at all (but personally I think we should get rid of the ascii mode completely at some point).

PR #977, opened 2023-03-09
add cuts in quasar redshift

cramirezpe:
This allows for limiting the quasar redshift at the correlation level.

In the context of EDR, we decided to perform this instead of doing it at the delta extraction level.


iprafols (approved):

cramirezpe:
Thank you very much Ignasi for finding the issue.

PR #979, opened 2023-03-10
Fix data model

cramirezpe:
This standardizes output fits files for image format. 

I am waiting for #977 to be merged before adding tests, as both PR need to modify test files.

- [ ] Also data model files have to be updated, as I added the attribute "units" to the rejection_log object.
- [ ] I need to update the code in some parts, as some fitsio reading function seeem to be case sensitive: calibration_correction.py:51


iprafols:
PR #977  is already merged. I have merged master into fix_data_model so that those changes are updated here as well. Let me know when you are ready to merge this PR

py/picca/delta_extraction/expected_flux.py
```diff
@@ -350,8 +350,11 @@ def hdu_cont(self, results):
             self.get_mean_cont(Forest.log_lambda_rest_frame_grid),
             self.get_mean_cont_weight(Forest.log_lambda_rest_frame_grid),
         ],
-                      names=['loglam_rest', 'mean_cont', 'weight'],
+                      names=['LOGLAM_REST', 'MEAN_CONT', 'WEIGHT'],
+                      units=['log(Angstrom)', '10**-17 erg/(s cm2 Angstrom)', ''],
```
iprafols, line 354: I'm thinking we should load this from the `Data` class or the `Forest` class, as different surveys might use different units for the flux

iprafols (changes requested):
Looks fine, we just need to update the tests and fix one concern before we are ready to merge

py/picca/delta_extraction/expected_flux.py
cramirezpe: I have added this. Only for fluxes. At first I did it also for wavelength, but I think it makes no sense to run picca in other units (for example BAL lines are defined in Angstroms, so we will need to write an unit converter)

iprafols (approved):

PR #988, opened 2023-03-31
Fix the resolution correction based on pixel level simulation findings

corentinravoux:
change the default resolution correction and add an option to have the previous correction.


armengau (approved):
Ok for me.
Only detail: the option "pixelization_correction" should be documented in compute_correction_reso_matrix(), although it's kinda duplicate with the help provided in the main

iprafols (approved):

PR #989, opened 2023-04-04
added tests for DLA and BAL masking

iprafols:
Currently, we were testing neither the BAL masking nor the DLA masking. This PR is necessary so that we can optimize the masks with a minimum set of guarantees that we are not changing the algorithms. 

I also fixed a couple issues with mask options inheritance


py/picca/delta_extraction/masks/lines_mask.py
```diff
@@ -5,9 +5,13 @@
...
 from picca.delta_extraction.mask import Mask
+from picca.delta_extraction.mask import ( # pylint: disable=unused-import
+    accepted_options, defaults)
+from picca.delta_extraction.utils import update_accepted_options
 
-accepted_options = ["filename", "keep pixels"]
-
+accepted_options = update_accepted_options(accepted_options, [
+    "filename"
+])
```
Waelthus, line 14: keep_pixels is always added in the mask.py now, right?

py/picca/tests/delta_extraction/mask_tests.py
```diff
@@ -114,24 +117,369 @@ def test_absorber_mask(self):
         reset_logger()
         self.compare_ascii(test_file, out_file)
 
-    def test_dla_mask(self):
+    def test_bal_mask_ai_remove(self):
```
Waelthus, line 120: the added tests should only add to existing tests, so better to have them added. Did not check all the internals...

Waelthus (approved):
Looks good to me regarding the minor changes outside the test. And as long as the routines are now actually tested and match the expectations I'm fine with the additions as well without having them checked in all detail

py/picca/delta_extraction/masks/lines_mask.py
iprafols: yes, it was read from there before too. We kind of missed it when all these changes were introduced, and it makes more sense that this option is defined here, where it is actually used, as opposed to it being defined in every child class

py/picca/tests/delta_extraction/mask_tests.py
iprafols: I added them, I just renamed the `test_dla_mask` to `test_dla_remove` that's why this line is flagged as removed

PR #992, opened 2023-04-11
Optimize find bins

iprafols:
I implemented Naim's suggestion. Did not have time to check the speed boost for this change, though. 
Note that there is a minor difference in the behaviour: when the wavelength is exactly in the middle of two bins, the old behaviour was to use the lower bin and the new behaviour uses the upper bin. However, this does not seem to be a relevant effect (the tests are passing, at least locally)

Note that I did discover some potential issues with the way we are loading the data. Fixing it will improve the speed boost as a couple of searches can be dropped.

### DIFF
--- a/py/picca/delta_extraction/utils.py
+++ b/py/picca/delta_extraction/utils.py
@@ -159,11 +159,11 @@ def find_bins(original_array, grid_array, wave_solution):
     """
     if wave_solution == "log":
         step = grid_array[1] - grid_array[0]
-        found_bin_new = np.floor((original_array - grid_array[0]) / step + 0.5).astype(np.int64)
+        found_bin = np.floor((original_array - grid_array[0]) / step + 0.5).astype(np.int64)
     elif wave_solution == "lin":
         aux = 10**grid_array[0]
         step = 10**grid_array[1] - aux
-        found_bin_new = np.floor((10**original_array - aux) / step + 0.5).astype(np.int64)
+        found_bin = np.floor((10**original_array - aux) / step + 0.5).astype(np.int64)
     else:  # pragma: no cover
         raise DeltaExtractionError(
             "Error in function find_bins from py/picca/delta_extraction/utils.py"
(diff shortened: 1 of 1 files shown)


iprafols:
ok, I think I found the issue. Apparently it has something to do with the definition of some arrays in Dr16ExpectedFlux. See #993  for details

py/picca/delta_extraction/utils.py
```diff
@@ -158,26 +158,26 @@ def find_bins(original_array, grid_array, wave_solution):
...
     if wave_solution == "log":
-        pass
+        step = grid_array[1] - grid_array[0]
+        found_bin = np.floor((original_array - grid_array[0]) / step + 0.5).astype(np.int64)
     elif wave_solution == "lin":
-        original_array = 10**original_array
-        grid_array = 10**grid_array
+        aux = 10**grid_array[0]
+        step = 10**grid_array[1] - aux
+        found_bin = np.floor((10**original_array - aux) / step + 0.5).astype(np.int64)
```
p-slash, line 166: You don't need to floor.

iprafols: true, the `astype` is doing the floor on its own. I added a commit to remove them

iprafols:
I'll set them running on fugu when Perlmutter is back up and running

Waelthus:
Looks good to me, too. Does this mean we're always handing uniform arrays in lin or log to find_bins, so the previous considerations of having a general function here, were just irrelevant? And can we interpret the running tests as a "results won't change anywhere"?

iprafols:
Seems like that is the case, I checked on a Fugu run and the deltas are exactly the same. I also checked the time improvement: 

On the master branch (showing only relevant lines):
```
[INFO]: Time spent reading data: 137.78101706504822
[INFO]: Time spent applying corrections: 4.229247331619263
[INFO]: Time spent applying masks: 1470.6445696353912
[INFO]: Time spent computing the mean expected flux: 474.09572672843933
[INFO]: Time spent extracting deltas: 1.6446754932403564
[INFO]: Time spent saving deltas: 25.603065490722656
[INFO]: Total time elapsed: 2192.240534067154
```

On this branch
```
[INFO]: Time spent reading data: 257.25135374069214
[INFO]: Time spent applying corrections: 4.226393461227417
[INFO]: Time spent applying masks: 1479.5172002315521
[INFO]: Time spent computing the mean expected flux: 260.9304323196411
[INFO]: Time spent extracting deltas: 1.6586024761199951
[INFO]: Time spent saving deltas: 26.908676862716675
[INFO]: Total time elapsed: 2108.7183299064636
```

There seems to be an improvement in the computation of the mean expected flux, but that is compensated by an increase in the time taken to read the data. Overall, there is a marginal gain. 

I think we are not seeing an improvement since we need to call the `np.clip` function at the end. The function will be definitely faster if we can skip it by enforcing the actual wavelength ranges. Let's discuss this on the telecon on Monday before we decide whether or not to merge.

p-slash:
Very surprised about the read data part of the timing. If anything, I would have expected it to get shorter. It could be that Perlmutter IO was not kind to that specific run. Getting rid of np.clip should definitely help overall though. The less if statements we have, the better

iprafols:
Reading data also includes the forest rebinning, which uses the find_bins function.

Waelthus (approved):
Looks fine to me

Issue #993, opened 2023-04-11
Dr16ExpectedFlux log_lambda_var_func outside of range


iprafols:
The way `log_lambda_var_func` is computed in method `_initialize_variance_wavelength_array` of `ExpectedFlux` (line 109) produces wavelengths that are outside the nominal valid range. This will cause issues with the function find_bins if a direct computation is used (see issue #978 and PR #992 ).  

For reproducibility/to give more details, I added the following piece of code at the end of the function:
```
        print("self.log_lambda_rest_frame_grid")
        print(self.log_lambda_var_func_grid)
        print("min, max of Forest.log_lambda_grid")
        print(Forest.log_lambda_grid[0], Forest.log_lambda_grid[-1])
        print("size of Forest.log_lambda_grid")
        print(Forest.log_lambda_grid.size)
        print("self.num_bins_variance")
        print(self.num_bins_variance)
        print("resize factor")
        resize = len(Forest.log_lambda_grid)/self.num_bins_variance
        print(resize)
        print("Proposed change")
        print(Forest.log_lambda_grid[::int(resize)])
```
Running the test `scriptsTest.test_delta_calib` yields:
```
self.log_lambda_rest_frame_grid
[3.5638775 3.5790275 3.5941775 3.6093275 3.6244775 3.6396275 3.6547775
 3.6699275 3.6850775 3.7002275 3.7153775 3.7305275 3.7456775 3.7608275
 3.7759775 3.7911275 3.8062775 3.8214275 3.8365775 3.8517275]
min, max of Forest.log_lambda_grid
3.5563025007672873 3.859302500767478
size of Forest.log_lambda_grid
1011
self.num_bins_variance
20
resize factor
50.55
Proposed change
[3.5563025 3.5713025 3.5863025 3.6013025 3.6163025 3.6313025 3.6463025
 3.6613025 3.6763025 3.6913025 3.7063025 3.7213025 3.7363025 3.7513025
 3.7663025 3.7813025 3.7963025 3.8113025 3.8263025 3.8413025 3.8563025]
```
Fixing this is quite simple, we just need to replace the entire function by 
```
        resize = len(Forest.log_lambda_grid)/self.num_bins_variance
        self.log_lambda_var_func = Forest.log_lambda_grid[::int(resize)]
```
However, this implies a whole update of the test suite. I don't think this will have any relevant impact on the final deltas, but they will likely change slightly. Thus, I wanted to discuss this before going ahead

PR #994, opened 2023-04-19
fixed issue 993

iprafols:
Here I changed the method `_initialize_variance_wavelength_array` from `ExpectedFlux` as suggested in issue #993. 

The impact of this change is that the functions eta, var_lss (and fudge) are computed in a slightly different wavelength grid. The new grid selects points from `Forest.log_lambda_grid` thus enforcing the actual wavelength cuts, whereas before we were having an array that was mildly out of bounds. This is an issue if we want to use direct computation in the function `find_bins`. 

As expected, I also needed to update the test suite. However, given that the fits to compute eta, var_ls, and fudge in the test do not converge (since we do not reach the minimum number of quasars), the changes are limited to the remapping of the wavelength grid for the VAR_STATS HDU in the different files. 

The exception to this is the file `true_iter_out_prefix_compute_expected_flux_log.fits.gz`. For this file, the change in `var_lss` is larger than expected (see red vs blue line in the plot). I investigated the origin of this change and I found out that in the `TrueContinuum` class, `sigma_lss` is computed in the `Forest.log_lambda_grid` grid and then interpolated to the `self.log_lambda_var_func_grid` grid before saving. This "native" grid is shown as the orange line in the plot. The change observed in the VAR_STATS HDU is associated with the noisy nature of this function. This might be an issue for the runs using `TrueContinuum` expected flux but is outside the scope of this PR (I will now file this as an issue). 

download-1

Because the fits to compute eta, var_ls, and fudge in the test do not converge, the impact of this change in the final deltas is not assessed by the automatic tests. I expect this change to be small, but I would like to verify this on a fugu run before merging this branch


iprafols:
As expected the changes are small: 

changes

I think we are now ready to merge

py/picca/delta_extraction/expected_flux.py
```diff
@@ -112,33 +112,9 @@ def _initialize_variance_wavelength_array(self):
...
-
-        # TODO: Replace the if/else block above by something like the commented
-        # block below. We need to check the impact of doing this on the final
-        # deltas first (eta, var_lss and fudge will be differently sampled).
-        #start of commented block
-        #resize = len(Forest.log_lambda_grid)/self.num_bins_variance
-        #print(resize)
-        #self.log_lambda_var_func_grid = Forest.log_lambda_grid[::int(resize)]
-        #end of commented block
+        resize = len(Forest.log_lambda_grid)/self.num_bins_variance
```
Waelthus, line 115: could use // integer division from the beginning to avoid the cast to int in the next line...

py/picca/delta_extraction/expected_flux.py
```diff
@@ -112,33 +112,9 @@ def _initialize_variance_wavelength_array(self):
...
+        self.log_lambda_var_func_grid = Forest.log_lambda_grid[::int(resize)]
```
Waelthus, line 116: wouldn't that be assymetric? I.e.. we're discarding on the high-lambda end always? I guess that's fine given that most observations are at the low-redshift end.
But wouldn't we actually want something like `np.logspace(np.min(Forest.log_lambda_grid),np.max(Forest.log_lambda_grid),self.num_bins_variance [maybe +1?])`?

Also the previous version had the variance bins defined to be bin centers, would we now define them as bin edges? Given that the lowest wavelength point is always included? Maybe that is the correct way  now, given that the find_bins function used to select the closest entries and now selects entries that have higher lambda only (is this actually what it does now?).

Waelthus (commented):
It's nice that results only change marginally, I added one comment regarding the definition of variance bins...

py/picca/delta_extraction/expected_flux.py
iprafols: good point, I pushed a commit to fix this

iprafols:
Replaced by #1000

PR #996, opened 2023-04-19
Cosmo class fix

iprafols:
This is a quick fix for the `Cosmo` class to compute distances. This solves the immediate problem of reporting a value of H0=100 and having it as an optional parameter to the `Cosmo` class. This value is needed so that the return units are Mpc/h due to some obscure programming of this class. This should be fixed at some point.


Waelthus (approved):
Sounds like a reasonable choice for the moment. Should re-iterate on it in the future when implementing the cosmological model more properly

PR #999, opened 2023-05-01
Fix raw io

andreicuceu:
Since the rebin option has been introduced in picca correlations, it's looking for the WAVE_SOLUTION field which already exists in delta extraction outputs. This adds that field to raw deltas as well.


Waelthus (approved):
Looks fine to me

PR #1000, opened 2023-05-16
Optimize find bins fix issue 993

iprafols:
This PR fixes issue #993  and optimized function find_bins to use direct calculation. The results are different since I changed the binning of the wavelength grids in the computation of the expected flux. Thus, I updated the test suite. Below I provide a comparison of the results of this branch with those from master with a fugu run. 

Timing comparison (over 3 runs):
Captura de pantalla 2023-05-16 a les 11 51 58

Variance statistics comparison:
download

Delta stack comparison:
download-1

Quasar continuum comparison:
download-2

Fits metadata comparison:
download-3

Lya autocorrelation comparison:
download-4
download-5


iprafols:
Note that tests are failing due to some updates of numba and/or numpy in the github actions server (and are unrelated to the changes here)

Waelthus:
I think it's  #1001. I'll have a look through all positions where this happens and will open another PR

py/picca/delta_extraction/utils.py
```diff
@@ -147,37 +148,40 @@ def find_bins(original_array, grid_array, wave_solution):
...
-        min_dist = np.finfo(np.float64).max
-        for index2 in range(grid_array_size):
-            dist = np.abs(grid_array[index2] - original_array[index1])
-            if dist < min_dist:
-                min_dist = dist
-                found_bin[index1] = index2
-            else:
-                break
+    step = grid_array[1] - grid_array[0]
+    found_bin = ((original_array - grid_array[0]) / step + 0.5).astype(np.int64)
```
Waelthus, line 184: Wouldn't this give an array that is displaced by exactly half a pixel and then cast to int? Is this for avoiding numerical issues from the division? Why not just use `(original - grid) // step` in that case?

py/picca/tests/delta_extraction/astronomical_object_tests.py
```diff
@@ -1668,14 +1668,14 @@ def test_forest_set_class_variables(self):
...
-            3.7363025, 3.7413025
+            3.7363025, 3.7413025,
         ])
         self.assertTrue(np.allclose(Forest.log_lambda_grid, log_lambda_grid))
 
         log_lambda_rest_frame_grid = np.array([
-            3.01953334, 3.02453334, 3.02953334, 3.03453334, 3.03953334,
-            3.04453334, 3.04953334, 3.05453334, 3.05953334, 3.06453334,
-            3.06953334, 3.07453334
+            3.01703334, 3.02203334, 3.02703334, 3.03203334, 3.03703334,
```
Waelthus, line 1676: this is a change by exactly 0.0025, i.e. one half pixel, did we want to change the definition? see other file

Waelthus (commented):
looks good overall, see comments regarding definition of restframe grid. Will still need to work on fixing version issues, see #1002

Waelthus (approved):

PR #1002, opened 2023-05-17
fix deprecated numpy/numba usages

### LINKED ISSUE #1001: Deprecation warnings/errors by numpy and numba
The usage of some data-type aliases, e.g. `np.int`,  has been deprecated in numpy, we need to change those in picca accordingly to allow running and testing on current numpy versions:

> module 'numpy' has no attribute 'int'.
> `np.int` was a deprecated alias for the builtin `int`. To avoid this error in existing code, use `int` by itself. Doing this will not modify any behavior and is safe. When replacing `np.int`, you may wish to use e.g. `np.int64` or `np.int32` to specify the precision. If you wish to review your current use, check the release note link for additional information.
> The aliases was originally deprecated in NumPy 1.20; for more details and guidance see the original release note at:
>     https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations

Also the usage of numba jit compilation when the `nopython` option is not supplied will change its default from `nopython=False` to `nopython=True`:

>   /home/runner/work/picca/picca/py/picca/xcf.py:1032: NumbaDeprecationWarning: The 'nopython' keyword argument was not supplied to the 'numba.jit' decorator. The implicit default value for this argument is currently False, but it will be changed to True in Numba 0.59.0. See https://numba.readthedocs.io/en/stable/reference/deprecation.html#deprecation-of-object-mode-fall-back-behaviour-when-using-jit for details.

Waelthus:
Here, we're fixing some deprecated behaviour and add warnings for usage of slow functions, e.g. regarding the wick covariances...
this should fix #1001


Waelthus:
Note that tests ran with `numpy==1.23.5` and `numba==0.56.4`, but not on `1.24.3` and `0.57.0`. So issues appear somewhere between those

Waelthus:
So 8 Arrays and thus fitting a (1,8) shape. But why was there no complaint before? Was there a change in numpy regarding treatment of those? 
I guess we should just use a list of numpy arrays then.

iprafols:
My take on this is that numpy is now not allowing arrays with elements with different shapes (which kind of makes sense). Not sure why it worked before. I modified the way we are unpacking the list and things should be fine now

Waelthus:
Personally, I I'd probably have unpacked those via zip(dmat_data) or sth... Instead of doing a ton of list comprehensions... 
But fine with me for now. Somebody else needs to accept though...

Waelthus:
Ok, from my point we could merge this. I guess that all the newly @njit-ted functions with warnings will actually break if called now, so we should probably just remove them without the deprecation warning. Or use the @jit still, have our own warnings until numba does its change and remove then.
I guess just removing would be fine given that very likely no-one uses those right now...

iprafols:
I have removed them

iprafols (approved):

PR #1003, opened 2023-05-25
Pk1d covariance improvements

corentinravoux:
Improvements of the covariance matrix calculation.
In particular, this branch can now compute the covariance matrix with the fit_snr method used to compute mean P1D.
This branch will be used to build the first covariance matrix of the EDR+M2 measurement.

I suggest using this branch to make all the improvements discussed in previous covariance matrix calculations. Do not hesitate to push on this branch if you have any changes.

I perform tests on Ohio EDR-mocks to assess the quality of the covariance matrix. In particular, I compared the error computed in the average calculation (standard deviation) with the diagonal of the covariance matrix:

- For the no_weights method, the diagonal is exactly the same by construction. I also verified on paper that the two estimates are theoretically equal.

- For the fit_snr method, the error on P1D is not computed with an estimator but with a fitting. It means that theoretically, with the covariance matrix calculation implemented, the error and the diagonal of the matrix are not equal. You will see an attachment showing the difference between those two. I **choose** to renormalize the diagonal by the error to have something consistent between the error and the diagonal covariance. This choice can be discussed (for example, do we extend this normalization to the full covariance matrix?).

I will also attach the small theoretical calculations I made.


corentinravoux:
Here are my calculations, I hope it is readable:
[Covariance matrix.pdf](https://github.com/igmhub/picca/files/11563218/Covariance.matrix.pdf)

py/picca/pk1d/postproc_pk1d.py
```diff
@@ -922,39 +906,79 @@ def compute_cov(
...
-
-                    if ikbin2 != -1:
+                    if (ikbin2 != -1) & (ikbin != -1):
+                        # index of the (ikbin,ikbin2) coefficient on the top of the matrix
+                        index = (nbins_k * ikbin) + ikbin2
+                        weight = 1 / selected_variance_estimated[ipk]
+                        weight2 = 1 / selected_variance_estimated[ipk2]
+                        covariance_array[index] = covariance_array[index] + selected_pk[
+                            ipk
+                        ] * selected_pk[ipk2] * np.sqrt(weight * weight2)
```
Waelthus, line 939: why is the weighting done with a sqrt(weight *weight2) instead of with the weights directly, see other comment

Waelthus (approved):

Waelthus:
I'm merging this anyway to allow usage of the existing improvements. @corentinravoux: we'll need to think about exact definitions again at a later time and update the estimators accordingly. The need to rescale does look very weird to me...

PR #1004, opened 2023-05-25
skip apply_masks and apply_corrections when not needed

p-slash:
These changes aim to prevent the creation of spurious pools. @iprafols does this make sense? It's been a while since I look at this in detail.

### DIFF
--- a/py/picca/delta_extraction/survey.py
+++ b/py/picca/delta_extraction/survey.py
@@ -113,6 +113,9 @@ class Survey:
 
     def apply_corrections(self):
         """Apply the corrections. To be run after self.read_corrections()"""
+        if not self.corrections:
+            return
+
         t0 = time.time()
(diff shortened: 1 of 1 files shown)


iprafols:
Hi Naim,

I think this makes sense to add. However, when there are no masks, we have an empty list, so really we should check using `len(X) == 0` I'll do a commit for this

iprafols (approved):

iprafols:
@p-slash  let me know if you agree with the changes I made and then I'll merge the PR

p-slash:
I mean you didn't need to check for length explicitly. Lists and dictionaries can go into if statements to check whether they are empty or not.

iprafols:
I see, I did not know that. However, I'm not particularly in favour of doing this. Using the length makes it more explicit that these are lists and not booleans.

PR #1005, opened 2023-05-31
Fix p1d weights

armengau:
Bug fix on SNR weighting for P1D.
Weights were not correctly calculated for quantities other than P1D itself, eg. redshift.
Originally spotted by @julienguy 

Minor impact on calculated `<z>` at the postprocessing stage, see attached plot: `<z>` for different k bins for the z=2.2 redshift bin. `<z>` is increased because low-z chunks have a smaller SNR on average (at z=2.2)

bugfix_meanz


Waelthus:
So the issue here is that we previously weighted every quantity by SNR(quantity) whereas we properly should weight everything by SNR(Pk)? 
Which then indeed shifts the <z> to somewhat higher positions as we see in the plot and thus will move the overall mean z to a higher value

Waelthus (approved):

Waelthus:
The merge with master was needed to be able to pass tests. If we have automatic testing of pk1d postprocessing results, those need to be adapted as well...

armengau:
> 

2 dummy github questions:
- you merged the branch but did not delete it afterwards ? (so that when looking at the list of branches, filtering eg 'p1d', it does not appear anymore ) ?
- how did you do "Merge remote-tracking branch 'origin/master' into fix-p1d-weights" ? (some action from the github web, or a git command from your shell + push?)

iprafols:
you definitely can do this:
```
git checkout master
git pull
git checkout fix-p1d-weights
git merge master
```
or you can try to do it directly (though I'm not sure it'll work:
```
git merge origin/master
```

iprafols:
Also, yes, we usually delete the branches once we merge PRs, we already have too many branches and it is easier to maintain this way. Note, however, that the branch could be restored if necessary and that this does not remove the branch locally.

Waelthus:
Actually I did it with the graphical tools provided in vscode... Merging directly from the remote is possible if the branch has been fetched before, `git pull` (on the master branch) does nothing else then:
```
git fetch origin master
git merge origin/master
```
and the same way you can merge any remote branch into any local branch...

PR #1006, opened 2023-06-02
Add workflow to auto-submit to pypi

Waelthus:


Waelthus:
I also removed the autoyapf again, as that never really worked...

Issue #1007, opened 2023-06-07
Problem with interpolations in delta extraction


andreicuceu:
While investigating a different problem with the true continuum code (probably coming in a separate issue soon), I plotted the weights of some forests. These are coming from an uncontaminated DESI Y5 mock. I attached a figure with a few forests I've picked based on their SNR. The problem is that the high SNR forests appear to have a step function in their weights. I then overplotted 1/var_lss, from which you can see that the steps align with the bins in var_lss.

This is caused by the interpolation of var_lss which is done with interp1d using the "nearest" option instead of linear or cubic as I would have expected. So a straight forward way to address this would be to change all interpolations to cubic. This would definitely break all the tests, so I wanted to get everyone's opinion before opening a pull request.

![forest_weights](https://github.com/igmhub/picca/assets/28951059/fef18bbc-06f0-48c8-b301-56b9060ba33f)

iprafols:
Hi Andrei, 

I agree we should change this. I was planning to implement our own interpolator following @p-slash  suggestion to speed up the searches, and I can implement a cubic interpolation too. That said, I would go ahead with the changes first, as these seem simple enough and preparing a new interpolator might take a bit of time (specially since I'm busy elsewhere too)

iprafols:
Also, this seems an important fix that would break backward compatibility. Thus, I would create a new tag once this is merged.

andreicuceu:
Okey, sounds good to me. I'll make a pull request.

andreufont:
Can we close this issue, @andreicuceu ?

andreicuceu:
Yes, happy to close as this was fixed by #1008

PR #1008, opened 2023-06-08
Change delta extraction interpolations to cubic [bump minor]

andreicuceu:
This is meant to address #1007. Changing interpolations in delta extraction from nearest to cubic will break the tests, so we have to update those as well. I've left the get_num_pixels and get_valid_fit interpolations in nearest, as I'm not sure those need updating.


py/picca/delta_extraction/expected_fluxes/dr16_expected_flux.py
```diff
@@ -243,7 +243,7 @@ def _initialize_get_var_lss(self):
         self.get_var_lss = interp1d(self.log_lambda_var_func_grid,
                                     var_lss,
                                     fill_value='extrapolate',
-                                    kind='nearest')
+                                    kind='cubic')
```
iprafols, line 246: I would also change this in lines 281, 285, 620 and 624. More for consistency that this having an impact on the measurements

iprafols (changes requested):
I would add some extra changes (see my other comment). I checked that there are no more related interpolations that would need this change so after this minor thing is fixed we should be able to merge. 

I noticed that there are also some interpolations done in the corrections (e.g. calibration correction line 80). However, here the grid is much finer so the change should be much smaller. I'm fine with not changing this for now, but maybe you want to check if this would affect your measurements as well. 

Finally, I saw that there are other interpolations when computing the correlation function (for example in picca_wick but I also think we can leave them as they are.

py/picca/delta_extraction/expected_fluxes/true_continuum.py
```diff
@@ -494,12 +494,12 @@ def _check_header_consistency(_key, _test_val, atol, _consistent, _err_msg):
...
                                     var_lss,
                                     fill_value='extrapolate',
-                                    kind='nearest')
+                                    kind='cubic')
 
         self.get_mean_flux = interp1d(log_lambda,
                                       mean_flux,
                                       fill_value='extrapolate',
-                                      kind='nearest')
+                                      kind='cubic')
```
p-slash, line 502: I have a warning here. Depending on the input file and its noise levels, cubic spline can act terribly. I remember the files from rawio analysis being very noisy. Andrei, you created a smooth spline version. Has that become the norm?

andreicuceu: Good point. Yes the new raw-stats files also save a spline version, but I don't think it's the default. The current files in picca (for 2.4A and 3.2A pixelization) are old and don't have that. So I'm gonna look into updating those as well.

py/picca/delta_extraction/expected_fluxes/dr16_expected_flux.py
p-slash: "Nearest" sounds like what is aimed here. If you feel like it, you can get rid of these interpolations while you're at it. They should not be interpolations in my opinion, but I don't know how deeply they are embedded into the code.

iprafols: They are basically used in the reports in the delta_attributes files, so that we can check problems in the fits. I don't think they are used elsewhere, so yes, if we don't need it when saving these files, we might as well remove the interpolation.

andreicuceu:
True continuum problems when recomputing var lss are now detailed in a separate issue (#1009). I've also added a warning when that option is used.

I think this is all we wanted to do with this pull request, so on my side I'm happy to merge. For completeness here is a plot of the 3D auto-correlation from a DESI Y5 mock with the old and new interpolations. The two are almost identical.

corr_with_new_interp

iprafols:
Great, so now we just need to fix the tests. Note that I added a new commit to fix the failed test reports

iprafols:
Hi @andreicuceu , are there any updates on this PR?

andreicuceu:
The only remaining step is to update the automatic tests. I've already been using this branch to run all my mocks for the last few weeks, and everything looks good. I've just been too busy lately to look at the tests, and it will probably not happen before the end of the DESI meeting either.

iprafols:
I'll try to update them in between the sessions tomorrow or on Wednesday

Waelthus:
fixes #1007

Issue #1010, opened 2023-06-13
New version system


Waelthus:
From now on, we'll increase the version number at the patch level (for full version major.minor.patch) whenever we merge a set of PRs. Releases will typically only be done for new minor or major versions as before.

@p-slash: you suggested a tool to simplify this yesterday, can you post that here?

p-slash:
This is a good source for version conventions: https://semver.org/

I use the bump2version package: https://pypi.org/project/bump2version/ and https://github.com/c4urself/bump2version. Its main purpose is to increase the version number with a single command, that is to create a tag and to add a commit. I still do it manually. So for example, Ignasi or Michael can be in charge of running this command whenever necessary and push to github. Also, this package shouldn't be a part of the picca distribution.

You can see https://github.com/p-slash/qsonic/blob/main/setup.cfg for an example setup.cfg. I also store the qsonic version in https://github.com/p-slash/qsonic/blob/main/py/qsonic/__init__.py file, and refer to that in setup.cfg `version = attr: qsonic.__version__` so that pip automatically knows the version number.

Waelthus:
Ok, I'll set this up, we could in principle have a github action running bump2version and committing the version update on every push to master, I guess, than each PR is it's own patch version increase. 
We could also auto-tag which then would also auto-push to pypi and make `pip install picca` effectively get the current master, not sure if that's ideal. @iprafols?

iprafols:
I would say we can try to have an action running bump2version for the patch version increase. However, I think we would need to have the option of specifying "minor update" or "major update" instead of "patch" so that it does not increase the version number incorrectly when we have bigger changes. Maybe it's easier to run it before merging and manually decide whether it's a patch, minor or major update.

What we could do is set a GitHub action to auto-publish to PyPI whenever we do a tag (a major or minor update, but not a patch). 

I think this would make more sense than publishing to Pypi with every patch, and that allows us to have the master branch for development, as it's supposed to happen

Waelthus:
In principle we could let the action check the commit msg of the merge for cases where we either want to disable the action and do this manually for a release, or to specify which version should be bumped...
I read your answer as a "let's just tag major versions" so we disable the automatic tagging feature for patches. Pushing to Pypi is anyway only done on tags...

iprafols:
The PR #1011  did not fix this completely.  The problem is that the action is not allowed to push to master (which is protected). Maybe we can set it so that the push is made to the branch being merged and not to master, and have this occur right before the merge?

[...]

PR #1014, opened 2023-06-15
adding merge-groups to actions

Waelthus:


Waelthus:
I activated merge queues with a queue length of 1, so we can run the github action for version bumping on that which would then merge to the non-protected merge-queue-branch before merging into master. This should always add a version per PR, as PRs are merged after each other... 
All tests will be run once more on the merge event and merging will only happen if they succeed...
We could try longer queues, not sure what would happen in that case.

Waelthus:
should close #1010

iprafols (approved):
Looks fine

PR #1017, opened 2023-06-16
change_auto_versioning

Waelthus:


Waelthus:
changed this to bump version at every commit after PR is generated, but only  push this in the queue, let's see if this actually runs there...

iprafols (approved):

PR #1018, opened 2023-06-21
SDSS fix

armengau:
When reading SDSS spectra (using the DR16Q catalog), using mode="spec" ie. files spec-plate-mjd[..].fits, I found for at least one file an issue reading the file, the HDUs were not there (no idea why) so picca crashed.

This commit is a simple workaround.


armengau:
@alxogm May be interested in this PR

iprafols (approved):

PR #1021, opened 2023-07-05
Update bump_version.yml

Waelthus:
This is one last try to fix the automated version bumping.
This is going back to what I originally put in, but using my personal access token instead of a github bot token. As I'm an admin (and the bot is not), I'm allowed to directly push the version change to master even with branch protection, which the bot is not.
As github has no way to automatically figure out infinite loops, this action cannot be allowed to run when only files affected by the version bumps change. So `.bumpversion.cfg` and `py/picca/_version.py` are on the exclusion list. I also added README and CHANGELOG for good manners so changes of those won't be version changes.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 7.0.2
+current_version = 7.0.3
 commit = True
 
 [bumpversion:file:py/picca/_version.py]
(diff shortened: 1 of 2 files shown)


Waelthus:
added the whole dir containing github actions to the exclusion list and bumped the version in .bumpversion.cfg to be in sync.

PR #1022, opened 2023-07-11
Fix dla model

### LINKED ISSUE #1019: Errors in the Voigt profile functions
There are errors in the methods defining Voigt profiles.

I wanted to understand the Vogt profile functions in order to improve their performance (using scipy). But the code in `py/picca/dla.py` is very difficult to understand.

Thankfully there is a different version of the Voigt profile functions in  `bin/picca_compute_fvoigt.py` (functions voigt and tau). However, they lead to different results. For example, for a random choice of z_dla and N_hi these are the differences (setting ``T=5e4`` for both):
Figure 31 (1)

I was able to understand the process followed in picca_compute_fvoigt, and I am fairly sure it is correct. I followed Rutten 2003 at 3.3.3. I also wrote independently a function using Garnett et al 2018 Section 6:
``` python
from scipy.special import voigt_profile

def garnett_tau(lamb, z, nhi):
    e = 1.6021e-19 #C
    epsilon0 = 8.8541e-12 #C^2.s^2.kg^-1.m^-3
    f = 0.4164
    mp = 1.6726e-27 #kg
    me = 9.109e-31 #kg
    c = 2.9979e8 #m.s^-1
    k = 1.3806e-23 #m^2.kg.s^-2.K-1
    T = 5*1e4 #K
    gamma = 6.265e8 #s^-1
    lamb_alpha = constants.ABSORBER_IGM["LYA"] #A
    
    lamb_rf = lamb/(1+z)
    lamb_rest = lamb/(1+z)
    
    v = c *(lamb_rest/lamb_alpha-1)
    b = np.sqrt(2*k*T/mp)
    small_gamma = gamma*lamb_alpha/(4*np.pi)*1e-10
    
    nhi_m2 = 10**nhi*1e4
    
    tau = nhi_m2*np.pi*e**2*f*lamb_alpha*1e-10
    tau /= 4*np.pi*epsilon0*me*c
    tau *= voigt_profile(v, b/np.sqrt(2), small_gamma)
    
    return tau
```
which I think is easier to understand. This new function perfectly matches the results in picca_compute_fvoigt.

After confirming this, I was able to find the errors in ``py/picca/dla.py``: 
https://github.com/igmhub/picca/blob/72f27b7368d5c2d31b19d1bcfcce65bc630fe2bc/py/picca/dla.py#L85
This should be ``6.265e8``

https://github.com/igmhub/picca/blob/72f27b7368d5c2d31b19d1bcfcce65bc630fe2bc/py/picca/dla.py#L95
This should be ``lambda_rest_frame / lambda_lya``

https://github.com/igmhub/picca/blob/72f27b7368d5c2d31b19d1bcfcce65bc630fe2bc/py/picca/dla.py#L99
The Voigt function defined here gives different values in the very suppressed region. This should have a small impact. I did not understand what the function does anyway. I think we should use external libraries whenever is possible.

https://github.com/igmhub/picca/blob/72f27b7368d5c2d31b19d1bcfcce65bc630fe2bc/py/picca/dla.py#L105
``lambda_rest_frame`` here should be ``lambda_lya``.

I tested the speed of the three functions
image

Note that the version we are using right now is more than a thousand times slower than any of the other two.

----

In conclusion, we need to implement one of this two methods as the standard one in ``picca/dla.py``. But deltas and correlations will be affected, so maybe it is better if we could check the impact of this change first.

cramirezpe:
Fixes #1019 . 

I used Garnett 2018 as a reference as it seemed more clear to me, in the issue the comparison with the old method is detailed.


cramirezpe:
I also added Naim because there are differences in the Lyman beta region between my code and his:
compare_voigt (1)

There might be an error with the constants somewhere..

iprafols:
After discussing it with @cramirezpe  I added a few changes to the code to further optimize. Basically, I merged the two tau functions (they were almost identical) and computed the prefactors that were independent of the DLA parameters outside of the function so that they are only computed once

I also improved the docstring adding the exact equations we use and renamed a few of the variables

From my side this PR is ok, but I think we should wait to merge until  @p-slash  has review it and the discrepancies with the Lyman Beta profiles are understood

p-slash:
> I also added Naim because there are differences in the Lyman beta region between my code and his: compare_voigt (1)
> 
> There might be an error with the constants somewhere..

The constants seem to be the same. Maybe we are seeing the thermal broadening. I think Lyb transition is weaker than Lya, so it will saturate at higher column densities. If this is the case, the choice of b (or T) will affect the profile. The other possibility is that we are seeing difference in approximations.

I will make a comment regarding what to use for oscillator strength and Einstein coefficients while replying to Michael.

py/picca/delta_extraction/masks/dla_mask.py
```diff
@@ -44,61 +46,55 @@ def dla_profile(lambda_, z_abs, nhi):
...
+GAUSSIAN_BROADENING_B = np.sqrt(BOLTZMAN_CONSTANT_K * GAS_TEMP / PROTON_MASS)
+# the 1e-10 appears as the wavelengths are given in Angstroms
+LORENTZIAN_BROADENING_GAMMA_PREFACTOR = 1e-10 / (4 * np.pi)
+TAU_PREFACTOR = (
+    ELEMENTARY_CHARGE**2 * 1e-10 / ELECTRON_MASS / SPEED_LIGHT / 4 / EPSILON_0)
 
-    z_abs: float
-    Redshift of the absorption
+def compute_tau(lambda_, z_abs, log_nhi, lambda_t, oscillator_strength_f, gamma):
+    r"""Compute the optical depth for Lyman-alpha absorption.
```
p-slash, line 73: This function doesn't seem specific to Lya

Waelthus:
I agree with @p-slash that Lyβ needs higher NHI to saturate and thus will be affected more by thermal broadening. What are we using for the b-parameter in both approaches? Are those fixed values for any DLA or something more sophisticated? How well do we know those values? And how large is the impact of the Lyβ differences?
Ideally, we should have a unified approach that is used for both analysis packages and the mocks with some characterization of its uncertainties.

py/picca/delta_extraction/masks/dla_mask.py
iprafols: I fixed the docstring

p-slash:
> 

I use `c=299792.458` km/s from Wikipedia

iprafols:
Hi @p-slash 
Cesar uses 2.9979e8, and `scipy.constants.speed_of_light= 299792458.0`, the same as you use. Probably it's better to use the same value so that we can rule out that the differences come from here. I'll do a commit to change this here

Waelthus:
Given that c=299792458 is an exact physical definition, I agree we should use that value and not some approximated one whereever possible...

cramirezpe:
Well, now there's difference both in Lyman-alpha and Lyman-beta:
compare_voigt (2)

cramirezpe:
All three agree (picca_old, picca_delta_extraction and qsonic) Thanks!
image

iprafols (approved):

iprafols:
then let's merge this and close the issue

PR #1023, opened 2023-07-12
Fix desi tile

iprafols:
I added a try/except bloc so that the DesiTile mode does not crash when there are some files with `_tmp` on the folder


Waelthus (approved):
This should be fine for the moment, esp given that it's only really relevant for analyses on daily. Maybe an explicit check for temp is better, but fine with it if this does the job.

PR #1024, opened 2023-07-12
Fix xcf1d

iprafols:
Fix longstanding issues in the xcf1d computation. The code is seldom used and that is why we did not notice it was broken (also it is not included in the tests)


Waelthus (approved):
Fine with me, given that `compute_xi_ferest_pairs_fast` now stores all the rebinned quantities itself.
We should add a test for xcf1d in the future, though. Maybe only with the new correlation analysis package...

PR #1025, opened 2023-07-12
simplified code to compute var stats fits

iprafols:
The defaults and the start values for the var stats fits were not the same. I modified the code so that they are


iprafols:
to solve the merge conflicts I first need to merge master and then recompute the new test suite

Waelthus (approved):
Totally makes sense. Let's get rid of those hardcoded values...

PR #1028, opened 2023-07-17
Update the github action for tests

Waelthus:
This updates the test github action to use python 3.9 - 3.11 to make sure we actually test on recent versions


Waelthus:
This also needed to update iminuit to more recent versions as the old version we pinned is unavailable for newer python.
Given that we're not using the fitter2 anymore, the previously required pinning might now be unnecessary.
If we want to support specific versions (e.g. 3.8) forever, we should run the tests on those as well

iprafols (approved):

Issue #1032, opened 2023-07-28
Instabilities in the tests of old analysis routines (picca_deltas)


Waelthus:
It looks like something in the automatic testing is unstable again. Two runs performed after just version bumps failed, but re-running the tests makes those pass. I think this only happens since releasing the requirements on iminuit to allow newer versions (and thus allow running with python 3.11 or later which cannot use the old version we pinned before).

However, if I see this right the only failure is in test1_deltas, i.e. running the old picca_deltas routines. Maybe we should just retire old picca deltas, old fitter2 and the corresponding tests completely (they will always stay available in tagged releases until at least v7.2), now that delta_extraction has been successfully used for a year.

iprafols:
let's retire the old picca deltas then

Waelthus:
we retired old picca in version 8.0.0

PR #1033, opened 2023-07-28
Remove old picca routines

Waelthus:
This PR intends to remove old picca routines, such as picca_deltas and picca_fitter2 with their dependencies (but leaving the parts that correlation function and p1d still need). Adresses #1032


Waelthus:
It's expected that code testing coverage will change, I'll use that as a hint regarding which parts may be removable e.g. in picca/{io.py,prep_del.py,...}

iprafols:
io.py will definitely not be removable since it is also used for the computation of the correlation function. Maybe for now we can just remove picca_deltas.py and the tests associated with it. We might as well remove the fitter2 now that we're on this. We have also been using Vega for a while now

Waelthus:
yes, I know that it won't be fully removable, but the other files regarding deltas will be. And for io I'd at least remove a bunch of spectra-read-in (read_data, read_from...) functions, but will keep deltas/objects/dlas/absorbers/drq alive.

Waelthus:
fitter2 is already removed here

iprafols:
cool, makes sense

iprafols (commented):
looks fine, I see that you did not remove the bits of code from io.py and the rest. Maybe you can also remove the fitter2 folder

Waelthus:
Most legacy code that isn't used by for cf et al is gone. Parts of data.Forest are still there as cf hijacks it's variables, prep_del is only kept for stacking, raw_io is kept in case it's still needed (as we don't test it afaik).
We might want to have a test of non-automatically tested parts of the code, e.g. dmat or wick

Which version do we want? 8.0 as this is breaking all bw compatibility to old picca?

iprafols:
I'd say to go to version 8, yes. It does make sense to keep the raw analysis for the meantime (maybe we should try to migrate it to be part of the delta_extraction code at some point. Also, I totally agree on having tests for the missing parts. 

I see that the test suit has changed. Was this needed due to the library changes that made the tests fail?

iprafols (approved):

Waelthus:
Ah, I originally thought I could remove everything related to picca_deltas tests. Just to figure out later that those files were partially inputs of the cf and p1d tests, so one commit needed partial revert...
In the end it should be removals of files and code only (and changing some comments). 

Am 1. August 2023 17:25:08 MESZ schrieb "Ignasi Pérez-Ràfols" ***@***.***>:

Waelthus:
closes #1032

PR #1034, opened 2023-07-28
[bump minor] P1d covariance fix

corentinravoux:
Branch to fix the P1D computation. 
Two important fix;
- The storing of k arrays for the covariance
- The computation of errors for the P1D

This PR is also here to have various discussions about the covariance matrix calculation.


corentinravoux:
Pytest is failing at the delta extraction step, I do not think it comes from this PR

iprafols:
Yes, we know, tests have become unstable again, probably due to some updates (see issue #1032 ) @Waelthus  is preparing a PR to fix this

Waelthus:
The pytest is working again, is this all that you wanna do here? Or is this in progress for later review @corentinravoux?

corentinravoux:
Thanks @Waelthus, For now the two fixes are really important.

corentinravoux:
If we do not need further discussions on covariance, we can merge.

py/picca/pk1d/postproc_pk1d.py
```diff
@@ -737,10 +735,15 @@ def compute_average_pk_redshift(
                 standard_dev, _, _ = binned_statistic(
                     data_snr, p1d_values, statistic="std", bins=snr_bin_edges
                 )
+                standard_dev_error, _, _ = binned_statistic(
```
Waelthus, line 738: this is for rebinning the statistics of interest? I.e. the individual P1d estimates for raw/noise/..., reso correction and so on?
And the line above is for the final P1d estimate (which is part of the P1d_table as well?
In both cases weigthning is done with the forest_snr column?

I guess in principle one could precompute p1d_values, standard_dev  (and everything else that is independent of the column of interest) outside the loop and not compute them once per column, except if you have pixels masked in a column (because of NaN values), but not in the p1d (which I'd consider weird as a NaN should have propagated to the final p1d)...

py/picca/pk1d/postproc_pk1d.py
```diff
@@ -737,10 +735,15 @@ def compute_average_pk_redshift(
...
+                    data_snr, data_values, statistic="std", bins=snr_bin_edges
+                )
                 standard_dev_full = np.copy(standard_dev)
                 mask = np.isnan(standard_dev)
+                mask |= np.isnan(standard_dev_error)
                 if len(mask[mask]) != 0:
```
Waelthus, line 744: here a `(~mask).sum()>0` should give the same answer, also typically masks are defined as 0 for throwing away and !=0 for keeping, so maybe define the masks as `~np.isnan`

py/picca/pk1d/postproc_pk1d.py
Waelthus: also the naming convention is weird, shouldn't this be `standard_dev_col` and the other one `standard_dev_p1d` or similar?

py/picca/pk1d/postproc_pk1d.py
```diff
@@ -749,22 +752,30 @@ def compute_average_pk_redshift(
...
+                    fitfunc_variance_pk1d,
+                    snr_bins,
+                    standard_dev_error**2,
+                    bounds=(0, np.inf),
+                )
 
                 # Model variance from fit function
                 data_snr[data_snr > MEANPK_FITRANGE_SNR[1]] = MEANPK_FITRANGE_SNR[1]
                 data_snr[data_snr < 1.01] = 1.01
                 variance_estimated = fitfunc_variance_pk1d(data_snr, *coef)
```
Waelthus, line 765: is there any idea by now why a weight of `1/(snr-1)**2` is needed? and not of `1/snr**2 ~ ivar`

py/picca/pk1d/postproc_pk1d.py
```diff
@@ -749,22 +752,30 @@ def compute_average_pk_redshift(
...
                 weights = 1.0 / variance_estimated
+                weights_error = 1.0 / variance_estimated_error
                 if apply_z_weights:
                     weights *= redshift_weights
                 mean = np.average(data_values, weights=weights)
                 if apply_z_weights:
                     # Analytic expression for the re-weighted average:
-                    error = np.sqrt(np.sum(weights * redshift_weights)) / np.sum(
-                        weights
+                    error = np.sqrt(np.sum(weights_error * redshift_weights)) / np.sum(
```
Waelthus, line 774: why would you want to use different weights for the mean than you'd use for the standard deviation?

I.e. for the mean you just weigh by the varance estimated for the p1d measurement, but for the std you weigh by the variance for that particular column which would be internally inconsistent.
When you say `error` here, should this be the error on the mean? Or the standard deviation of the dataset around the mean

Waelthus (commented):
ok, not fully sure what you're trying to do here, see above. This change sounds like afterward weights of errors and means will be inconsistent... But maybe I missed something.

Waelthus:
Regarding further discussion of how to compute covariances, maybe open an issue

corentinravoux:
Thanks for your review.

corentinravoux:
Thanks for your review,
The point of the second fix is indeed to have different weight for the error and the average.
With the fix Eric as done, only the P1D are used to do the weighting, and it gives the same error bar for all the columns. I fixed that by separating the weights with P1D (to compute the average) and the weights with col (to compute the error).
I hope this answer most of you interogations, I will try to implement the changes you mentioned

py/picca/pk1d/postproc_pk1d.py
corentinravoux: I agree on changing the naming convention and on precomputing the p1d, I will make the change.

py/picca/pk1d/postproc_pk1d.py
corentinravoux: Here I wanted to know if there are any nan, so the equivalent should be mask.sum() > 0 ? I do not know if this is a more optimized version. I mean between a sum and applying a mask, the timing should be similar.

py/picca/pk1d/postproc_pk1d.py
corentinravoux: Maybe ask @armengau on this point. This is a specificity of the SNR weighting for which only the SNR>1.01 are taken into account. I guess they tested with the snr - 1 method and gave closer P1D to the original measurement compare to just snr.

py/picca/pk1d/postproc_pk1d.py
corentinravoux: See answer below for detail on that

py/picca/pk1d/postproc_pk1d.py
corentinravoux: Ok I understand your comment know. I will do the change

py/picca/pk1d/postproc_pk1d.py
Waelthus: yes, that's what I understood as well, just wondered if they found a theoretical reason why that should be the case in the meantime

Waelthus:
But then I'm not sure if that's a useful thing to do. The part of the other columns that makes it into the final measurements is after all weighted in the same way as the P1d measurements. So the weights should be the same, but the absolute errors should not be. They should be based on the column at hand using the P1d weights...

Waelthus:
What I mean is that we do not necessarily want to compute the mean and (co-)variance of a column based on it's own inverse variance weights, but based on the inverse variance weights of the P1d column and the means, variances of the column of interest, I think it should be something like:

`Var(col)=sum(W_i**2 Var_i(col))/(sum(W_i))**2` (ignoring the covariance between modes) where W_i are the weights based on P1d, and Var_i are the individual variances inside that column. If the column is P1d, `W_i=Var_i**-1` and then `Var=sum(W_i)/(sum(W_i))**2=1/sum(W_i)`, i.e. for that column both approaches are the same as expected. Including covariances Var_i -> Cov_ij, W_i**2 -> W_i W_j and the sums are over both i and j...

If one did this, comparisons between different columns will be simplified, as the impact of each spectrum is the same regardless of column.

I'd be happy to merge the current version for now and discuss (co-)variance estimation in an issue with both of you.

Waelthus (approved):

corentinravoux:
I do not know how to create the tag and if I am allowed ? Let's call it 8.0.2 ?

Waelthus:
so you want to use the version from after this merge as a tag?

corentinravoux:
Yes if that is ok for you, including covariance calculation and all those fixes

PR #1035, opened 2023-08-01
Add version to config

cramirezpe:
For those who install picca through pip, git hash is not possible to get. 

Having the version printed in the config file won't be perfect, but it is better than nothing.


iprafols (approved):

Waelthus:
given that a version is equivalent to a git hash for `pip install picca` this sounds like a good idea. And for people running on their own branches this should give both the latest master version and  the commit hash which makes things clearer as well.

@iprafols: I guess we'll need to include the test_file to the version updating and exempt it from triggering the version bumping action, I'll merge master into here to check

Waelthus:
or rather there shouldn't be an issue, as we don't check if the results actually match (which would be impossible for the timestamp at least)

PR #1037, opened 2023-08-07
[bump minor] Add new FVoigt script

andreicuceu:
A simplified version of the FVoigt computation, based on the code by @moonlovist. Work in progress.


bin/picca_compute_fvoigt_hcd.py
```diff
@@ -0,0 +1,144 @@
...
+
+import numpy as np
+from scipy.constants import speed_of_light
+from scipy.special import wofz
+
+from picca import constants
+
+
+def voigt_profile(x, sigma=1, gamma=1):
+    return np.real(wofz((x + 1j*gamma) / (sigma * np.sqrt(2))))
```
Waelthus, line 12: is there a reason to define the voigt profile function here instead of just using `scipi.special.voigt_profile` in the first place?

Waelthus:
also: please add docstrings and ideally a test...

julienguy (commented):
Hi, I suggest we add more information about the input files and output file in the script help string.
What are the definitions of the columns, what are their units.

bin/picca_compute_fvoigt_hcd.py
Waelthus: I think the difference might be in using the Voigt function V in case of scipy and H in case of this computation, defined like this on wikipedia 
image.
Anyway, using wofz or voigt_profile straight from scipy is fine (with the correct normalizations), just improves readability as you don't need to redefine things here, scipy is using wofz internally anyway (but in a c-based routine).

iprafols:
is this ready for review?

andreicuceu:
Yes, this has been tested and working for a while now.

bin/picca_compute_fvoigt_hcd.py
```diff
@@ -0,0 +1,189 @@
...
+import argparse
+
+import numpy as np
+from scipy.constants import speed_of_light
+from scipy.special import wofz
+
+from picca import constants
+
+
+def voigt_profile(x, sigma=1, gamma=1):
```
iprafols, line 11: docstring missing

bin/picca_compute_fvoigt_hcd.py
iprafols: If both are the same (once properly normalized), I would suggest changing this to using `scipi.special.voigt_profile` given that this is what we use in the DLA masking module, but this would not be a show stopper for the merge

bin/picca_compute_fvoigt_hcd.py
```diff
@@ -0,0 +1,189 @@
...
+
+    Returns
+    -------
+    array
+        Flux corresponding to a voigt profile
+    """
+    # wave = lambda in A and N_HI in log10 and 10**N_hi in cm^-2
+    wave_rf = wave / (1 + z)
+
+    e = 1.6021e-19  # C
```
iprafols, line 36: These should be defined as constants outside of the function. Consider importing them from `py/picca/delta_extraction/masks/dla_mask.py` or at least change the variable names to have a similar naming convention

bin/picca_compute_fvoigt_hcd.py
```diff
@@ -0,0 +1,189 @@
...
+    epsilon0 = 8.8541e-12  # C^2.s^2.kg^-1.m^-3
+    f = 0.4164
+    mp = 1.6726e-27  # kg
+    me = 9.109e-31  # kg
+    c = speed_of_light  # m.s^-1
+    k = 1.3806e-23  # m^2.kg.s^-2.K-1
+    T = 1e4  # K
+    gamma = 6.265e8  # s^-1
+
+    wave_lya = constants.ABSORBER_IGM["LYA"]  # A
```
iprafols, line 46: Do we want to allow people to run the LYB profile as well?

bin/picca_compute_fvoigt_hcd.py
```diff
@@ -0,0 +1,189 @@
...
+        Redshift
+    N_hi : float
+        Log10 column density
+
+    Returns
+    -------
+    array
+        Flux corresponding to a voigt profile
+    """
+    # wave = lambda in A and N_HI in log10 and 10**N_hi in cm^-2
```
iprafols, line 33: This could be moved to the docstring

bin/picca_compute_fvoigt_hcd.py
```diff
@@ -0,0 +1,189 @@
...
+        Cosmology object
+    differential : bool, optional
+        Whether we have to account for a dlambda / dr factor, by default False
+
+    Returns
+    -------
+    (array, array)
+        (comoving distance grid, profile as a function of comoving distance)
+    """
+    z = wave / constants.ABSORBER_IGM["LYA"] - 1
```
iprafols, line 80: same here

iprafols (changes requested):
On top of the comments I added, I would run `pylint` and `yapft` to follow the same coding standards as the rest of the new parts of picca

bin/picca_compute_fvoigt_hcd.py
Waelthus: they are as the scipy.special.voigt_profile is just a cythonized wrapper around woffz anyway, I'd mildly prefer using existing functions (with some comment about the difference in intrinsic normalization) over having our own version for smaller things like this, even more given we're using the scipy function elsewhere anyway...

iprafols:
Hi all, I modified the file so that it now uses functions already defined in dla_mask, instead of defining new functions that do the same things. @andreicuceu can you check that I did not accidentally add a normalisation issue?

bin/picca_compute_fvoigt_hcd.py
```diff
@@ -6,56 +6,8 @@
...
-    u = (wave_rf - wave_lya) / Deltat_wave
-    H = voigt_profile(u, np.sqrt(1/2), a)
-
-    absorbtion = np.sqrt(np.pi) * e**2 * f * wave_lya**2 * 1e-10 * H
-    absorbtion /= (4 * np.pi * epsilon0 * me * c**2 * Deltat_wave)
-
-    # 10^N_hi in cm^-2 and absorb in m^2
-    tau = (10**N_hi * 1e4 * absorbtion).astype(float)
-    return np.exp(-tau)
+from picca.delta_extraction.masks.dal_mask import (
```
Waelthus, line 9: typo dal_mask -> dla mask

bin/picca_compute_fvoigt_hcd.py
```diff
@@ -161,7 +113,8 @@ def main():
     wave = np.arange(2000, 8000, 1)  # TODO this grid may be too sparse
     integrand = np.empty((dN_NHI.size, wave.size // 2 + 1))
     for i, NHI in enumerate(dN_NHI):
-        profile_wave = get_voigt_profile_wave(wave, args.z_dla, NHI)
+        profile_wave = np.exp(
+            -compute_tau(wave, args.z_dla, NHI, LAMBDA_LYA, OSCILLATOR_STRENGTH_LYA, GAMMA_LYA)
```
Waelthus, line 117: should the Lyβ absorption be added as well as we do this in the DLA mask? If not this does exactly the same thing as in the masking function, so it should be fine. Or rather, if it weren't fine we should not change things here, but over there...

iprafols: I'd be happy to add it, yes. What do you think @andreicuceu ?

iprafols: we would just need to replace line 116 by `profile_wave = dla_profile(wave, args.z_dla, NHI)` (and fix the imports)

bin/picca_compute_fvoigt_hcd.py
```diff
@@ -0,0 +1,142 @@
...
+    comov_dist_weights, weights_comov_dist = profile_wave_to_comov_dist(
+        weights_wave[:, 0], weights_wave[:, 1], fidcosmo)
+
+    weight_interp = np.interp(
+        comov_dist_prob, comov_dist_weights, weights_comov_dist, left=0, right=0)
+    mean_density = np.average(dla_prob_comov_dist, weights=weight_interp)
+
+    wave = np.arange(2000, 8000, 1)  # TODO this grid may be too sparse
+    integrand = np.empty((dN_NHI.size, wave.size // 2 + 1))
+    for i, NHI in enumerate(dN_NHI):
```
Waelthus, line 115: is it correct, that in this case we want to iterate over DLAs with different column densities at the same redshift? I.e. essentially integrating the absorption of many absorbers at the same point according to some cddf? In which case this is the correct thing to do.
But it could be more clearly documented what exactly this script is supposed to be doing...

bin/picca_compute_fvoigt_hcd.py
Waelthus: This reminds me that we should have at least a very basic test trying to at least run this file, so we're sure that we don't completely break it later on.

iprafols:
ok so tests are still missing

iprafols:
tests are failing due to a weird coveralls error. Maybe their server is down. Otherwise, they might have updated something and the new version is not backwards compatible. We should try to rerun failing tests in a while to see if it was just their server being down or not

bin/picca_compute_fvoigt_hcd.py
Waelthus: ok, cleared this with @moonlovist and the z is just an eff. redshift where all DLAs are placed

Waelthus:
somewhere things are going wrong, here: 
grafik
expected:
grafik

andreufont:
What is the status of this PR, @iprafols , @Waelthus , @andreicuceu ?

corentinravoux:
We briefly discussed this PR in the latest Lya software telecon. If I remember correctly Ting mentioned that he wanted to work on that, I am not seeing him in the picca package. @andreicuceu any news on this PR ?

Waelthus:
I personally last looked at this with @moonlovist (i.e. Ting) in Hawaii. Not sure what the current status is. But given that the PR just adds a new script, there shouldn't be any issues outside of this new routine when merging later.
I'd suggest merging master back into here so that @moonlovist can finish the routine with an up-to-date branch and won't encounter issues inside the new routine when merging. I'd also suggest adding a test for the routine while coding it up.

corentinravoux:
@moonlovist What is the status of this PR ?

corentinravoux:
After discussion on Lya software telecon, we suggested to push this PR as it is quite independent from the rest of the code.

corentinravoux:
Merging once the tests are completed

corentinravoux:
I ran a formatter on the code to answer your request @iprafols 
Merging the PR.

corentinravoux:
Please delete your branch if you do not need it

PR #1038, opened 2023-08-16
P1d nan fix

corentinravoux:
Fix for the previous PR. In certain cases (no chunks in a specific wavenumber and redshift bin), the new changes crashed.
Improvement of the masking of np.nan by the way.

### DIFF
--- a/py/picca/pk1d/postproc_pk1d.py
+++ b/py/picca/pk1d/postproc_pk1d.py
@@ -704,6 +704,18 @@ def compute_average_pk_redshift(
         n_array[ikbin] = num_chunks
 
         if weight_method == "fit_snr":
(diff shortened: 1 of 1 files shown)


corentinravoux:
Is it possible to push another version for this fix ?

Waelthus (approved):
I'm fine with the change.
If the source is actually just empty k-bin/z-bin combinations for certain spectra, e.g. because of more masking for the spectrum, i.e. shorter spectrum, things should be ok. Else we should figure out what sets the nan

corentinravoux:
Yes I think the np.nan orginate from empty SNR bins, and the k, z empty bins were already dealt with a warning not accounted for by the previous PR.

PR #1043, opened 2023-10-12
if no variabes are floated, do not perform fit

iprafols:
When fixing all of eta, var_lss, and fudge, do not perform a fit with all the variables fixed


Waelthus (approved):
this looks fine and does match the behaviour I'd expect...

Issue #1044, opened 2023-10-18
Wrong redshift of absorber in fast_metal_dmat


cramirezpe:
There's an issue in the new picca_fast_metal_dmat that is causing the redshift of the absorbers to be at a smaller redshift than it should.

Calum just shared this plot:
![image](https://github.com/igmhub/picca/assets/58682644/478b1633-5d1b-4c12-bc43-2b1237db7580)
The difference in number of points is probably due to the use of higher resolution in the fast one.

We still need to check the xdmat case.

andreufont:
@calumgordon - could you check what happens with the ~500 bins in the blue histogram that have z_eff_civ=0?

andreufont:
About the total number of point in that histogram, shouldn't these matrices have 2500^2 elements, i.e., (50x50)^2? This would be 6250000 (>6M) but I don't see that many points in the histogram...

calumgordon:
It's the CIVxCIV correlation in the metal dmat, which has 150x100 bins in the fast case, and 75x50 in the slow. 

For the bins with z_eff = 0, there isn't a lot of info easily available, so I'm not sure.

calumgordon:
Ok, here is a plot of RT_CIV, where we can see the missing bins from 176 to 200 Mpc/h. I've also checked and this indeed corresponds to (200-176)/4  *  75 = 450 bins. 


![RT_CIV](https://github.com/igmhub/picca/assets/57399293/4e35c2f7-70a8-4dd8-832e-8bfcdc1ffffa)

andreufont:
This is soooo confusing. I first thought that it would be a few rt bins that happens to not be covered because of the 0.1% of data used in the computation of matrices, but the fact that they are all localised at rt > 175 Mpc/h means that this is more likely a bug somewhere... It would be quite a coincidence.

My next guess is that somewhere in the code, one uses the CIV redshift (and not the Lya) to compute maximum angular separation to include in the search for neighboring quasars, and that this cause quasars at > 175 Mpc/h (in Lya coordinates) to not be included because they correspond to > 200 Mpc/h in SiIV coordinates

calumgordon:
RT_LYA - RT_CIV vs RT_LYA. There is a difference between the co-ordinates as expected but it's quite small.
![rt_diff_lya_civ](https://github.com/igmhub/picca/assets/57399293/f99d166d-c966-4776-88f0-3cde7b1bc126)

andreufont:
[...]

PR #1045, opened 2023-10-18
Redshifts in metal matrices [bump minor]

julienguy:
Record in the metal matrices the weighted mean redshift of the input to the remapping matrix (i.e. the true absorber redshift) and not the output of the remapping (the 'incorrect' Lya redshift). Both redshifts (input and output) could be useful, but for the purpose of interpreting correctly the metal biases coming from a fit using the metal matrices, the metal redshifts are obviously the correct ones to use. This is what was done in the previous code, so we are fixing here an inconsistency.

This PR addresses issue #1044 .


andreufont:
@iprafols - this looks like a trivial bug fix, and it would be great to have it included in a new tagged version. Thanks!

iprafols (approved):

PR #1046, opened 2023-10-24
Remove blinding civ [bump minor]

iprafols:
Blinding is removed when running on metal forests


bin/picca_fast_metal_dmat.py
```diff
@@ -419,7 +419,7 @@ def main():
             )
             userprint("Please specify the directory with option --delta-dir")
             sys.exit(1)
-    blinding = io.read_blinding(args.delta_dir)
+    blinding = io.read_blinding(args.in_dir, args.lambda_abs, args.lambda_abs2)
```
andreufont, line 422: @iprafols - you have changed here delta_dir for in_dir. Was this intended, or a copy-paste bug?

bin/picca_fast_metal_xdmat.py
```diff
@@ -402,7 +402,7 @@ def main():
             )
             userprint("Please specify the directory with option --delta-dir")
             sys.exit(1)
-    blinding = io.read_blinding(args.delta_dir)
+    blinding = io.read_blinding(args.in_dir, args.lambda_abs, args.obj_name)
```
andreufont, line 405: Same here

py/picca/io.py
```diff
@@ -232,16 +238,49 @@ def read_blinding(in_dir):
...
+    # This is for ImageHDU format
+    if "LAMBDA" in hdul:
         header = hdul["METADATA"].read_header()
+
+        # read blinding
         blinding = header["BLINDING"]
-    else: # This is for BinTable format
+
+        # check if we are doing metal forests
+        if "MIN_RF_WAVE" in header and "MIN_RF_WAVE" in header:
```
andreufont, line 249: @iprafols - you probably meant MAX_RF_WAVE in one of these clauses

py/picca/io.py
```diff
@@ -232,16 +238,49 @@ def read_blinding(in_dir):
...
             blinding = "none"
 
+        # check if we are doing metal forests
+        if "MIN_RF_WAVE" in header and "MIN_RF_WAVE" in header:
+            lya_wave = ABSORBER_IGM["LYA"]
+            min_rf_wave = header["MIN_RF_WAVE"]
+            max_rf_wave = header["MAX_RF_WAVE"]
+            if min_rf_wave > lya_wave and max_rf_wave > lya_wave:
+                if lambda_abs2 is None:
+                    lambda_abs2 = lambda_abs
```
andreufont, line 277: Why do you need to set lambda_abs2 here?

py/picca/io.py
```diff
@@ -232,16 +238,49 @@ def read_blinding(in_dir):
...
+                blinding_tracers = ["LYA", "QSO"]
+                if tracer1 not in blinding_tracers and tracer2 not in blinding_tracers:
```
andreufont, line 279: @iprafols - I think you want to have an "or" here, since we don't want to blind QSO x CIV either. Basically, you want to blind only when both tracers are either LyA or QSO

andreufont (changes requested):
@iprafols - I added several comments on the code, most of them are important enough that should be fixed. 

I'm also surprised by the large number of files changed, in particular some log files related to picca delta attributes? Were these added by mistake?

andreufont:
Another comment: when @julienguy runs his analysis of CIV auto-correlations, he actually tells Picca to run the Lya auto-correlation but passes CIV deltas. He does this because he wants to see use the "wrong" Lya coordinates that we use in the main analysis, but without having to deal with the Lya correlation "contaminating" the CIV signal.

In order to blind these, then, I do think that we should also modify picca_deltas so that the CIV deltas are not blinded to start with.

bin/picca_fast_metal_dmat.py
iprafols: ups, copy-paste bug, I'll fix it right away

iprafols:
Hi @andreufont , I fixed the comments on the code. Thanks!

iprafols:
@andreufont , are you happy with the results? should we merge this?

bin/picca_xcf_angl.py
```diff
@@ -253,7 +253,7 @@ def main(cmdargs):
     xcf.lambda_abs = constants.ABSORBER_IGM[args.lambda_abs]
 
     # read blinding keyword
-    blinding = io.read_blinding(args.in_dir)
+    blinding = io.read_blinding(args.in_dir, args.lambda_abs, args.lambda_abs_obj)
```
andreufont, line 256: I would have expected the second argument here would be 'obj_name' as in picca_xcf, but I noticed that this variable is named different in this (not very popular) script, so we can leave it as is for now.

py/picca/io.py
```diff
@@ -232,16 +238,49 @@ def read_blinding(in_dir):
...
-    else: # This is for BinTable format
+
+        # check if we are doing metal forests
+        if "MIN_RF_WAVE" in header and "MAX_RF_WAVE" in header:
+            lya_wave = ABSORBER_IGM["LYA"]
+            min_rf_wave = header["MIN_RF_WAVE"]
+            max_rf_wave = header["MAX_RF_WAVE"]
+            if min_rf_wave > lya_wave and max_rf_wave > lya_wave:
+                if tracer2 is None:
+                    tracer2 = lambda_abs
```
andreufont, line 255: @iprafols - where have you defined lambda_abs? Shouldn't you use here tracer2 = tracer1?

andreufont: Have you tested this code? It looks like it should have crashed...

py/picca/io.py
```diff
@@ -232,16 +238,49 @@ def read_blinding(in_dir):
...
             blinding = "none"
 
+        # check if we are doing metal forests
+        if "MIN_RF_WAVE" in header and "MAX_RF_WAVE" in header:
+            lya_wave = ABSORBER_IGM["LYA"]
+            min_rf_wave = header["MIN_RF_WAVE"]
+            max_rf_wave = header["MAX_RF_WAVE"]
+            if min_rf_wave > lya_wave and max_rf_wave > lya_wave:
+                if tracer2 is None:
+                    tracer2 = lambda_abs
```
andreufont, line 277: Again, this should be `tracer2 = tracer1`

py/picca/io.py
andreufont: @iprafols - it looks like you have fixed this above (for the ImageHDU format) but you have not fixed it for the BinTable format. It still says "and" instead of "or"

andreufont (changes requested):
@iprafols - has this been tested? I found a couple of bugs that I think should be fixed, and it would also be good if someone looking into these (Andrea? Julien? Cesar?) checked that the branch works as intended.

iprafols:
I fixed the bugs, but indeed we should fully test this (I have not had the time to properly do it)

andreufont (commented):
The changes look good. Let's see if @Andrea-MG and @julienguy can run their CIV codes on this branch and get unblind results, before we merge.

iprafols:
tests (python 3.10) are failing due to a weird coveralls error. Maybe their server is down. Otherwise, they might have updated something and the new version is not backwards compatible. We should try to rerun failing tests in a while to see if it was just their server being down or not

andreufont (changes requested):
I don't think we are doing this right. When one of the tracers is not LYA or QSO, the rest-frame range shouldn't matter. In other words, if you are looking at CIV absorption in the LYB region, you should not blind either.

andreufont:
@iprafols , @Andrea-MG - I think we should change the order of the if clauses. 

```
if tracer1 not in blinding_tracers or tracer2 not in blinding_tracers:
    # one of the tracers is not relevant, so do not blind (ever)
    blinding=None
else:
    # we are now in the danger zone, the only way out is if there is no LyA absorption in this region
    if min_rf_wave > lya_wave:
        blinding=None
```

Andrea-MG:
> The changes look good. Let's see if @Andrea-MG and @julienguy can run their CIV codes on this branch and get unblind results, before we merge.

I sent feedback to @iprafols and after the fixes I could run successfully cf, xcf, dmat and xdmat with these changes, testing with CIV and Lyb forests.

Andrea-MG (approved):
I have run successfully cf, xcf, dmat and xdmat with these changes, testing with CIV and Lyb forests with @iprafols and get in all "blinding = none" in all.

py/picca/io.py
```diff
@@ -232,16 +238,60 @@ def read_blinding(in_dir):
...
+            # check if we are doing metal forests
+            if "MIN_RF_WAVE" in header_lambda and "MAX_RF_WAVE" in header_lambda:
+                lya_wave = ABSORBER_IGM["LYA"]
+                lyb_wave = ABSORBER_IGM["LYB"]
+                min_rf_wave = header_lambda["MIN_RF_WAVE"]
+                max_rf_wave = header_lambda["MAX_RF_WAVE"]
+                if ((min_rf_wave > lya_wave and max_rf_wave > lya_wave) or
+                    (min_rf_wave < lyb_wave and max_rf_wave < lyb_wave)):
+                    # we are doing a metal forest, so do not blind (ever)
+                    blinding = "none"
```
andreufont, line 266: @iprafols , @Andrea-MG - Are you sure about this? It looks to me that the this code would not blind the QSO x LyA(LyB) correlation. 

I believe the "if" statement in line 263 should only have the first line (check that both rf_wave are above lya_wave), but it should not have the second condition.

Issue #1047, opened 2023-10-31
Fix and update BAL wavelengths for LyB region


paulmartini:
The list of BAL lines in [bal_mask.py](https://github.com/igmhub/picca/blob/7bd2cc5be838b772e315c5aa1d82b52b7f09e9c0/py/picca/delta_extraction/masks/bal_mask.py#L26) has the wrong wavelength for LyB (1020 vs. the correct value of 1025.7). For a "LyB region" that extends to rest frame 920A, we should also mask some of the other Lyman lines. I'd at least add 972.5 and 949.7. A few other strong BAL features NIII 989.9 and CIII 977.0.

iprafols:
Hi Paul, not sure why I can't appoint you as a reviewer, but I created a PR (#1048 ) fixing this. Can you please review it and confirm that the changes are indeed fixing this issue?

paulmartini:
Hi @iprafols -- Can we close this issue based on [PR1048](https://github.com/igmhub/picca/pull/1048)? Just noticed it is still open.

iprafols:
Sure

PR #1048, opened 2023-10-31
fix and update BAL wavelengths for Lyb region

iprafols:
I fixed the BAL wavelengths following issue #1047 . 

I also sorted the wavelengths


paulmartini (approved):
I've reviewed the wavelengths (and change overall) and this looks good.

Waelthus (approved):
As Paul is the expert here, his agreement should be enough

PR #1049, opened 2023-11-13
Fix raw mocks bug

andreicuceu:
We cannot run raw mock correlations with the rebin option due to missing delta_lambda field in the header of raw mock deltas.


iprafols (approved):

iprafols:
we just need to wait for the fix to the coverall workflow to make sure all tests pass here

PR #1050, opened 2023-11-13
Add flag to turn off non-blind cosmology messages

andreicuceu:
Picca will briefly become a dependency of Vega (as I will discuss tomorrow), and we will be calling the Cosmo class from outside. Therefore we need a flag to turn off the print statements done when there is no blinding, because they are not relevant when this is called from Vega.


iprafols (approved):
I think we could even remove these messages. Honestly, with the blinding and the fact that they are printed in logs we seldom read,  they have become close to useless. Maybe we can just leave the blinding / not blinding message and remove the others (but let's discuss this in the telecon this afternoon)

PR #1054, opened 2023-12-08
[bump major] Unblinding y1

andreicuceu:
Removes the automatic blinding of Y1 correlations. Only picca_export is changed, with the rest of the pipeline unaffected.

Note that the blinding field in the header of exported correlations ('desi_y1') will remain, because we still need Vega to know about it for full-shape blinding.

### DIFF
--- a/bin/picca_export.py
+++ b/bin/picca_export.py
@@ -11,6 +11,7 @@ import os.path
 
 from picca.utils import smooth_cov, compute_cov
 from picca.utils import userprint
+from picca.delta_extraction.utils import UNBLINDABLE_STRATEGIES
 
 
 def main(cmdargs):
@@ -264,7 +265,7 @@ def main(cmdargs):
     ]
 
     # Check if we need to unblind
-    if blinding in ['desi_y1', 'desi_m2']:
+    if blinding in UNBLINDABLE_STRATEGIES:
         userprint("Y1 correlations are not blinded.")
         blinding = 'none'
         data_name = 'DA'
--- a/bin/picca_export.py
+++ b/bin/picca_export.py
@@ -81,12 +81,6 @@ def main(cmdargs):
         choices=['lyaxlya', 'lyaxlyb', 'qsoxlya', 'qsoxlyb'],
         help='Type of correlation. Required to apply blinding in DESI')
 
-    parser.add_argument(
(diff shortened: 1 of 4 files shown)


iprafols:
I modified the code so that we can run the entire pipeline with the `unbllind` flag set. I have also changed its default value to True. Otherwise, it looks fine

iprafols:
@andreicuceu , let me know if you're happy with the changes and then I'll merge the PR

Waelthus:
I added a bump major tag to make it very clear that important things changed.

iprafols:
Hi Andrei, so do you want to *always* have the blinding flag all the way to `picca_export`? Or having it as the default choice would also make sense? 

If you always want to have the blinding flag on, then I suggest we move the variable `UNBLINDABLE_STRATEGIES` to `pica_export` and then we don't check for this at the delta extraction level. 

If you can accept the possibility of having the blinding flag off for unblinded data, then I can just change back the default value.

What do you think? Since this is a bigger  change, maybe we can start by changing the default value of `unblinding` to `False` and then discuss this on a picca task force telecon (maybe next week?) What do you think?

alxogm:
Hello, is there any blocking factor to merge this PR? I would like to have a tagged version to re-run the unblinded baseline analysis.

andreicuceu:
Hi @iprafols, just got back to this.

Yes, I would like the blinding flag to always be there.

In truth, that is not really a blinding flag anymore. It is a field that tells us which part of the survey the data comes from (EDR, Y1, Y3, etc.). We then use this information to apply or not blinding. So it might make sense to even rename that field to something like `latest_survey` (to indicate which is the latest survey we have data from in the dataset being analyzed). However, this is a change that would break backwards compatibility and be much larger than I want this pull request to be. So I suggest we leave it as an issue for the future.

For now, I think it would make sense to just move the `UNBLINDABLE_STRATEGIES` to `picca_export`, as you suggest.

iprafols:
Hi @andreicuceu, I moved the variable `UNBLINDABLE_STRATEGIES` to `picca_export` and removed the check from `delta_extraction`. Can you check that you are ok with the changes? From my side we are ready to merge

andreicuceu:
Hi @iprafols, look good. I'm happy to merge.

PR #1055, opened 2024-01-22
Moving constants to using scipy.constants

iprafols:
Some constants were defined in DLA mask files, but are now changed to using scipy.constants


Waelthus:
need to update test files with different values as DLA profiles are subtly changed in here (on a level of 1e-4 or so)

Waelthus (approved):

PR #1058, opened 2024-02-06
read version directly from version file

iprafols:
This is a small change so that picca version is loaded from the _version.py file when loading the config. We discovered that depending on how picca was installed using pip the picca version saved in the .config.ini file might be bogus. This should not affect the git hash (also saved there).


Waelthus (approved):
I'm fine with the change.
Do we know under which circumstances the incorrect version was reported? Is this something we might wanna store as well to detect e.g. inconsistent environments (assuming that causes the issue)?

PR #1060, opened 2024-02-19
now keeping tile, petal, and loc

iprafols:
This should propagate the read values of tile, petal and night values to the delta files, if they are available in the catalogue. They are stored as a string. Lines of sight with multiple observations contain all the values separated by "-".


iprafols:
@julienguy , this should fix your complaint the other day about not having the tile information in the delta files. Can you review that this is indeed what you needed?

py/picca/delta_extraction/data_catalogues/desi_data.py
```diff
@@ -338,6 +338,12 @@ def format_data(self,
                     "z": row['Z'],
                 }
                 args["log_lambda"] = np.log10(spec['WAVELENGTH'])
+                if "TILEID" in row:
+                    args["tile"] = row['TILE']
```
Waelthus, line 342: Isn't it called TILEID?

py/picca/delta_extraction/data_catalogues/desi_data.py
```diff
@@ -338,6 +338,12 @@ def format_data(self,
...
+                if "PETAL_LOC" in row:
+                    args["petal"] = row['PETAL_LOC']
+                if "NIGHT" in row:
+                    args["night"] = row['NIGHT']
```
Waelthus, line 347: Should we add e.g. FIBER or EXPID? Are these fields all read earlier from HDU2 (EXP_FIBERMAP) of the coadd-files, or just when reading spectra-files?

Waelthus (approved):
I'm fine with this change assuming it's actually keeping the info with it. Could add more fields potentially

py/picca/delta_extraction/data_catalogues/desi_data.py
Waelthus: Just checked and here we'd just copy over info from the catalogue, is all this info in the actual quasar cats? or would we need to assemble the data from the coadd-files?

Waelthus (changes requested):
I think at least the typo needs fixing

py/picca/delta_extraction/data_catalogues/desi_data.py
iprafols: Yes, I have fixed the bug (fixed it in a commit). Should we rename the `tile` member in `DesiForest` to `tileid` as well?

Waelthus (approved):

py/picca/delta_extraction/data_catalogues/desi_data.py
Waelthus: maybe that would be good for consistency... Esp given that e.g. eBOSS did have a PLATE instead of TILEID anyway, so it's a DESI specific thing.

julienguy (changes requested):
I just ran a test with this branch and this is just not working. Nothing is saved for NIGHT PETAL TILEID in the METADATA of the delta files.

Example run, ini file: 
```/global/cfs/cdirs/desicollab/users/jguy/lya/instrumental-systematics/deltas/delta-civ-tile-2-sub-1/picca_delta.ini```
And delta files in :
```
/global/cfs/cdirs/desicollab/users/jguy/lya/instrumental-systematics/deltas/delta-civ-tile-2-sub-1/Delta
```

Waelthus:
In case we still want to add this feature:
- I think the issue is that in this PR we tried to read the relevant information from the catalogs, but those only contain the summarized info like LASTNIGHT etc iirc
- I think the fix could be as easy as reading the info from the coadd-files (which iirc store both summarized and per-exp data). Those are open anyway at the point those attributes are set in this version of the code, we probably just need to read the additional headers...
- Not sure if this PR would still be wanted/needed though or if some part of the functionality has been added elsewhere

corentinravoux:
@iprafols  or @calumgordon  if I remember from the latest Lya soft telecon, you wanted to rework this PR, any news ?

Waelthus:
I guess given this PR is based on an out-of-date master with merge conflicts, it might be easier to start from scratch.
I tried to generate a fix like I suggested above, it's part of #1087 and so far works for the desi fast healpix mode only.
I guess a fix for other modes could look similar...

calumgordon:
We more or less have this functionality from your cross-exposure branch no?

corentinravoux:
In the end the cross-exposure is now in the main branch. And I do not think that the cross-exposure have the same purpose than the PR here, because I am making the continuum fit on all exposures, but I am not saving the information of each exposures (tile, petal, loc).

corentinravoux:
I guess we can wait to discuss that in a software telecon. We should also clean the issues

PR #1064, opened 2024-06-11
[bump minor] p1d masking change

Waelthus:
change the p1d combination to reject short spectra or spectra with heavy masking based on a certain fraction of spectrum, not number of pixels.


Waelthus:
The P1d tests run, but obviously results change with this change, could update the tests if we decide to merge this or let them use the previous defaults

bin/picca_Pk1D.py
```diff
@@ -103,11 +103,19 @@ def process_all_files(index_file_args):
         first_pixel_index = (np.argmax(selected_pixels) if
                              np.any(selected_pixels) else len(selected_pixels))
 
+
+        if linear_binning:
+            max_num_pixels_forest_theoretical=(1180-1050)*(delta.z_qso+1)/pixel_step  #currently no min/max restframe values defined, so hard coding for the moment
```
corentinravoux, line 108: We cannot get the extremum values from the deltas ? It would avoid the hard coding part

bin/picca_Pk1D.py
```diff
@@ -103,11 +103,19 @@ def process_all_files(index_file_args):
...
+        else:
+            max_num_pixels_forest_theoretical=(np.log(1180)-np.log10(1050))/pixel_step
         # minimum number of pixel in forest
-        min_num_pixels = args.nb_pixel_min
+        if args.nb_pixel_min is not None:
+            min_num_pixels = args.nb_pixel_min
+        else:
+            min_num_pixels = int(args.nb_pixel_frac_min*max_num_pixels_forest_theoretical)     #this is currently just hardcoding values so that spectra have a minimum length changing with z; but might be problematic for SBs
```
corentinravoux, line 115: Is it because the rest frame wavelength are different for SB ? In that case we should really get the rest frame limits from the deltas

bin/picca_Pk1D.py
```diff
@@ -168,7 +176,15 @@ def process_all_files(index_file_args):
...
-            if num_masked_pixels > args.nb_pixel_masked_max:
+            
+            if (args.nb_pixel_masked_max is not None):
+                max_num_masked_pixels = args.nb_pixel_masked_max
+            elif linear_binning:
+                max_num_masked_pixels = int(args.nb_pixel_masked_frac_max*(np.max(lambda_new)-np.min(lambda_new))/pixel_step)           #this only accounts for masking inside the spectrum, not at the adges
+            else:
+                max_num_masked_pixels = int(args.nb_pixel_masked_frac_max*(np.max(log_lambda_new)-np.min(log_lambda_new))/pixel_step)
+                            
+            if num_masked_pixels > max_num_masked_pixels:
```
corentinravoux, line 187: So now the default is the fraction right ? but you can still force a mode with a number of pixels as previously.

bin/picca_Pk1D.py
```diff
@@ -464,6 +494,23 @@ def main(cmdargs):
...
+            # could make this fraction a new default, but that would probably cause trouble for SBs
+            # args.nb_pixel_frac_min = 0.13
+            args.nb_pixel_min = 75    #this is the previous default
+    elif not (args.nb_pixel_frac_min is None or args.nb_pixel_min is None):
+        print("both nb_pixel_frac_min and nb_pixel_min were set, using the latter")
+        args.nb_pixel_frac_min=None
+
+    if args.nb_pixel_masked_frac_max is None and args.nb_pixel_masked_max is None:
+            # could make this, i.e. 10% of the estimated forest length the new default
+            args.nb_pixel_masked_frac_max = 0.21
```
corentinravoux, line 507: Is this the default value ? why it is set differently than the values in the parser ?

corentinravoux:
I added some comments, I do not remember the status of the tests on the number of chuncks that we can get back.

bin/picca_Pk1D.py
Waelthus: I think the deltas only store the obs_frame cuts and redshift_cuts, not the restframe range used for the forest. But I'd need to double check. Also not sure if this would create some issues regarding the sidebands...

bin/picca_Pk1D.py
Waelthus: I'm setting it differently to allow both options, either setting a number of pixels or setting a fraction...

Waelthus:
I think it were not many chunks, mostly because the max masked values used in current analyses are way higher than the previous defaults already... So probably not urgent to merge either. But it would allow a way to actually have the cuts defined more evenly across redshifts, the old way would allow masking significant parts of the spectrum at z=2, but not nearly as much at z=4

corentinravoux:
What I suggest, and regarding the strategy in Y1: You can make how we were masking as the standard, and the masking fraction as a potential option. We can then consider the masking fraction as a variation of the Y1 baseline ? If we consider that this is a good enough improvement, we will add it to the next baseline

corentinravoux:
@Waelthus What do you tink we do with this PR ? I guess keeping both option is still a good idea.

bin/picca_Pk1D.py
Waelthus: the idea is: you can give a fraction or number via the appropriate keywords, if you give both the code prints a message and uses the fraction, if you give none, the old default pixel number is used, elsewise it uses whatever you tell it to use...

Waelthus:
I changed the defaults so that the old defaults should be the new ones as well, but we can alternatively use the new defaults. If you have a test-dataset where running and comparing master and this PR yields the same result while some form of rejection/masking is expected to happen (not sure if that's the case for the auto tests), I guess we could merge this.

corentinravoux (approved):
I ran on a larger sample (~10% of iron) and there was no rejection differences. It is good that we have both masking option now

Waelthus:
sounds good to go for me then. Let's merge this after seeing if the other PR succeeded in fixing the deployment.

PR #1067, opened 2024-07-04
[bump minor] Pk1d covariance new

corentinravoux:
Two improvements:
- Implementation of a new estimate of the covariance matrix, detailed here: https://www.overleaf.com/7592125432tksctmwfpngw#c4fed4
- Covariance matrix calculation computed in a vectorized way, and with memory optimization. Now working in ~ 10 NERSC node minutes for a Y1 sample.


corentinravoux:
image
Comparison of covariance matrix calculation on a small example (left before, right after), there are not identical since the definition has changed, but their shape are similar.

Waelthus:
Ok, this sounds like great improvement, I guess 10 NERSC mins should be much better than the previous version and at least similar structures are obtained. How small is small? E.g. is a smoothing step necessary to make any sense out of the right/bottom 1/3 of the covar?
Also do we have an idea where the subtle differences come from? Is this just numerical from e.g. different order of operations or is the difference in method actually analytically expected to produce slightly different results? E.g. at the strong correlation spikes around pixel [0,8] (is that correlation due to the SiII/III correlations? Or something to do with DLAs/ cont fitting? Looks like it's the longest range correlations we have in here, but somewhat weaker in the new approach...)
I guess we cannot do inter-redshift-bin correlations just yet, but that should be ok...

Could you run this on a significantly large set of data? I'll go through the code changes now...

py/picca/pk1d/postproc_pk1d.py
```diff
@@ -862,12 +818,370 @@ def compute_average_pk_redshift(
...
+            * np.outer(std_pk, std_pk)
+            / np.sqrt(np.outer(covariance_diag, covariance_diag))
+        )
+
+    covariance_array = np.ravel(covariance_matrix)
+
+    return covariance_array
+
+
+def compute_cov_not_vectorized(
```
Waelthus, line 1181: this is not only not vectorized, but the actual complete old method, i.e. not only slower, but also doing slightly different things potentially by actually summing up each individual mode every time. Maybe clarify that in the docstring

py/picca/pk1d/postproc_pk1d.py
```diff
@@ -862,12 +818,370 @@ def compute_average_pk_redshift(
...
+
+    sum_p1d_weights = np.nansum(p1d_weights, axis=0)
+    weights_sum_product = np.outer(sum_p1d_weights, sum_p1d_weights)
+
+    p1d_groups_product_sum = np.zeros((nbins_k, nbins_k))
+    covariance_weights_product_sum = np.zeros((nbins_k, nbins_k))
+    weights_product_sum = np.zeros((nbins_k, nbins_k))
+
+    for i, p1d_group in enumerate(p1d_groups):
+        p1d_groups_product_sum = np.nansum(
```
Waelthus, line 1143: shouldn't all of those be +=? Else you only ever look at the last entry... And wouldn't need to pre-allocate before this step

py/picca/pk1d/postproc_pk1d.py
```diff
@@ -862,12 +818,370 @@ def compute_average_pk_redshift(
...
+    """
+
+    if len(p1d_groups) == 0:
+        return np.full(nbins_k * nbins_k, np.nan)
+
+    mean_pk = mean_p1d_table["meanPk"][(nbins_k * izbin) : (nbins_k * (izbin + 1))]
+    mean_pk_product = np.outer(mean_pk, mean_pk)
+
+    sum_p1d_weights = np.nansum(p1d_weights, axis=0)
+    weights_sum_product = np.outer(sum_p1d_weights, sum_p1d_weights)
```
Waelthus, line 1136: this is very confusing naming given that below you have `weights_product_sum`

py/picca/pk1d/postproc_pk1d.py
```diff
@@ -862,12 +818,370 @@ def compute_average_pk_redshift(
...
 def compute_cov(
+    weight_method,
+    mean_p1d_table,
+    nbins_k,
+    izbin,
+    p1d_weights,
+    covariance_weights,
+    p1d_groups,
+):
+    """Compute the covariance of a set of 1D power spectra.
```
Waelthus, line 1098: add some info regarding the approach to clarify what the 2 alternative routines are doing

py/picca/pk1d/postproc_pk1d.py
```diff
@@ -862,12 +818,370 @@ def compute_average_pk_redshift(
...
+        )
+        if number_worker == 1:
+            output_cov = [func(*p1d_group) for p1d_group in p1d_groups]
+        else:
+            with Pool(number_worker) as pool:
+                output_cov = pool.starmap(func, p1d_groups)
+
+        for izbin in range(nbins_z):
+            covariance_array = output_cov[izbin]
+            i_min = izbin * nbins_k * nbins_k
```
Waelthus, line 891: in principle some of those computations have been done earlier and one could pass them to this function from above (which would especially simplify changes in the structure of z-bins later on when we might want to use different numbers of k-bins per z-bin)

py/picca/pk1d/postproc_pk1d.py
```diff
@@ -862,12 +818,370 @@ def compute_average_pk_redshift(
...
+    Return
+    ------
+    covariance_array (array-like):
+    Array of covariance coefficients.
+    """
+
+    if len(p1d_groups) == 0:
+        return np.full(nbins_k * nbins_k, np.nan)
+
+    mean_pk = mean_p1d_table["meanPk"][(nbins_k * izbin) : (nbins_k * (izbin + 1))]
```
Waelthus, line 1132: given that one only really needs the relevant slice for the redshift bin in this function, why not pass that slice instead of izbin? That would avoid needing to define the slicing operation several times throughout this code...
Alternatively having functions computing the p1d_table_slice and the cov_table_slice would make the code more flexible later on (when we might deal with different number of k-bins per redshift), e.g.:
```
def p1d_table_slice(izbin, nbins=nbins_k):
     return slice((nbins * izbin), (nbins * (izbin + 1))
```

py/picca/pk1d/postproc_pk1d.py
```diff
@@ -862,12 +818,370 @@ def compute_average_pk_redshift(
...
+    p1d_sub_table = Table(
+        [
+            p1d_table["sub_forest_id"][select_z],
+            k_index[select_z],
+            p1d_table["Pk"][select_z],
+        ],
+        names=("sub_forest_id", "k_index", "pk"),
+    )
+    if weight_method == "fit_snr":
+        snrfit_z = snrfit_table[izbin * nbins_k : (izbin + 1) * nbins_k, :]
```
Waelthus, line 1012: see below/above regarding the redefinition of the redshift indexing

py/picca/pk1d/postproc_pk1d.py
```diff
@@ -862,12 +818,370 @@ def compute_average_pk_redshift(
...
+    covariance_matrix = (weights_product_sum / weights_sum_product) * (
+        (p1d_groups_product_sum / covariance_weights_product_sum) - mean_pk_product
+    )
+
+    # For fit_snr method, due to the SNR fitting scheme used for weighting,
+    # the diagonal of the weigthed sample covariance matrix is not equal
+    # to the error in mean P1D. This is tested on Ohio mocks and data.
+    # We choose to renormalize the whole covariance matrix.
+    if weight_method == "fit_snr":
+        std_pk = mean_p1d_table["errorPk"][(nbins_k * izbin) : (nbins_k * (izbin + 1))]
```
Waelthus, line 1168: see above

py/picca/pk1d/postproc_pk1d.py
```diff
@@ -862,12 +818,370 @@ def compute_average_pk_redshift(
...
+
+    p1d_weights, covariance_weights, p1d_groups = (
+        output_cov[:, 0, :],
+        output_cov[:, 1, :],
+        output_cov[:, 2, :],
+    )
+    return p1d_weights, covariance_weights, p1d_groups
+
+
+def compute_groups(nbins_k, p1d_los):
```
Waelthus, line 1050: this is a bit confusing naming, very close to the function just above, but actually the computing step performed inside for each subforest, maybe call it `compute_groups_for_one_forest` or similar

corentinravoux:
This is a very small example, like 2-3% of Y1, just for the test.
The main difference between the estimators is that now we are treated the weights properly in the normalization of all the covariance matrix, so it can the profile according the average shape of weights over redshift range. I think it is hard to interpret.
For "the strong correlation spikes around pixel [0,8]" I think this is due to the lack of modes that increases the stat error bar at very large scales. This can also be impacted by what you mention.

In Y1 run, associated to systematics: 
image

py/picca/pk1d/postproc_pk1d.py
```diff
@@ -862,12 +818,370 @@ def compute_average_pk_redshift(
...
+        snrfit_z = snrfit_table[izbin * nbins_k : (izbin + 1) * nbins_k, :]
+        p1d_sub_table["weight"] = 1 / fitfunc_variance_pk1d(
+            p1d_table["forest_snr"][select_z],
+            snrfit_z[k_index[select_z], 2],
+            snrfit_z[k_index[select_z], 3],
+        )
+    else:
+        p1d_sub_table["weight"] = 1
+
+    p1d_sub_table = p1d_sub_table[p1d_sub_table["k_index"] >= 0]
```
Waelthus, line 1021: are we storing negative k_index somewhere? Why is this cut needed?

py/picca/pk1d/postproc_pk1d.py
```diff
@@ -862,12 +818,370 @@ def compute_average_pk_redshift(
...
+
+    covariance_weights (array-like):
+    Weights for all subforest used inside the main covariance sum.
+
+    p1d_groups (array-like):
+    Individual p1d pixels grouped in the same wavenumber binning for all subforest
+    """
+
+    select_z = (p1d_table["forest_z"] < zbin_edges[izbin + 1]) & (
+        p1d_table["forest_z"] > zbin_edges[izbin]
```
Waelthus, line 1001: this is the `p1d_table` of each chunk and not the `mean_p1d_table` of the combined stats, so my comments regarding indexing shouldn't apply here...

py/picca/pk1d/postproc_pk1d.py
```diff
@@ -862,12 +818,370 @@ def compute_average_pk_redshift(
...
+    ]
+
+    del grouped_table
+
+    func = partial(
+        compute_groups,
+        nbins_k,
+    )
+    if number_worker == 1:
+        output_cov = [func(*p1d_los) for p1d_los in p1d_los_table]
```
Waelthus, line 1035: not sure if this should be `*p1d_los` or just `p1d_los`, I'd have guessed the latter, was this tested on a single core as well?

py/picca/pk1d/postproc_pk1d.py
```diff
@@ -862,12 +818,370 @@ def compute_average_pk_redshift(
...
+            with Pool(number_worker) as pool:
+                output_cov = pool.starmap(func, p1d_groups)
+
+        for izbin in range(nbins_z):
+            covariance_array = output_cov[izbin]
+            i_min = izbin * nbins_k * nbins_k
+            i_max = (izbin + 1) * nbins_k * nbins_k
+            cov_table["covariance"][i_min:i_max] = covariance_array
+
+    if compute_bootstrap:
```
Waelthus, line 895: is the bootstrap method tested? If yes, would be nice to know it's results, if no please at least add comments that this isn't tested yet...

Waelthus (changes requested):
This looks fine overall, needs somewhat more commenting, and it would be good to have some of the selections defined in one place for flexibility/easier maintanence

py/picca/pk1d/postproc_pk1d.py
```diff
@@ -91,9 +90,9 @@ def read_pk1d(filename, kbin_edges, snrcut=None, zbins_snrcut=None):
             chunk_table["forest_snr"] = float(chunk_header["MEANSNR"])
             chunk_table["forest_id"] = int(chunk_header["LOS_ID"])
             if "CHUNK_ID" in chunk_header:
-                chunk_table[
-                    "sub_forest_id"
-                ] = f"{chunk_header['LOS_ID']}_{chunk_header['CHUNK_ID']}"
+                chunk_table["sub_forest_id"] = (
+                    f"{chunk_header['LOS_ID']}_{chunk_header['CHUNK_ID']}"
+                )
```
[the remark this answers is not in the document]
corentinravoux: Yes sorry, I made a run of "black" inside on of the commit. So at least now this is all formatted !

Waelthus:
> In Y1 run, associated to systematics (I will show it at the upcoming DESI meeting)

Ouch, this looks much more correlated than I expected... Can you chainge the colormap so that it actually goes from -1 to 1 and maybe also that it's white/yellow for 0 correlation and diverging outside? I looked at it and at first glance thought green is zero, but that isn't actually right and it looks like except when comparing the largest and smallest modes available things are >50% correlated everywhere

py/picca/pk1d/postproc_pk1d.py
```diff
@@ -345,111 +345,66 @@ def compute_mean_pk1d(
...
+            kbin_centers = (kbin_edges[1:] + kbin_edges[:-1]) / 2
+            k1_array, k2_array = np.meshgrid(kbin_centers, kbin_centers, indexing="ij")
+            k1_array = np.ravel(k1_array)
+            k2_array = np.ravel(k2_array)
+            n_array = np.ravel(
+                np.outer(
+                    mean_p1d_table["N"][(nbins_k * izbin) : (nbins_k * (izbin + 1))],
+                    mean_p1d_table["N"][(nbins_k * izbin) : (nbins_k * (izbin + 1))],
+                )
+            )
```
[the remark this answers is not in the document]
corentinravoux: - Yes this is cleaner, I will change it
- Yes we have the same k binning, to counter the effect you are mentioning, we take a very high value of kmax for all z (well above the resolution cut). It has the advantage to have a consistent format of covariance matrix. I guess it is ok to keep it this way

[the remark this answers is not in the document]
Waelthus: ah, that means we're calculating for very large k even if that is unusable and cut down the k-ranges to more reasonable values later? That would be totally fine and would simplify the coding here (on the expanse of a little more computations at low-z, but given the speed improvement from this PR it shouldn't be an issue)...

py/picca/pk1d/postproc_pk1d.py
corentinravoux: The += method would not deal with np.nan and they are a lot of np.nan inside the individual p1d. At the end the covariance matrix would probably have a lot of np.nan.
Here, we the nansum, we sum iteratively, it should contain all entry

py/picca/pk1d/postproc_pk1d.py
corentinravoux: Yes but we have indeed a sum of product and a product of sums. I renamed them with the of, it should be a little bit clearer

py/picca/pk1d/postproc_pk1d.py
corentinravoux: Added

py/picca/pk1d/postproc_pk1d.py
corentinravoux: That is what we are doing on the new method too, no ?
I have added a comment on this function

py/picca/pk1d/postproc_pk1d.py
corentinravoux: Do you mean the "i_min = izbin * nbins_k * nbins_k" ? 
I have only put that way for clarity, but we can directly put it inside the cov_table.
I am unsure that we would benefit a lot from a varying k grid (see my previous comment).

py/picca/pk1d/postproc_pk1d.py
corentinravoux: But you are right concerning the first comment, I will pass the two slices

py/picca/pk1d/postproc_pk1d.py
corentinravoux: done

py/picca/pk1d/postproc_pk1d.py
corentinravoux: This was a convention for the case when a k bin computed for the p1d does not fall inside the output k binning. I removed those bins as they do not contribute to the covariance.

corentinravoux: This cut is needed because I made a method to compute the p1d groups in a vectorized way, and -1 change the last k bin, but it should not change any bin

py/picca/pk1d/postproc_pk1d.py
corentinravoux: this is *p1d_los because p1d_los is a table with 5 array, that we give to func. Yes I tested it

py/picca/pk1d/postproc_pk1d.py
corentinravoux: The bootstrap method works and gives very similar results than the classical covariance, which makes sense considering the number of chunks.
All the Y1 plot I am showing is with bootstrap

py/picca/pk1d/postproc_pk1d.py
Waelthus: yes, but in the old one the approach is split in a completely different way by looping over every mode while in the new we're precomputing intermediate summaries that are then reused multiple times, and in addition do everything in a vectorized way...

py/picca/pk1d/postproc_pk1d.py
Waelthus: oops, I actually wanted to change that statement after I reread the code and saw what is actually done. Please add a comment explaining this not so common way of summing. Agreed that the nans need to be removed and that this is likely the simplest way...

py/picca/pk1d/postproc_pk1d.py
Waelthus: maybe adding what is summed into the name could help, but not sure...

py/picca/pk1d/postproc_pk1d.py
Waelthus: oh, with the function definition I just wanted to introduce something that would work as the current implementation, but that could be changed at a single place easily later if the approach should be changed...

py/picca/pk1d/postproc_pk1d.py
Waelthus: the complaint was not regarding the code in this line, but mostly regarding the repetition of lines like this throughout the file, see below for similar cases

py/picca/pk1d/postproc_pk1d.py
Waelthus: I still don't fully get how this line is changing anything. Wouldn't the selector be always true? Or are there nans or -1s in that table for bins that just aren't filled and that you want to remove here (in which case I would test for not having exactly that filler value and not via >=0)?

py/picca/pk1d/postproc_pk1d.py
Waelthus: but why isn't that needed below in the mapped version, shouldn't that be a starmap then or similar? But I guess if it works it works...

Waelthus:
maybe we should also add a line for covar computation here:
https://github.com/igmhub/picca/blob/02f6c2e61a204254e59e43bbcaecd596d2f8a418/py/picca/tests/test_2_pk1d.py#L88-L97
and compare the output to a saved state. That would at least make sure this code is run regularly (e.g. useful for library version changes) and we don't change stuff by accident, even if the covar of so few spectra will be garbage...

Waelthus (approved):
Given that this PR seems to do what it's supposed to, I'd be fine with merging. Please consider adding comments to the code where there are still semi-open questions (so that we e.g. don't wonder why we did certain steps in the future and remove those). Also please consider adding a line to the test in your favourite config with the relevant output (or send me the line you'd like to test and I'd add it).

py/picca/pk1d/postproc_pk1d.py
corentinravoux: Added

py/picca/pk1d/postproc_pk1d.py
corentinravoux: Ah yes this is indeed a good way, it is done now.

py/picca/pk1d/postproc_pk1d.py
corentinravoux: With the implementation right now, this is just to remove p1d pixels which were not associated with any wavenumber bin. 
I made a method previously for which this cut was necessary, but now it is just to reduce a little the input dataset, removing unecessary pixels.
Comment added.

corentinravoux:
Comments and tests added, waiting for pytest, if successful, let's merge it

corentinravoux:
Here since the bootstrap is just a run of the covariance matrix, it is not essential to test it. 
But yes the seed for the bootstrap can be added later.
I think we can merge now

Issue #1062, opened 2024-03-13
Test numpy 2.0 compatibility or pin to 1.x


Waelthus:
Numpy 2.0 is now in beta and will have ABI breaking changes, i.e. not be fully backwards compatible. 
See current release notes here: https://github.com/numpy/numpy/releases/tag/v2.0.0b1

As we rely on both numpy and other packages depending on numpy, this will likely require us to use the old version until requirements are updated. The testing will likely fail after the update as well...

PR #1069, opened 2024-07-09
Fixed error when package downloaded from tarball instead of cloned

iprafols:


iprafols:
Errors are occurring due to numpy v2.0 as predicted in #1062

iprafols:
Correction: the error was fixed with PR #1066. We should be ready to merge

Waelthus (approved):
sounds good

PR #1070, opened 2024-07-09
Update numpy2.0 [bump minor]

iprafols:
Given the recent PR #1066  it seems that there are no backward compatibility issues with numpy (at least the tests are passing locally). Thus, I think we should update to numpy 2.0. I also updated the other requirements and simplified the github workflow (I think the issue with fitsio installation was fixed, let's see if the automatic tests confirm this).

This should fix issue #1062


iprafols:
Ok, it seems we need to wait for the server to update something as it is not finding numpy2.0

corentinravoux:
Hi @iprafols, do you want us to run specific installation tests on this PR ?
Considering the test automatically run, it seems like it is working

Waelthus:
Sounds fine with me, just fixed the linting that was still using python 3.8 and thus couldn't install numpy 2.0.
Are there any important parts of the code left that do not automatically tested and which should be double checked manually?

Waelthus (approved):

iprafols:
I don't think so, I guess we can just merge

Issue #1063, opened 2024-05-07
Possible issue with rebinning


iprafols:
@andreicuceu noted that there is potentially an issue with the way we are doing the rebinning.

Original message by Andrei:
I think there's a small bug in picca delta extraction when defining the wavelength grid. These are true continuum deltas for the same forest, run with both picca and qsonic. You can see that for mocks the grid is slightly shifted compared to the data, so the bin centers are at 3599.9, 3600.7, etc.  instead of 3600.0, 3600.8, etc.. QSOnic has the correct grid, while picca shifts everything by 0.1A to the data grid. This is not a big deal, but it would be good to fix both the bug in picca and the shifted grid in QQ.

![issue](https://github.com/igmhub/picca/assets/9192975/966f4734-2d6f-4a56-972b-e65dcb77aaca)

iprafols:
This is not a PICCA issue, that PICCA behaves as expected, but rather a quick quasars issue. While we wait for QuickQuasars to be fixed, it suffices to specify lambda min = 3599.9 and  lambda max = 5499.9 in the [data] section in the delta extraction config

PR #1071, opened 2024-07-11
Fast desihealpix [bump minor]

iprafols:
There are two fixes here:

1. I fixed the wavelength limits in the desi mocks tests according to issue #1063 
2. I added fast reading modes. For now, they are limited to the default BAO options but are ~25% faster on automatic tests (achieving the same results to machine precision)

Before merging this PR, I will assess the speed increase on a run with y1 mocks, and make sure the results are indeed stable. I will report the findings here


iprafols:
I have finished running on a set of mocks for both the normal and the new fast modes. For this test, I used the mocks stored in `/global/cfs/projectdirs/desi/users/acuceu/mock_links/colore_qq_y1/v9.0.0.0.0`. Results of the tests are stored at `/global/cfs/projectdirs/desi/users/iprafols/lya_working_group/picca/fast_healpix_speed_test`

The reading time dropped from 80.2197 to 35.0075. This corresponds to a decrease of 56% of the reading time, twice as much as observed in the automatic tests, meaning scaling of the changes is good. Total run time for the fast mode was 1352.0694 for the fast mode and is *not* dominated by the reading time, but rather the continuum fitting.

Results are the same to machine precision excepf for the following:

corentinravoux:
Hi @iprafols do you want me to test this PR or you still need to work on the pixel issues you were mentioning ?
Maybe we need to discuss this in TG telecon before.

iprafols:
We definitely need to discuss this in a TG discussion. I am still finishing to compile a report of the small differences I found when running on the mocks

Waelthus:
After that TG discussion I guess it would also be useful to port the changes to tile-based mode, it's not used as much, but the speedup would be even more valuable there and I guess we'd want to have the ability to run on e.g. daily...

iprafols:
I have finished running on a set of mocks for both the normal and the new fast modes. For this test, I used the mocks stored in /global/cfs/projectdirs/desi/users/acuceu/mock_links/colore_qq_y1/v9.0.0.0.0. Results of the tests are stored at /global/cfs/projectdirs/desi/users/iprafols/lya_working_group/picca/fast_healpix_speed_test. Note that for this test I was using a minimum number of pixels in the forest of 50 (instead of the 150 that we use in the baseline for BAO analysis)

The reading time dropped from 80.2197 to 35.0075. This corresponds to a decrease of 56% on the reading time, twice as much as observed in the automatic tests, meaning scaling of the changes is good. The total run time for the fast mode was 1352.0694 for the fast mode and is not dominated by the reading time, but rather the continuum fitting.

Results are the same to machine precision except for the following:
- In the FIT_METADATA header of the delta_attributes.fits.gz files there are 8 quasars with different fitted SLOPE. This is not observed in delta_attributes1.fits.gz, and delta_attributes2.fits.gz files. In delta_attributes3.fits.gz only 3 of the lines of sight show this difference and in delta_attributes4.fits.gz 12 quasars show this difference, plus 15 show different ZERO_POINT. 
- I also checked for differences in the individual forests and found 59 forests with some differences. They are short forests, and only 13 of them have more than 150 pixels. Except for one case (with forest size 124 pixels), where the differences appear for all pixels, only a few pixels are affected. Overall I observed 440 different values of DELTA, 218 of CONT, and 143 of WEIGHT. For forests with length larger than 50 pixels these numbers decrease to 109, 0 and 10 respectively.


The LOS_ID of the 8 forests with different slope are: 68109, 30166004, 40050004, 40137619, 40143207, 50132006, 50176817, 180313901. Surprisingly only half of them appear in the table below.

See below the list of offending forests:
| LOS_ID       |  HEALPIX | forest_length | affected DELTA pixels | affected CONT pixels | affected WEIGHT pixels |
| ----------- | --------- | -------------- | ---------------------- | --------------------- | ------------------------ |
| 60020083 | 102 | 94 | 1 | 0 | 0 |
| 50015884 |  104 | 179 | 1 | 0 | 0 |
| 260009980 | 105 | 117 | 1 | 0 | 0 |
| 260035604 | 121 | 96 | 1 | 0 | 0 |
| 210344440 | 126 | 148 | 2 | 0 | 0 |
| 100162382 | 129 | 109 | 1 | 0 | 0 |
| 50078540 | 134 | 129 | 1 | 0 | 0 | 
| 260079301 | 136 | 109 | 15 | 15 | 0 |
| 90154823 | 163 | 137 | 1 | 0 | 0 |
| 30009517 | 164 | 97 | 1 | 0 | 0 |
| 60119410 | 165 | 132 | 1 | 0 | 0 |
| 60105137 | 165 | 181 | 1 | 0 | 0 |
| 80194913 | 178 | 94 | 1 | 0 | 0 |
| 240086662 | 186 | 67 | 3 | 0 | 0 |
| 80134130 | 193 | 127 | 1 | 0 | 0 |
| 310161353 | 199 | 77 | 1 | 0 | 0 |
| 40122766 | 230 | 113 | 1 | 0 | 0 |
| 300161877 | 231 | 72 | 1 | 1 | 0 |
| 180291076 | 250 | 144 |  1 | 0 | 0 |
| 170329056 | 253 | 243 | 1 | 0 | 0 |
| 300220736 | 262 | 69 | 5 | 3 | 0 | 
| 140335642 | 271 | 142 | 1 | 0 | 0 |
| 180313901 | 283 | 262 | 1 | 0 | 0 |
| 180330644 | 283 | 190 | 1 | 0 | 0 |
| 160301347 | 284 | 147 | 1 | 0 | 0 |
| 300029079 | 29 | 119 | 1 | 0 | 0 |
| 120274629 | 304 | 116 | 1 | 0 | 0 |
| 70233106 | 305 | 75 | 2 | 1 | 0 |
| 70215222 | 306 | 69 | 7 | 6 | 0 |
| 160360137 | 316 | 143 | 1 | 0 | 0 |
| 110265872 | 317 | 87 | 2 | 0 | 0 |
| 120319628 | 319 | 303 | 93 | 0 | 10 |
| 97247 | 323 | 57 | 3 | 2 | 0 |
| 44073 | 323 | 115 | 1 | 0 | 0 |
| 270192708 | 326 | 141 | 1 | 0 | 0 |
| 120267719 | 335 | 79 | 1 | 0 | 0 |
| 100247166 | 349 | 87 | 1 | 0 | 0 | 
| 90277518 | 351 | 100 | 1 | 0 | 0 | 
| 120361506 | 352 | 108 | 94 | 47 | 4 | 
| 60140831 | 353 | 96 | 1 | 0 | 0 |
| 50176778 | 371 | 201 | 1 | 0 | 0 |
| 50218741 | 371 | 78 | 1 | 0 | 0 |
| 250179886 | 374 | 148 | 20 | 2 | 5 | 
| 210102906 | 379 | 68 | 3 | 1 | 0 | 
| 210038251 | 379 | 206 | 5 | 0 | 0 | 
| 100298453 | 380 | 63 | 11 | 9 | 0 |
| 50176817 | 387 | 251 | 1 | 0 | 0 |
| 250215707 | 390 | 104 | 1 | 0 | 0 | 
| 90235601 | 399 | 84 | 1 | 0 | 0 |
| 150024109 | 404 | 164 | 1 | 0 | 0 |
| 40143207 | 418 | 318 | 1 | 0 | 0 |
| 150382834 | 420 | 157 | 1 | 0 | 0 |
| 150373033 | 420 | 115 | 1 | 0 | 0 |
| 90339938 | 431 | 118 | 1 | 0 | 0 |
| 40174900 | 433 | 88 | 8 | 7 | 0 |
| 80283171 | 445 | 132 | 1 | 0 | 0 |
| 30166004 | 448 | 124 | 124 | 124 | 124 |
| 130372412 | 469 | 219 | 1 | 0 | 0 |
| 280018349 | 69 | 79 | 1 | 0 | 0 |

My conclusion from this is that the difference is driven by some fit instabilities and has no impact on the measured correlation functions. To be discussed in a telecon: Maybe we should increase the cut to 300 pixels? 

I will rerun the test using a minimum of 150 pixels to see if this changes the results

Waelthus:
I'm still a bit puzzled by those differences. I think they are fine overall and as you said could be fitting instabilities. But it's weird that when changing only the initial reading of data, the first observable change occurs at the third fitting iteration. With good agreement at the first 2 stages (is that machine precision [single or double] or something less?) it doesn't sound like it should be just amplification of numerical errors (but that depends how well the agreement is before). Could be an effect of having read objects in a different order and thus performing the fits in a different order though.

Would it make sense to e.g. sort/fix the order of objects (in a random, but predeterminable way) before the fitting stage, such that they are processed in a predictable order independent of how they're read? If that was easy, one could also have a seed for that random order allowing to reproduce things exactly or estimate the error budget from fitting instabilities by trying different seeds. This would also allow varying or fixing the order of fitting between iterations.
Just an idea to potentially stabilize those fits...

iprafols:
I have re-run on the same mocks but now requiring a minimum of 150 pixels in the forest. Now the delta_attributes files are all the equal except for the latest iteration, where we observe some differences.

iprafols:
Maybe we can get more stable results by changing the tolerance parameter in iMinuit initialization. From iMinuit [documentation](https://scikit-hep.org/iminuit/reference.html#iminuit.Minuit.tol):

```
propertytol: float
Access tolerance for convergence with the EDM criterion.

Minuit detects converge with the EDM criterion. EDM stands for Estimated Distance to Minimum, 
it is mathematically described in the [MINUIT paper](https://doi.org/10.1016/0010-4655(75)90039-9). 
The EDM criterion is well suited for statistical cost functions, since it stops the minimization when 
parameter improvements become small compared to parameter uncertainties.

The convergence is detected when edm < edm_max, where edm_max is calculated as

Migrad: edm_max = 0.002 * tol * errordef

Simplex: edm_max = tol * errordef

Users can set tol (default: 0.1) to a different value to either speed up convergence at the cost of a 
larger error on the fitted parameters and possibly invalid estimates for parameter uncertainties or 
smaller values to get more accurate parameter values, although this should never be necessary as 
the default is fine.

If the tolerance is set to a very small value or zero, Minuit will use an internal lower limit for the 
tolerance. To restore the default use, one can assign None.

Under some circumstances, Migrad is allowed to violate edm_max by a factor of 10. Users should 
not try to detect convergence by comparing edm with edm_max, but 
query [iminuit.util.FMin.is_above_max_edm](https://scikit-hep.org/iminuit/reference.html#iminuit.util.FMin.is_above_max_edm).
```

iprafols:
I have run a test on Jura data. The reading time dropped from 1896.9065821170807 to 280.78776836395264. I checked the logs and found similarly small discrepancies. In the delta attributes file, there are 8 values of the ZERO_POINT in HDU FIT_METADATA that are different. Differences are of order 1e-8 - 1e-7 and relative differences are of order 1e-5. There are also 7 values of the SLOPE that have similar differences. The delta_attributes_1.fits to delta_attribute3.fits do not present any differences. Delta_attributes4.fits has 10 values different in both the SLOPE and the ZERO_POINT with similar discrepancies. I still have to check all the individual deltas, but this will take some time, and I don't see any problem with merging this already 

I just need someone to review the changes @corentinravoux @calumgordon @andreicuceu @Waelthus

corentinravoux:
I have added few comments on the code. The main thing that need to be done right now is to add a Warning or Exception in the case when Fast methods are run with P1D mode.
Other than that, since the reading is done in a complete separate way, I think it is good to be merged. I do not consider the numerical instabilities you show as concerning.

Waelthus:
I'd be fine with it as is. And agree with Corentin that an Exception should be thrown when trying to use fast reading with P1d mode with (given it is not implemented yet). 
Zero point changes by relative 1e-5 looks like the order of magnitude is something around single_precision_accuracy * num_pix_per_forest, but we always had instabilities of that order, they are hard to debug, not changing the overall results, and no-one has the time to root them out. So agreeing that those are of little concern, although I must admit that I'd prefer not to have numerical instabilities on that level...
So green light from my side.

Waelthus (approved):
I just added the Exception.

PR #1072, opened 2024-07-30
Average mean pk

corentinravoux:
For Y1 mocks, it is not possible to compute the average P1D of the 20 mocks (it hits maximal memory).
Implementation of an averaging of mean P1D + one script

### DIFF
--- a/bin/picca_Pk1D_average_mean.py
+++ b/bin/picca_Pk1D_average_mean.py
@@ -19,7 +19,7 @@ def main(cmdargs):
         "--in-dir",
         type=str,
         default=None,
-        required=True,
+        required=False,
         help="String to glob all mean p1d to average",
     )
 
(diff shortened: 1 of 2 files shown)


bin/picca_Pk1D_average_mean.py
```diff
@@ -0,0 +1,76 @@
...
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter,
+        description="Compute the averaged 1D power spectrum averages",
+    )
+
+    parser.add_argument(
+        "--in-dir",
+        type=str,
+        default=None,
+        required=False,
+        help="String to glob all mean p1d to average",
```
Waelthus, line 23: `in-dir` would normally mean you put in a directory name and a predefined glob will be running within that dir, here it looks like a full glob-expression needs to be supplied, which I think is ok as well, but maybe should use a different name...

Waelthus (approved):
I'd be fine with those changes if you are, basically it's just running the avg in the end for the combine plus some compat tests between input files.
If this isn't tested yet, maybe consider adding a test.

PR #1074, opened 2024-09-12
[bump minor] add IVAR to all forest outputs

Waelthus:
Adding the IVAR to outputs. Tests aren't fixed yet.


Waelthus:
all tests but script tests are passing now (at least locally on my machine), so updating the offending test files (delta files for BAO need to have the IVAR in the tests), once those are fixed things should be ready to go.

Waelthus:
Tests should be fine after merging #1075. 
Note that this change would always add the IVAR to delta files, if we want to make it optional, additional logic and tests would be needed. Someone should also change the class diagrams accordingly such that Forest objects have (potentially optional) IVAR output and it's not a specialty of only Pk1dForests anymore...
Elsewise this would be ready to merge from my side.

iprafols (approved):
Things look good, I'd say to merge this

corentinravoux:
The PR looks fine ! In any case I think this is a good thing to always add IVAR, I use it also for the Lya tomography.

Waelthus:
bumping at least the minor version as the data format changes, even if only additions happen

calumgordon (commented):
Looks good to me too.

PR #1075, opened 2024-09-12
Fix testing

Waelthus:
There has been a deprecation on one of the github action modules storing artifacts when tests fail. That leads all tests to be failing before updating to a newer version which happens in this PR.


Waelthus:
seems like the fix is actually working and tests pass again, so ready to merge whenever someone approves.

iprafols (approved):

PR #1076, opened 2024-09-18
[bump minor] Change PyPI action

Waelthus:
This changes the github action to use Trusted Publisher functionality instead of a token linked to my account (https://docs.pypi.org/trusted-publishers/using-a-publisher/).


Waelthus:
the version bump is added so that releasing is actually triggered

iprafols (approved):
Great! I was also checking how to use the PyPi trusted publishers mechanism. I think this fixes the previous patch. I'd merge once the tests finish

PR #1077, opened 2024-09-26
added missing command [bump minor]

iprafols:


iprafols:
The pylint tests were failing in the first commit because of the new pylint release. Github actions are now using pylint 3.3  which has new checks, concerning the maximum number of positional arguments. I modified the config file to increase its default (as we did previously with the maximum number of arguments). Because the tests will keep failing until we do so, I added the [bump minor] tag to the PR

calumgordon (approved):

PR #1079, opened 2024-10-23
Redshift evolution in dmat

julienguy:
Adding redshift evolution in the distortion matrix calculation both for auto and cross-correlation.

The matrix elements are scaled with ((1+z1)/(1+zref))^(alpha-1)*((1+z2)/(1+zref))^(alpha2-1).
- this redshift scaling coefficient is part of the contribution of each pair in the numerical integrals (so it is properly weighted).
- alpha, alpha2 are the bias evolution power law indices.
- alpha2=alpha=2.9 for lya auto, alpha2=alpha_obj=1.44 in lyaxQSO. 
- can be set to different values with options in the scripts.
- zref is computed from the data set. It is returned in the ATTRI HDU table column Z as it was before. 
- this redshift column Z is read by vega and defines the redshift of the undistorted model (that is then multiplied by the distortion matrix).

This modification to the distortion matrix improves the fit of colore mocks.

### DIFF
--- a/bin/picca_dmat.py
+++ b/bin/picca_dmat.py
@@ -245,6 +245,9 @@ def main(cmdargs):
                         help='Rebin factor for deltas. If not None, deltas will '
                              'be rebinned by that factor')
 
+    parser.add_argument('--no-redshift-evolution',
+                        action='store_true',
+                        help='Ignore redshift evolution when computing distortion matrix')
 
     args = parser.parse_args(cmdargs)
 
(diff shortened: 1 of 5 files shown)


julienguy:
Tests are failing because the distortion matrices are now different. This reminds me I need to add an option to fall back to the previous version. I will do that.

andreufont (approved):
These look good, thanks!

PR #1084, opened 2024-11-04
[bump minor] Update publish-to-test-pypi.yml

### LINKED ISSUE #1083: Issues in automatic upload to PyPI
@corentinravoux @calumgordon It looks like since version 6.7.0 the upload to PyPI is broken so pip install picca will install v6.6.1 at the moment. Not sure what exactly broke it, could either be one of the changes between commit 30a2530a42162e78f5a30762d3ba547a54eadac0 and commit 6be220021cef1ef926168fdf52a0550037c0e15e, i.e. from PR #1081. Could also be a change on the Github side, making the previous setup unusable...

Waelthus:
This should hopefully fix #1083. Bumping the version so that the rule is triggered on merge

### DIFF
--- a/.github/workflows/bump_version.yml
+++ b/.github/workflows/bump_version.yml
@@ -13,7 +13,7 @@ jobs:
   bump_version:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v4
       with:
         token: ${{ secrets.PAT_MW_TOKEN_ACTION }}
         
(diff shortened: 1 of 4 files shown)


Waelthus:
updated versions of github actions as well, the previous versions were based on nodejs v16, leading to deprecation warnings like `The following actions use a deprecated Node.js version and will be forced to run on node20: actions/checkout@v3, actions/setup-python@v4.` which would turn into errors at some point I guess.

corentinravoux:
Hi Michael, this is mainly setting up the work flows and actions in github right ? 
I never used it so I am trusting you on this, especially if this is fixed

corentinravoux (approved):

Waelthus:
I'll merge this now and we'll see if it fixes the issue

PR #1085, opened 2024-11-05
[bump minor] Fix covariance p1d

corentinravoux:
New estimator derived by Jean-Marc.
Few fix about snr and redshift weights


py/picca/pk1d/postproc_pk1d.py
```diff
@@ -1212,21 +1221,21 @@ def compute_cov(
...
-            [weights_product_of_sum, np.outer(p1d_weights[i], p1d_weights[i])], axis=0
+        weights_product_sum = np.nansum(
+            [weights_product_sum, np.outer(p1d_weights[i], p1d_weights[i])], axis=0
         )
 
     del p1d_groups, covariance_weights, p1d_weights
 
-    covariance_matrix = (weights_product_of_sum / weights_sum_of_product) * (
-        (p1d_groups_product_sum / covariance_weights_product_of_sum) - mean_pk_product
+    covariance_matrix = ((weights_sum_product /weights_product_sum) - 1)**(-1) * (
```
Waelthus, line 1237: the -1 in the factor here seems to be the second actual change (i.e. the correction for finite sample size aka 1/(N-1) factors in variance calculations)

py/picca/pk1d/postproc_pk1d.py
```diff
@@ -1142,7 +1151,7 @@ def compute_groups_for_one_forest(nbins_k, p1d_los):
         if number_in_bins != 0:
             weight = p1d_los["weight"][mask_ikbin][0]
             p1d_weights_id[ikbin] = weight
-            covariance_weights_id[ikbin] = np.sqrt(weight / number_in_bins)
+            covariance_weights_id[ikbin] = weight / number_in_bins
```
Waelthus, line 1154: This changes the weights to be proper weights and not sqrts? Most other changes are just renaming things...

py/picca/pk1d/postproc_pk1d.py
Waelthus: Overall I think this code could have some simplification/clarification. Maybe we should add the covariance document to picca and explicitely mention which quantities are which with equation numbers etc as comments.

py/picca/pk1d/postproc_pk1d.py
```diff
@@ -820,15 +824,20 @@ def compute_average_pk_redshift(
...
+                else:
+                    mean = np.average(data_values, weights=weights)
                 if apply_z_weights:
                     # Analytic expression for the re-weighted average:
                     error = np.sqrt(np.sum(weights_col * redshift_weights)) / np.sum(
                         weights_col
                     )
                 else:
                     error = np.sqrt(1.0 / np.sum(weights_col))
+                    # Variance estimator derived by Jean-Marc, we keep the estimated one.
```
Waelthus, line 837: are we now using Jean-Marc's estimator here, and is that just the same as before? Or are we using something else?

corentinravoux: This was just a test and I saved it, but we keep the same as before for now. Honestly it does not change a lot the level of the statistical error.

py/picca/pk1d/postproc_pk1d.py
corentinravoux: Yes this is the second change, which gives a unbiased version of the covariance. Same here no big impact. In the end the implementation and estimator will be detailled in the Y1 validation paper

Waelthus: the issue for me is not where exactly the documentation is, but at least on first glance it's somewhat hard to follow the code even looking at JMs note next to it. So we should probably try to streamline/comment it more in the relevant parts, to help the next person (or your future self) getting here. E.g. sum_product and product_sum are somewhat easily confused (but I don't have a great suggestion for other naming conventions which is why I suggested comments).
That said, the change from the biased to unbiased estimator is finished in here, so lets merge whenever you're fine to do so

Waelthus (approved):

py/picca/pk1d/postproc_pk1d.py
Waelthus: so the estimator by JM (on which we base the covar output as well) is different from the estimator actually used. I think this is sub-optimal and we might want to use JMs in the future. We'd definitely need to document the estimator actually used in the papers

corentinravoux: Yes the two estimators are differents but I am renormalizing the covariance matrix so that the diagonal is the same than our estimator. It mainly impact very small scales, e.g. on a small test example: 
image

corentinravoux: Right now we are cutting k < 2.0 AA^-1

Waelthus:
It's still weird to have matching statistical uncertainty estimates via covar and bootstrap but use a different estimator as baseline. Can you remind me where this came from? But fine as long as we document it... 

Am 6. November 2024 14:46:20 MEZ schrieb corentinravoux ***@***.***>:
>@corentinravoux commented on this pull request.
>
>
>
>>                  if apply_z_weights:
>                     # Analytic expression for the re-weighted average:
>                     error = np.sqrt(np.sum(weights_col * redshift_weights)) / np.sum(
>                         weights_col
>                     )
>                 else:
>                     error = np.sqrt(1.0 / np.sum(weights_col))
>+                    # Variance estimator derived by Jean-Marc, we keep the estimated one.
>
>Right now we are cutting k < 2.0 AA^-1
>
>-- 
>Reply to this email directly or view it on GitHub:
>https://github.com/igmhub/picca/pull/1085#discussion_r1831046933
>You are receiving this because your review was requested.
>
>Message ID: ***@***.***>
-- 
Sent from my phone...

corentinravoux:
The estimator of variance that we use is basically the results of the global fit Marie-Lynn is performing on all the individual P1D (dependence of classical standard deviation as a function of wavenumber and SNR). Once the fit is realized, the weights for the averaging are defined from 1/sigma(SNR, k)**2  where sigma is the evaluation of the fit for the SNR and k considered, and the variance is the direct result of the fit. 
That is what we choose. So here the variance is the result of the fit, not an estimator.
Again, the difference between those two is small,

corentinravoux:
Also, I tested with the estimator of JM for the variance, there is still a small difference with the covariance estimator on the diagonal. I do not know if this is statistical effect or not

Waelthus:
ok, so it's just SNR weighting of the individual spectra based on that fit instead of running an estimator on the data directly... I guess that also explains why mostly high-k are different. So fine to merge from my perspective

PR #1087, opened 2024-11-21
add reading of exposure info from coadds

Waelthus:
This branch actually reads a bunch of exposure metadata from the coadds and saves them into the final deltas. Tested it on a small dataset. Currently only implemented in FastHealpix mode and still needs proper tests for a merge.

### DIFF
--- a/py/picca/delta_extraction/astronomical_objects/desi_forest.py
+++ b/py/picca/delta_extraction/astronomical_objects/desi_forest.py
@@ -69,9 +69,19 @@ class DesiForest(Forest):
         del kwargs["targetid"]
 
         self.tile = []
-        if kwargs.get("tile") is not None:
-            self.tile.append(kwargs.get("tile"))
-            del kwargs["tile"]
+        if kwargs.get("tileid") is not None:
+            self.tile.append(kwargs.get("tileid"))
+            del kwargs["tileid"]
+
+        self.fiber = []
+        if kwargs.get("fiber") is not None:
+            self.fiber.append(kwargs.get("fiber"))
+            del kwargs["fiber"]
+
+        self.expid = []
+        if kwargs.get("expid") is not None:
+            self.expid.append(kwargs.get("expid"))
+            del kwargs["expid"]
 
         # call parent constructor
         kwargs["los_id"] = self.targetid
@@ -99,6 +109,8 @@ class DesiForest(Forest):
         self.night += other.night
         self.petal += other.petal
         self.tile += other.tile
+        self.expid += other.expid
+        self.fiber += other.fiber
         super().coadd(other)
 
     def get_header(self):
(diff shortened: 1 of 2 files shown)


corentinravoux:
Hi @Waelthus, 
Is this branch the new branch you were mentioning in the other PR ? in that case, what is the current status ?

Waelthus:
not sure which other PR you mean.
What this PR does is adding additional metadata into output files, not sure if that has been done some other way in the meantime.
Testing is expected to fail as output files are not updated...

corentinravoux:
Ah ok sorry, I thought this was a new branch for https://github.com/igmhub/picca/pull/1060 which diverged too much.
Ok in that case, we will probably try to discuss this PR in a software telecon

julienguy:
I merged master into this branch, and was able to run it without trouble.
It provides the following information about each skewer in the METADATA table:
```
['LOS_ID', 'RA', 'DEC', 'Z', 'MEANSNR', 'TARGETID', 'NIGHT', 'PETAL', 'TILEID', 'EXPID', 'FIBER']
```
For spectra with several exposures, NIGHT, PETAL, TILEID, EXPID, FIBER are given as a string with dash-separated integer numbers. Like `'265-249-1394'` for one FIBER entry. I think a comma separated list would be better (allowing for negative integer numbers for instance if it ever happens) but except for that it's very useful and should be merged in my opinion. 
I am going to make the changes ("," instead of "-" as a separator).

julienguy:
It is still work in progress because it is difficult to pass all the tests for files without EXP_FIBERMAP (as is the case for simulations that we do not want to break).

Waelthus:
Currently the linter complains about code-duplication, this apparently can only be ignored globally (adding duplicate-code to the list of disabled rules) or via refactoring (which I'm not sure if it's useful here).
The test suite fails in three ways for me:
1) OSErrors in the ExpectedFluxTests and DataTests as the input files partially have different extensions than expected from them in the updated header extraction approach, i.e. no EXP_FIBERMAP as Julien noted
2) AssertionErrors in some ScriptsTests (e.g. test_delta_calib) as the outputs of the end-to-end tests have changed and need to be regenerated with the updated headers (can be done easily once 1 is fixed)
3) subprocess.CalledProcessErrors on some ScriptsTests (e.g. test_delta_lin),  not sure what those are exactly (and if they only affect my local machine or would be fixed by the different install method proposed by Naim)

julienguy:
pylint now says lines are too long. this thing is a joke. why do we have this? and why does is show one comment at a time? it's my 3rd attempt to pass this test after fixing all the others (after hours to work around extremely rigid tests that are essentially preventing us to do any change to the data format). this seems really not a good use of our time. I also personally don't find that cutting in half the lines to satisfy pylint is making the code easier to read.

iprafols:
Hi @julienguy what you're saying is a bit weird. It usually (or at least used to) give all the complaints at the same time. You can run this locally and get a full report by typing `pylint py/picca/delta_extraction/` from the picca base folder. Also, if you run `yapf --style google file.py -i` in the offending files, most of the issues will go away. 

On another note, if pylint is now only giving one complaint, this is a separate issue that we should fix

iprafols:
I just checked the logs and I can see the entire report. I added a commit to fix the lining

PR #1091, opened 2025-02-11
Read sdss

armengau:
Parallelise file reading in `SdssData.read_from_spec`

- Could also parallelize `read_from_spplate`, but this is not needed for my current purpose (P1D); and it may be more tricky.

### DIFF
--- a/py/picca/delta_extraction/astronomical_objects/sdss_pk1d_forest.py
+++ b/py/picca/delta_extraction/astronomical_objects/sdss_pk1d_forest.py
@@ -3,6 +3,7 @@ in the Pk1D analysis
 """
 from picca.delta_extraction.astronomical_objects.pk1d_forest import Pk1dForest
 from picca.delta_extraction.astronomical_objects.sdss_forest import SdssForest
+from picca.delta_extraction.astronomical_objects.forest import Forest
 
 class SdssPk1dForest(SdssForest, Pk1dForest):
     """Forest Object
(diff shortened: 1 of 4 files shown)


iprafols:
Hi @armengau , maybe it would be more elegant (and not complicated) to modify the Forest.mask_fields in the initialisation of the forests in `py/picca/delta_extraction/data.py`

armengau:
> Hi @armengau , maybe it would be more elegant (and not complicated) to modify the Forest.mask_fields in the initialisation of the forests in `py/picca/delta_extraction/data.py`

Eg., in the code block` if self.analysis_type == "PK 1D"`:
https://github.com/igmhub/picca/blob/9bfc2f61c0ffca0c2fae6a8f1d2b054e0a557013/py/picca/delta_extraction/data.py#L453
add all the lines about mask_fields in `Pk1dForest.consistency_check`?

I can test this...

armengau:
@iprafols I've found, and tested, a better way, similar to what's done in the case of DESI pk1d forest, see the code modif.
Now I'm ready to be merged, if that's fine on your side.

iprafols (approved):

iprafols:
I reviewed the PR and the changes seem fine. The failed test is indeed due to a fix in the spalling of analysis in the error message. The same fix should be applied to the tests. @armengau are you doing it or should I?

armengau:
@iprafols Looks ok now!
Thanks

PR #1094, opened 2025-03-18
Update utils.py with bootstrap covariance estimate

p-slash:
I wrote a simple bootstrap estimator for the covariance matrix for test purposes.


corentinravoux:
@p-slash Do you need a review for this PR ?
The function is good but it is not usable without a bin script, or linking the new functions into the code, do you plan to do that ?

p-slash:
The bootstrap covariance seems to work fine. The results are close to the original covariance estimate. It is somewhat slower than I expected, but time should not be a problem. I just created an option to use bootstrap covariance in picca_export.py

The default is 10000 bootstrap in picca_export.py, which can be changed if desired.

1D plot compares the diagonals of the current covariance to the bootstrap covariance. The differences can be attributed to noise in estimation. 2D plot is the relative difference between two matrices (delta C / v v^T where v is the mean diagonal of both estimates). Here, we see a structure in scale dependence.

image


image

corentinravoux:
Hi @p-slash, this is putting your estimator of the covariance as the default now, shouldn't we do the opposite ?

p-slash:
Would it be ok if I implemented the bias correction to the sample variance estimator? I will revert back the default behavior of doing bootstraps. Then, let's merge this branch if that's ok.

andreicuceu (approved):

Issue #1097, opened 2025-05-12
Editable install does not work as intended


p-slash:
`pip` hard copies the files in the `bin` directory, so changes to the source do not propagate until another `pip install -e` is executed. One solution is to modernize the installation by moving to `setup.cfg` with `[options.entry_points]` and `console_scripts`. This requires moving the `bin` directory inside the `py/picca` folder and listing every executable `setup.cfg`.

Waelthus:
Agreed that this is suboptimally done atm esp regarding the bin files (which also is causing differences between editable and normal installations and iirc was the reason to test on editable ones). But I'd opt for moving the setup info directly to `pyproject.toml` which is the current way recommended for python. Afaik, the main reason we still have a `setup.py` is that it allows getting some info, e.g. the list of all binaries, some additional data files and the git hashes programatically at install time.
If this is a non-issue now, or if people know better ways, I'd be all in favor for changing this.

PR #1098, opened 2025-05-14
[bump minor] Modernize picca installation with pyproject.toml

p-slash:
This PR moves setup.py to pyproject.toml and addresses issue #1097. I also moved the contents of the bin folder to py/picca (that is how I believe this works). I only implemented `picca_cf.py`, `picca_dmat.py`, and `picca_export.py`. I can include other scripts that are actively used. Note, I also preserve the `.py` extension of executables even though that is not the default expectation of pip (hence the quote marks).

This is a tricky change since the files have moved directories. Any PR changing the contents of the bin/ folder should pay close attention to the accuracy of their merge product.


pyproject.toml
```diff
@@ -1,3 +1,61 @@
...
+[tool.setuptools]
+package-dir = { "" = "py" }
+
+[tool.setuptools.packages.find]
+where = ["py"]
+namespaces = true
+
+[tool.setuptools.package-data]
+picca = [
+    "fitter2/models/*/*.fits",
```
Waelthus, line 41: I think we removed fitter2 at some point, so this line could go away

pyproject.toml
```diff
@@ -1,3 +1,61 @@
...
+dynamic = ["dependencies", "version", "readme"]
+description = "Package for Igm Cosmological-Correlations Analyses"
+requires-python = ">=3.7"
+license = "GPL-3.0-or-later"
+license-files = ["LICENSE" ]
+authors = [
+    { name = "Nicolas Busca" },
+    { name = "Helion du Mas des Bourboux" },
+    { name = "Ignasi Pérez-Ràfols", email = "iprafols@gmail.com" },
+    { name = "Michael Walther" },
```
Waelthus, line 16: this could be extended by more collaborators, the Email should probably be the one of a current TG lead

Waelthus (changes requested):
I'm generally in favor of those changes assuming they work fine, maybe one should change the tests as well to install picca in non-editable mode, few minor comments as well

pyproject.toml
p-slash: It is still in setup.py, so this should be addressed separately to avoid obfuscation of changes

pyproject.toml
p-slash: This PR is not responsible for the email list, but only for moving setup.py to pyproject.toml

p-slash:
Michael, 

As a reviewer, you should confirm that the changes work as intended, not just assume. Please confirm that this PR does not break the pipeline. Let me know if more scripts need to be added to pyproject.toml.

Regarding your comments, they unnecessarily increase the scope of the PR and the changes would be obfuscated. We should address them in separate PRs.

pyproject.toml
```diff
@@ -1,3 +1,61 @@
...
+# Distortion
+"picca_dmat.py" = "picca.bin.picca_dmat:main"
+"picca_xdmat.py" = "picca.bin.picca_xdmat:main"
+"picca_fast_metal_dmat.py" = "picca.bin.picca_fast_metal_dmat:main"
+"picca_fast_metal_xdmat.py" = "picca.bin.picca_fast_metal_xdmat:main"
+# P1D
+"picca_Pk1D.py" = "picca.bin.picca_Pk1D:main"
+"picca_Pk1D_cross_exposure.py" = "picca.bin.picca_Pk1D_cross_exposure:main"
+"picca_Pk1D_postprocess.py" = "picca.bin.picca_Pk1D_postprocess:main"
+# Add the rest here
```
Waelthus, line 61: might want to add some more (I guess that needs discussion what people still use), e.g. the following has last commits from <1 year ago, but don't know if still used or if e.g. the functionality went elsewhere:
- picca.bin.picca_Pk1D_average_mean:main

Waelthus (approved):

corentinravoux:
Merging main to test compatibility. Tests ok, launching merging.

corentinravoux:
Please delete your branch if you do not need it

PR #1103, opened 2025-05-26
update pyproject to allow pypi

Waelthus:
Fixes the Pypi uploads by explicitely telling that the Readme is markdown and not rst.


corentinravoux (approved):
Very small change, if it is working this is perfect.

PR #1106, opened 2025-06-23
Binning CF in r, mu grid

p-slash:
This branch introduces the capability to estimate the correlation function in the (r, mu) grid. The multipoles can be exported using `picca_export.py`, although it is unclear if this is the correct place to do so.


p-slash:
Thanks @corentinravoux !

PR #1114, opened 2025-09-15
[bump minor] Clean-up pkg_resources

andreicuceu:
We need to stop using pkg_resources, as it's deprecated. This error has been appearing for a while, and it's now critical that we address it.

`picca/py/picca/constants.py:10: UserWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html. The pkg_resources package is slated for removal as early as 2025-11-30. Refrain from using this package or pin to Setuptools<81.`

As far as I can tell, there are 3 places we use it:
- [py/picca/constants.py](https://github.com/igmhub/picca/blob/f45f5cf22bafa26668475bcece64b0deb1ff9a29/py/picca/constants.py#L10)
- [py/picca/delta_extraction/expected_fluxes/true_continuum.py](https://github.com/igmhub/picca/blob/f45f5cf22bafa26668475bcece64b0deb1ff9a29/py/picca/delta_extraction/expected_fluxes/true_continuum.py#L9)
- [py/picca/tests/test_helpers.py](https://github.com/igmhub/picca/blob/f45f5cf22bafa26668475bcece64b0deb1ff9a29/py/picca/tests/test_helpers.py#L10)

I started this pull request to clean up the first instance because this also gets called from Vega. The part of the code that uses this function is very outdated, so I decided to remove that functionality entirely. This relates to an old blinding setup that was used at the level of the correlation function.

@iprafols could you please have a look at the other two instances? I think in the case of the true continuum, we still need the functionality, so we need to find a replacement. I recommend switching to pathlib, and having something like this function from lyatools: https://github.com/andreicuceu/lyatools/blob/main/lyatools/submit_utils.py#L194-L231. Also tagging @p-slash in case he has knows another clean way to replace it.


andreicuceu:
Note that we'll need to push an update to Vega at the same time as this is merged to maintain compatibility, because Vega directly calls `picca_constants.Cosmo`.

PR #1116, opened 2025-09-23
fixed picca_compute_fvoigt

iprafols:
At the Lya workshop meeting, we found that the script was broken at some point. We took an older version of the script that was working, and fixed a couple of minor things (e.g. added a header to the output file). We believe the script is now working. See plot with the results:
download


corentinravoux:
Is this merged with the script that @andreicuceu  found in NERSC ?

PR #1124, opened 2026-02-08
Template cosmo

calumgordon:
This branch uses a fiducial comology template to compute distances in Picca. The same template can be passed to Vega, making it easy to be consistent in the fiducial cosmologies we use for the catalogue + the fits. It also allows us to store templates for specific runs in Picca.

Note: this currently requires camb to be installed (I haven't added anything in requirements). I'm not necessarily asking for this branch to be merged, but I think it could be a nice idea.


iprafols:
Tests are failing due to the np.in1d function which was removed. I think there was already a PR merged fixing this. We should merge master back into this branch to fix that. 

I'm not sure I understand your comment about camb, I see it added to the requirements.

Otherwise, I think it makes sense to use a templateif it is easier to ensure consistency between picca and vega

corentinravoux:
Now passing test, @iprafols if you agree with the changes we can push to main

iprafols (approved):

PR #1138, opened 2026-05-13
Fix Issue 1137

iprafols:
Description of the changes:

- Added FITORDER keyword in the header FIT_METADATA from delta attributes files
- Removed FITORDER keyword in the header STACK_DELTAS from the delta attributes files
- picca_(x)cf and  picca_dmat now require the delta attributes files in order to read the `order` variable. If not passed, the code looks for it in the standard placing
- I also updated the other scripts (picca_wick,...) that were failing tests. I'm not sure if we decided to remove the tests for these files or not.
- If the `order` variable cannot be read then crash the code (as opposed to continue with `order=1`). This is the main change in behaviour
- In the tests, the fits comparison was not complaining if the new files had new keys in the headers. This is a dangerous behaviour since it means that new things we decide to will to be tested in the future
- I updated the test suite with the changes in the headers
- I updated the correlation tests to include the delta_attribute arguments since the file does not follow the standard placing


py/picca/bin/picca_cf.py
```diff
@@ -343,6 +343,18 @@ def main(cmdargs=None):
...
 
+    parser.add_argument(
+        "--delta-attributes",
+        type=str,
+        default=None,
+        required=False,
+        help=(
+            "Filename for the delta attributes file. This will be used to read the "
+            "order of the polynomial used for the continuum fitting, which is needed "
+            "for the projection of the delta field. If None, the order will be set to 1"
```
andreufont: If None, it will look for the file or crash, right?

andreufont: This is true for all the other scripts, of course.

iprafols: It will crush, yes. Sorry I forgot to change the description since I orignally coded this way. I have pushed a commit fixing this

py/picca/bin/picca_cf.py
```diff
@@ -343,6 +343,18 @@ def main(cmdargs=None):
         "be rebinned by that factor",
     )
 
+    parser.add_argument(
+        "--delta-attributes",
```
andreufont: I noticed that in the script `picca_fast_metal_dmat.py` we already have an input variable called `--in-attributes` that does the same thing. Probably good to use the same variable name?

iprafols: Do you mean changing `--delta-attributes' by ' --in-attributes` in the scripts I changed for this PR? I'm happy to do it, but I wanted to double-check that this is what you are asking

py/picca/delta_extraction/expected_flux.py
```diff
@@ -387,22 +387,13 @@ def hdu_stack_deltas(self, results):
         results: fitsio.FITS
         The open fits file
         """
-        header = fitsio.FITSHDR()
```
andreufont, line 390: Wait, so this information was already stored? 

Could you clarify why the previous way of storing it was not good, and why the new one is? I'm a bit confused now.

py/picca/data.py
```diff
@@ -369,7 +369,7 @@ def __init__(self, los_id, ra, dec, z_qso, plate, mjd, fiberid, log_lambda,
         self.fname = None
 
     @classmethod
-    def from_fitsio(cls, hdu, pk1d_type=False):
+    def from_fitsio(cls, hdu, pk1d_type=False, order=1):
```
andreufont: Are you sure you want to have a default value for order here? It seems problematic. I would have by default `order=None` if anything.

py/picca/data.py
```diff
@@ -378,6 +378,8 @@ def from_fitsio(cls, hdu, pk1d_type=False):
             pk1d_type: bool - default: False
                 Specifies if the fits file is formatted for the 1D Power
                 Spectrum analysis
+            order: int - default: 1
```
andreufont: Same here. Having a default of 1 seems dangerous.

py/picca/bin/picca_cf.py
andreufont: Either that, or change `--in-attributes` to the new variable name, although this seems more annoying for those that had scripts using that already. 

Basically, I thin we should avoid having two variable names for the same thing in different scripts.

py/picca/io.py
```diff
@@ -245,7 +245,7 @@ def read_blinding(in_dir):
     return blinding
 
 
-def read_delta_file(filename, z_min_qso=0, z_max_qso=10, rebin_factor=None):
+def read_delta_file(filename, z_min_qso=0, z_max_qso=10, rebin_factor=None, order=1):
```
andreufont: Same here, I would remove this default value (and pass `None` instead by default)

py/picca/io.py
```diff
@@ -335,7 +339,11 @@ def read_deltas(in_dir,
         z_min_qso: float - default: 0
             Specifies the minimum redshift for QSOs
         z_max_qso: float - default: 10
-            Specifies the maximum redshift for QSOs
+            Specifies thet maximum redshift for QSOs
+        delta_attributes: str or None - default: None
+            Filename for the delta attributes file. This will be used to read the
+            order of the polynomial used for the continuum fitting, which is needed
+            for the projection of the delta field. If None, the order will be set to 1
```
andreufont: If None, it will try to find the file or crash, right?

py/picca/io.py
```diff
@@ -363,7 +371,15 @@ def read_deltas(in_dir,
     if rebin_factor is not None:
         userprint(f"Rebinning deltas by a factor of {rebin_factor}\n")
 
-    arguments = [(f, z_min_qso, z_max_qso, rebin_factor) for f in files]
+    if delta_attributes is None:
+        delta_attributes = in_dir + "../Log/delta_attributes.fits.gz"
+        userprint(f"WARNING: delta_attributes file not given, setting to {delta_attributes}\n")
+    userprint(f"Reading delta attributes from {delta_attributes}\n")
```
andreufont: Should you add here a check that the file exists?

py/picca/bin/picca_cf.py
iprafols: I changed everything to `--in-attributes`

py/picca/delta_extraction/expected_flux.py
iprafols: This was stored in the STACK_DELTAS HDU, but this variable is related to the fit so it does not make sense for it to be stored there. This is why we did not know it was there in the first place. We are now saving the fit results along with this parameter.

py/picca/data.py
iprafols: I replaced the default by `None`

py/picca/data.py
iprafols: same here

py/picca/io.py
iprafols: same here

py/picca/io.py
iprafols: yes

py/picca/io.py
iprafols: if the file does not exist, then we get an OSError and the code crashes as intended so I don't think we need to check the file existance

py/picca/bin/picca_cf.py
```diff
@@ -344,14 +344,15 @@ def main(cmdargs=None):
...
+        "--in-attributes",
         type=str,
         default=None,
         required=False,
         help=(
             "Filename for the delta attributes file. This will be used to read the "
             "order of the polynomial used for the continuum fitting, which is needed "
-            "for the projection of the delta field. If None, the order will be set to 1"
+            "for the projection of the delta field. If None, it will look for the file at the "
+            "standard location and crush if not found "
```
andreufont: crash I'd say

andreufont (commented):
Let me try the branch at NERSC.

We also need to think about new picca_cf runs on old deltas, that do not have the header info. With the new code these would crash... This is annoying, but we might have to default to one afterall. Or add an option `--order` in picca_cf/xcf/dmat/xdmat so that we can recycle old delta files.

py/picca/bin/picca_cf.py
iprafols: I fixed the spelling error

iprafols:
> We also need to think about new picca_cf runs on old deltas, that do not have the header info. With the new code these would crash... This is annoying, but we might have to default to one afterall. Or add an option `--order` in picca_cf/xcf/dmat/xdmat so that we can recycle old delta files.

We could write a more detailed error message with the bit of code to add the missing keyword in the file. I'd need to confirm it, but I'm pretty confident we could even recover the correct value used for that run from the delta_extraction config files that are saved with the results

andreufont:
As we discussed on Zoom, let's try to use the `.config.ini` file to recover this information.

iprafols:
I added a bit of code to enhance backwards compatibility. Pseudo code to explain the workflow:
```
if delta_attributes is None:
   assume standard naming for the delta_attributes file

try to read FITORDER from FIT_METADATA HDU
if file is correct but the keyword is not present (older delta_attributes file)
    1. try to read FITORDER from the STACK_DELTAS 
    if 1 does not work
        2. try to read it from the deltas config file (.config.ini)
        if 2 does not work: crash
if file cannot be opened (e.g. file not found because logs were saved in a non-standard place)
    1. try to read it from the deltas config file (.config.ini)
    if 1 does not work: crash
```

andreufont:
That's great, Ignasi. Is it ready for me to try it out?

andreufont:
@iprafols - I managed to make it work, after doing a minor change to the code:
```
 -        delta_attributes = in_dir + "../Log/delta_attributes.fits.gz"
 +        delta_attributes = in_dir + "/../Log/delta_attributes.fits.gz"
```
With this, it was able to find the value stored in the STACK (these were old deltas), and I could see a message saying "setting the order to 0".

iprafols:
The missing `/` must be because you did not end the `in_dir` variable with it. Having two such symbols is not an issue so this is an easy fix (I already pushed the commit). In the email I received your message said something about an error? What was it? Is it fixed? Did the code work?

andreufont:
All good, the error was only when running things on a terminal, no problem with slurm scripts

andreufont:
@iprafols - should we worry about "true-continuum", or "raw-analysis" runs? In these analyses we should not need to look at the `order` variable (there is no continuum fitting involved), so I'd imagine that there is no "attributes" file associated to it. Would the code crash now? It shouldn't...

Similar for runs using LyCAN. There is no distortion, so we shouldn't need to find the `order` variable.

iprafols:
I have been checking where `order` is used (and how). As expected, it is only used in the distortion matrix computation. Indeed, some analyses will not need it, and the code should not crash, but we still want it to crash if not given when we are computing the distortion matrix. 

I thought of having `order=None` as a default (which I implemented), but the way this is coded, it would mean effectively choosing `order=0` if the distortion matrix is computed. Essentially, the extra term is computed with an `if order==1` clause. Thus, I have added a couple of checks in the computation of the distortion matrix so that the code crashes (and points to the warnings we added) if `order is None`. @andreufont, can you take a look in case I missed some instances?

andreufont:
The variable "order" should be used in two places: in the computation of the distortion matrices, and in the "projection" of deltas. Neither of these should be done in true-continuum or LyCAN analyses. I agree the code should crash if `order is None` in these two functions.

py/picca/io.py
```diff
@@ -256,6 +257,8 @@ def read_delta_file(filename, z_min_qso=0, z_max_qso=10, rebin_factor=None):
             Specifies the maximum redshift for QSOs
         rebin_factor: int - default: None
             Factor to rebin the lambda grid by. If None, no rebinning is done.
+        order: int - default: 1
+            Order of the polynomial used for the continuum fitting. This is needed for the projection of the
```
andreufont: Broken doc string

iprafols:
I had missed the projection, thanks! A similar issue was happening; the code was behaving as `order=0` even when `order=None`. I fixed this and improved the docstrings

py/picca/io.py
```diff
@@ -335,7 +340,11 @@ def read_deltas(in_dir,
         z_min_qso: float - default: 0
             Specifies the minimum redshift for QSOs
         z_max_qso: float - default: 10
-            Specifies the maximum redshift for QSOs
+            Specifies thet maximum redshift for QSOs
+        delta_attributes: str or None - default: None
+            Filename for the delta attributes file. This will be used to read the
+            order of the polynomial used for the continuum fitting, which is needed
+            for the projection of the delta field. If None, the order will be set to 1
```
andreufont: This problem is not fixed yet, right? The docstring still says "If None, the order will be set to 1"

py/picca/io.py
```diff
@@ -363,7 +372,69 @@ def read_deltas(in_dir,
...
+            userprint(f"Setting order={order} for the polynomial used for the continuum fitting")
+        else:
+            order = None
+            userprint("WARNING: `order` not found in delta config file")
+            userprint(
+                "WARNING: Setting order=None, this will lead to an error if the distortion matrix"
+                " is computed")
+            userprint(f"Setting order={order} for the polynomial used for the continuum fitting")
+
+    arguments = [(f, z_min_qso, z_max_qso, rebin_factor, order) for f in files]
```
andreufont, line 464: @iprafols - The code above is great, but I was wondering if we could "hide it" in a separate function, so that the function `read_deltas` is shorter and easier to understand for the reader. 

I believe it should be trivial to have something like 
`order = self.find_order(in_dir, delta_attributes)` 
and then just move these 60 lines of code to that function.

py/picca/data.py
andreufont, line 645: I would probably crash here as well if you are inside `project` with `order = None`. I don't think it should happen (we already have a no-project option for LyCAN or true-continuum, but just in case).

andreufont (commented):
I only added a couple of minor comments on docstrings, a suggested crash point, and if you have time and energy a request for a separate function `def find_order` to hide some of the new code from `read_deltas`.

andreufont (commented):
I confirm that the code works well both for old deltas (guessing the right location) and for new deltas.

py/picca/io.py
iprafols: done

py/picca/io.py
iprafols: fixed

py/picca/io.py
iprafols: fixed

py/picca/data.py
iprafols: I'm confused, in the code I see, this is already crashing (see the beginning of the function). Maybe you were reviewing an earlier version?

andreufont: Yes, it looks like you made changes while I was reviewing. All good now.

andreufont (approved):

PR #1140, opened 2026-05-19
blind DR3 data with new blinding strategy

andreufont:
Now that we will start analyzing data from DR3, it is important to blind this properly. We do not have the templates yet, so for now this should just cause the code to crash for any analysis including DR3 data.


iprafols (approved):
I'm happy with the changes. These changes should make `picca_export` crash at line 412, but we should double-check that this is indeed happening

andreufont:
I've done a quick hack where I have changed (temporarily) the definition of the DR2 / DR3 boundary so that I would get some deltas with DR3 blinding, and I measured correlations from these. They were correctly flagged as blinding = `desi_dr3` and it crashed at the export step with the following error message:
`ValueError: Expected blinding to be one of dict_keys(['desi_y3']). Found desi_dr3.`

So I think we are good to go!

iprafols:
Agreed, I'll set the merge then

PR #1141, opened 2026-05-21
modify picca_export to look for the new dr3 blinding template

andreufont:
As soon as we have a new blinding template (hopefully later today), this will allow us to read it to blind the DR3 correlations at the picca_export stage

### DIFF
--- a/py/picca/bin/picca_export.py
+++ b/py/picca/bin/picca_export.py
@@ -405,7 +405,6 @@ def main(cmdargs=None):
 
     # Check if we need blinding and apply it
     if 'BLIND' in data_name or blinding != 'none':
-        #blinding_dir = '/global/cfs/projectdirs/desi/science/lya/y1-kp6/blinding/'
         blinding_dir = '/global/cfs/projectdirs/desi/science/lya/lya_blinding/bao/'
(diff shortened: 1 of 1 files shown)


andreufont (commented):
Let's wait for the final location of the files, hopefully soon.

iprafols:
The name of the blinding files suggest a date. I would wait to merge until those filenames are confirmed

andreufont:
Yes, absolutely, that's why I marked the PR as "draft".

At this point I just wanted to cross-check that these are the changes that will need to be implemented.

Once we have the final filenames I'll ask you to approve and merge the PR.

iprafols:
Sorry, I was probably not clear enough. Except for the naming of the files I think the changes are fine and are what needs to be implemented.

andreufont (commented):
@iprafols - the final blinding template file is already generated and in the right location.

I think we are ready to merge!

iprafols (approved):
I would maybe remove the commented line pointing to the old path. I think it does not make any sense to keep that if the `blinding_templates` will not work even if that is used. In any case, I think we can merge

andreufont:
@iprafols - Done! 

Can you merge the PR, or do we need to wait for someone else's approval?

corentinravoux:
Let's merge, those are very small change and I want to reformat all the scripts

PR #1145, opened 2026-06-03
[bump minor] Format scripts

corentinravoux:
Format all the scripts, correct errors from a linter and improve the consistency in the way the scripts are launched.

PR #1148, opened 2026-06-17
Make sure that blind_corr_type is the one expected in the blinding template file

andreufont:
Somehow in DR3 the blinding file expects `lyaxqso` and `lybxqso`, but picca_export.py insisted on using `qsoxlya` and `qsoxlyb`. 

I didn't want to modify the template file, and I didn't want to break other people's scripts using picca_export, so I ended up renaming the variable internally in the code before reading the blinding template file.

### DIFF
--- a/py/picca/bin/picca_export.py
+++ b/py/picca/bin/picca_export.py
@@ -429,9 +429,9 @@ def main(cmdargs=None):
             blind_corr_type = args.blind_corr_type
             # match name expected in blinding template file
             if blind_corr_type == 'qsoxlya':
-                blind_corr_type == 'lyaxqso'
+                blind_corr_type = 'lyaxqso'
             if blind_corr_type == 'qsoxlyb':
-                blind_corr_type == 'lybxqso'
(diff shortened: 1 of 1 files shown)


andreufont:
@andreicuceu - this is a tiny PR to fix the issue I mentioned yesterday at the call. I don't think it is worth doing anything fancier at this point, and I checked that this works.

corentinravoux:
Hi @andreufont, sorry but I am doing a formating of all scripts in this PR: https://github.com/igmhub/picca/pull/1145 can you wait for this PR to be pushed and then make your modifications inside that ? 
If that is urgent do not mind that, I will do the merging

andreufont:
@HiramHerrera , did you also find this problem when doing the DR3 Lya BAO analyses? 

If so, I'd say that we should merge now, otherwise we can not use the main branch to work on DR3, and this is dangerous.

corentinravoux:
This PR was merged with the PR https://github.com/igmhub/picca/pull/1145, consider that this is pushed, closing

PR #1152, opened 2026-07-09
Uniqpix [bump minor]

iprafols:
Adding a new class to read from uniqpix formatted data


iprafols:
Note that I have tested that this branch works at NERSC, results are found at `global/cfs/cdirs/desicollab/users/iprafols/lya_working_group/picca/desiuniqpix_test`

py/picca/delta_extraction/quasar_catalogues/desi_quasar_catalogue.py
```diff
@@ -125,6 +140,36 @@ def add_healpix(self):
...
 
+    def add_uniqpix(self, in_dir):
+        """Add uniqpix information to the catalogue"""
+        self.logger.progress("Adding uniqpix information to the catalogue")
+
+        self.catalogue["UNIQPIX"] = np.zeros(len(self.catalogue),
+                                             dtype=np.int64)
+
+        for survey in np.unique(self.catalogue["SURVEY"]):
+            pix2upix_file = f"{in_dir}/{survey}/dark/hpix2upix-{survey}-dark.json"
```
corentinravoux, line 151: Is this general enough ? Are we sure that the format will not change ? Maybe define the format of the hpix to upix converter outside of the functions ?

py/picca/delta_extraction/quasar_catalogues/desi_quasar_catalogue.py
```diff
@@ -13,16 +15,17 @@
 from picca.delta_extraction.utils import update_accepted_options
 
 accepted_options = update_accepted_options(
-    accepted_options,
-    ["catalogue", "in_nside", "keep surveys"])
+    accepted_options, ["catalogue", "in_nside", "keep surveys"])
 
 defaults = {
+    "add uniqpix": "False",
```
corentinravoux, line 21: Would it be possible to not set this variable by hand ? I think it might be better to set this variable to True automatically when the uniqpix mode is used

corentinravoux:
Hi @iprafols thanks for the work, did you check that the deltas were the same as DR3, or at least very similar ?

corentinravoux:
We can use this PR to bump the version to 9.21 if that is ok for you

py/picca/delta_extraction/quasar_catalogues/desi_quasar_catalogue.py
iprafols: In principle the information we have is that this will not change and similar files will be included in all the DR from now on, so I think this should be fine

iprafols:
> Hi @iprafols thanks for the work, did you check that the deltas were the same as DR3, or at least very similar ?

I checked the logs to make sure there were no reading issues. Given that the file format is the same, and that the only thing changing is the location of the file, I think this is enough to ensure that the results are unchanged. Keep in mind that I did not even change the reading function, I simply added the uniqpix computation and updated the file paths.

iprafols:
> We can use this PR to bump the version to 9.21 if that is ok for you

Sure, I added the [bump minor] tag at least so that we don't forget. I'm not sure if the automatic version update is still working or not. If it still works, then the version should be updated automatically upon merging this PR

corentinravoux:
No the bumpversion is not working anymore, I will work on that at some point.

corentinravoux (approved):

corentinravoux:
Nice PR, let's push that and I will incorporate that in the MPI new PR.
Can you just manually change the version number please ?

py/picca/delta_extraction/quasar_catalogues/desi_quasar_catalogue.py
iprafols: There is a 
```
# set add uniqpix flag
config["add uniqpix"] = "True"
```
block inside `DesiUniqPix` that automatically sets it, but I did not want it to be set in general so that we can still work with the other modes

iprafols:
@corentinravoux , pushed version update

PR #1154, opened 2026-07-20
First test of vibe coded MPI version of picca (use of opus 4.8)

### LINKED ISSUE #1146: Random OOM problems when running picca_delta_extraction on the calibration region
As discussed on Slack a few times, quite often one runs into OOM errors when trying to run picca_delta_extraction on the CIII region that we use for calibration. This is the step that uses more quasars (>2M in DR2), so it is not a surprise that we only see it here. 

It has been suggested by @corentinravoux and @m-herbold that one might be able to avoid this by using the `exclusive` flag in the SLURM scripts, but I don't understand how this would help. 

As far as I understand, when you ask for a CPU node using the `regular` queue (as we always do in picca_delta_extraction), then you do get the entire node, you are not sharing it with others that might use some of your memory.

With DR3 in mind, I think we should find a stable solution to this problem.

corentinravoux:
Testing running picca in MPI, need for speed improvement

### DIFF
--- a/py/picca/delta_extraction/data_catalogues/desi_healpix_mpi.py
+++ b/py/picca/delta_extraction/data_catalogues/desi_healpix_mpi.py
@@ -1,0 +1,234 @@
+"""This module defines the class DesiHealpixMpi, the MPI-parallel reader for
+DESI healpix data.
+
+The workspace is decomposed by healpix: rank ``r`` owns every reading healpix
+``h`` with ``h % size == r`` and reads, corrects, masks, filters, continuum-fits
+and writes only the forests of the pixels it owns. Because all the surveys of a
+given healpix are owned by the same rank, forests sharing a targetid are coadded
+locally exactly as in the serial reader, so no communication is needed while
+reading.
+
+The ranks only communicate at the points where a genuinely global quantity is
+required:
+
+- ``find_nside``:   the automatic nside needs the global object count, so the
+                    coordinates are gathered on rank 0, the nside is decided
+                    there and broadcast back;
+- the continuum fit: the three sample-wide stacks (mean continuum, variance
+                    functions, delta stack) are Allreduce-summed. This is handled
+                    entirely by Dr16ExpectedFluxMpi through the ``reduce_sum``
+                    hook, not here;
+- ``save_deltas``:  the output nside is coarser than the reading nside, so a
+                    coarse output pixel can gather forests from several reading
+                    pixels living on different ranks. The forests are therefore
+                    redistributed by output healpix so that each rank writes a
+                    disjoint set of delta files.
+"""
+import os
+
+import healpy
+import numpy as np
+
+from picca.delta_extraction.data_catalogues.desi_data import (
+    merge_new_forest,
+    verify_exposures_shape,
+)
+from picca.delta_extraction.data_catalogues.desi_healpix import (  # pylint: disable=unused-import
+    DesiHealpix,
+    DesiHealpixFileHandler,
+    accepted_options,
+    defaults,
+)
+from picca.delta_extraction.errors import DataError
+from picca.delta_extraction.utils_mpi import get_comm, redistribute_by_key
+
+
+class DesiHealpixMpi(DesiHealpix):
+    """Reads the spectra from DESI using healpix mode, decomposing the healpix
+    pixels over MPI ranks.
+
+    Only the parallelisation strategy differs from DesiHealpix: get_filename and
+    the PK 1D setup are inherited, read_data reads the pixels owned by this rank,
+    find_nside and save_deltas add the required communication.
+
+    Methods
+    -------
+    (see DesiHealpix in py/picca/delta_extraction/data_catalogues/desi_healpix.py)
+    __init__
+    read_data
+    find_nside
+    save_deltas
+
+    Attributes
+    ----------
+    (see DesiHealpix in py/picca/delta_extraction/data_catalogues/desi_healpix.py)
+
+    comm: mpi4py.MPI.Comm
+    The world communicator
+
+    mpi_rank: int
+    Rank of the current process
+
+    mpi_size: int
+    Number of MPI processes
+    """
+
+    def __init__(self, config):
+        """Initialize class instance
+
+        Arguments
+        ---------
+        config: configparser.SectionProxy
+        Parsed options to initialize class
+        """
+        self.comm = get_comm()
+        self.mpi_rank = self.comm.Get_rank()
+        self.mpi_size = self.comm.Get_size()
+
+        # DesiHealpix.__init__ reads the catalogue and calls read_data (below)
+        super().__init__(config)
+
+        # write a per-rank rejection log so that the ranks do not write to the
+        # same file concurrently
+        if self.rejection_log is not None:
+            directory = os.path.dirname(self.rejection_log.file)
+            basename = os.path.basename(self.rejection_log.file)
+            name = basename.split(".")[0]
+            extension = basename[len(name):]
+            self.rejection_log.file = os.path.join(
+                directory, f"{name}_rank{self.mpi_rank}{extension}")
+
+    def read_data(self):
+        """Read the data owned by this rank.
+
+        Every rank builds the same list of files, then reads only the files
+        whose healpix it owns (``healpix % size == rank``). Forests are not
+        gathered: each rank keeps and processes its own subset.
+
+        Return
+        ------
+        is_mock: bool
+        False for DESI data and True for mocks
+
+        Raise
+        -----
+        DataError if no quasars were found on any rank
+        """
+        grouped_catalogue = self.catalogue.group_by(["HEALPIX", "SURVEY"])
+
+        is_mock = False
+
+        arguments = []
+        healpix_of_file = []
+        for group in grouped_catalogue.groups:
+            healpix, survey = group["HEALPIX", "SURVEY"][0]
+
+            filename, is_mock_aux = self.get_filename(survey, healpix)
+            if is_mock_aux:
+                is_mock = True
+
+            arguments.append((filename, group))
+            healpix_of_file.append(int(healpix))
+
+        num_owned = sum(healpix % self.mpi_size == self.mpi_rank
+                        for healpix in healpix_of_file)
+        if self.mpi_rank == 0:
+            self.logger.info(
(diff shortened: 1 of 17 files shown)


corentinravoux:
Waiting on NERSC to be back to finish this

corentinravoux:
Full report for comparison between MPI and non-MPI: 
A: /global/cfs/cdirs/desi/users/ravouxco/pk1d/desi_data/matterhorn/p1d_healpix/v0/data/p1d_out/p1d_lya_noisepipeline_linesDESIEDR_catcatmatterhorn_v0_nobalBI_dlacombined_gp_cnn_toolkit_v0_balai_v0_mpi/Delta  (1252 files)
B: /global/cfs/cdirs/desi/users/ravouxco/pk1d/desi_data/matterhorn/p1d_healpix/v0/data/p1d_out/p1d_lya_noisepipeline_linesDESIEDR_catcatmatterhorn_v0_nobalBI_dlacombined_gp_cnn_toolkit_v0_balai_v0/Delta  (1252 files)
comparing 1252 common files with rtol=1e-05 atol=1e-07 on 32 processes

[97/1252] DIFF delta-1125.fits.gz
    LOS 39628138027683606 / DELTA_BLIND: 1/60 pix differ, max|abs|=1.596e-07 max|rel|=3.597e-05
[136/1252] DIFF delta-1183.fits.gz
    LOS 39628119925066217 / DELTA_BLIND: 1/60 pix differ, max|abs|=1.229e-07 max|rel|=6.626e-05
[245/1252] DIFF delta-1316.fits.gz
    LOS 39627965901836575 / DELTA_BLIND: 1/54 pix differ, max|abs|=3.449e-07 max|rel|=1.489e-04
[333/1252] DIFF delta-1441.fits.gz
    LOS 39627821072523729 / DELTA_BLIND: 1/57 pix differ, max|abs|=1.663e-07 max|rel|=3.044e-05
[1112/1252] DIFF delta-803.fits.gz
    LOS 39628430626524475 / DELTA_BLIND: 1/67 pix differ, max|abs|=1.191e-07 max|rel|=7.967e-05

======================================================================
files compared      : 1252
lines of sight      : 507940
files with a diff   : 5
files only in A / B : 0 / 0
result              : DIFFERENT

corentinravoux:
Delta compared (MPI vs no MPI version), log fixed. Only need one review

py/picca/delta_extraction/data_catalogues/desi_healpix_mpi.py
```diff
@@ -0,0 +1,260 @@
...
+                                        self.keep_single_exposures,
+                                        self.logger)
+
+        # Read only the files of the healpix pixels owned by this rank. The
+        # global file index is kept so that the keep_single_exposures keys stay
+        # globally unique.
+        forests_by_targetid = {}
+        for index, (this_arg, healpix) in enumerate(
+                zip(arguments, healpix_of_file)):
+            if healpix % self.mpi_size != self.mpi_rank:
```
iprafols: I'm wondering whether it would be better to move the healpix check to the previous loop in line 123 where the argument list is created. Then we should replace line 133 with a simple `len(arguments)`, and line 137 with a `grouped_catalogue.ngroups`. 

Doing it this way, we perform the same number of `if` checks (but they are moved to the first loop). The second loop is shortened, and we also save some of the list allocations for variables `arguments` and `healpix_of_file`. Overall, we should be a bit more efficient in each of the runners.

py/picca/delta_extraction/data_catalogues/desi_healpix_mpi.py
```diff
@@ -0,0 +1,260 @@
...
+                continue
+            forests_by_targetid_aux, _ = reader(this_arg)
+            if self.use_non_coadded_spectra & self.keep_single_exposures:
```
iprafols: Quick question. Would it be interesting to implement the MPI for class DesiHealpixFast as well? I was checking the differences and the all the checks that follows here were dropped in that class. These are probably needed for P1D analysis, but we did gain significant time for the BAO analysis when running with the optimized class.

py/picca/delta_extraction/data_catalogues/desi_healpix_mpi.py
```diff
@@ -0,0 +1,260 @@
...
+                    for key, items in forests_by_targetid_aux.items()
+                }
+            merge_new_forest(forests_by_targetid, forests_by_targetid_aux)
+
+        if self.use_non_coadded_spectra & self.keep_single_exposures:
+            forests_by_targetid = verify_exposures_shape(forests_by_targetid)
+
+        self.forests = list(forests_by_targetid.values())
+
+        # A given rank may legitimately own no forest; only raise if the whole
```
iprafols: What happens to the rank if it has no forests? Is it flagged or something? Or does it die? I'm thinking more about the later steps of the analysis

py/picca/delta_extraction/data_catalogues/desi_healpix_mpi.py
```diff
@@ -0,0 +1,260 @@
...
+        gathered = self.comm.gather(messages, root=0)
+        if self.mpi_rank == 0:
+            for rank_messages in gathered:
+                for message in rank_messages:
+                    self.logger.progress(message)
+        else:
+            for message in messages:
+                self.logger.progress(message)
+
+    def find_nside(self):
```
iprafols: I believe this function is no longer used anywhere, so maybe we can drop it? Particularly since it requires communication between the ranks. Or am I wrong?

py/picca/delta_extraction/expected_fluxes/mpi.py
```diff
@@ -0,0 +1,165 @@
+"""MPI-parallel expected flux: a mixin plus a factory that generates the MPI
```
iprafols: This is an interesting way to have all ExpectedFlux classes converted to MPI. We should check that it works for all the classes (in particular we should test `FixedFudgeDr16ExpectedFlux`,  `TrueContinuum `, and `MeanContinuumInterpExpectedFlux`

py/picca/delta_extraction/expected_fluxes/mpi.py
```diff
@@ -0,0 +1,165 @@
...
+
+from picca.delta_extraction.expected_flux import ExpectedFlux
+from picca.delta_extraction.utils_mpi import allreduce_sum, get_comm
+
+MPI_SUFFIX = "Mpi"
+
+# cache of already-built classes, keyed by class name
+_MPI_CLASSES = {}
+
+
```
iprafols: We should add here a list of classes that are not going to be MPIed (empty for now), but with a comment specifying that we should add here the list of classes that do not follow the same 'factory logic'. I'm thinking about potential classes that compute the quasar continua differently (e.g. with lycan)

iprafols: These classes should raise an Error (or a Warning) when `build_mpi_expected_flux` function is called

py/picca/delta_extraction/survey.py
```diff
@@ -206,16 +211,42 @@ def load_config(self, config_file):
         self.config = Config(config_file)
         self.num_processors = self.config.num_processors
 
+    def _read_shared(self, build):
```
iprafols: The presence of this function means we might need to rethink the masks and corrections (but mostly the masks). 

For the masks, there are two parts: reading and application. In some cases (such as the BAL masking or the DLA masking), we read basic parameters, then compute the array that needs to be applied, and then we apply it. The computation can be performed either at the reading time or at the application time. In the MPI logic, we should move all the computation to the application time so that it is run in parallel.

iprafols: I'm pointing this so that we don't forget, but this could be addressed in a follow-up PR

py/picca/tests/delta_extraction/mpi_tests.py
```diff
@@ -0,0 +1,82 @@
...
+        ]
+        print("Running command: ", " ".join(command))
+
+        try:
+            subprocess.run(command, check=True, capture_output=True,
+                           env=dict(os.environ, THIS_DIR=THIS_DIR,
+                                    PYTHONPATH=PY_DIR))
+        except CalledProcessError as e:
+            print(e.stderr)
+            raise e
```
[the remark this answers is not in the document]
iprafols: I agree with the suggestion

py/picca/tests/delta_extraction/mpi_tests.py
```diff
@@ -0,0 +1,82 @@
...
+
+        # compare attributes
+        test_files = sorted(glob.glob(f"{test_dir}/Log/delta_attributes*.fits.gz"))
+        out_files = sorted(glob.glob(f"{out_dir}/Log/delta_attributes*.fits.gz"))
+        for test_file, out_file in zip(test_files, out_files):
+            self.assertTrue(test_file.split("/")[-1] == out_file.split("/")[-1])
+            self.compare_fits(test_file, out_file)
+
```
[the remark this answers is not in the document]
iprafols: It doesn't hurt to add this check

py/picca/tests/delta_extraction/mpi_tests.py
```diff
@@ -0,0 +1,82 @@
...
+        # compare deltas
+        test_files = sorted(glob.glob(f"{test_dir}/Delta/delta-*.fits.gz"))
+        out_files = sorted(glob.glob(f"{out_dir}/Delta/delta-*.fits.gz"))
+        self.assertTrue(len(out_files) == len(test_files))
+        for test_file, out_file in zip(test_files, out_files):
+            self.assertTrue(test_file.split("/")[-1] == out_file.split("/")[-1])
+            self.compare_fits(test_file, out_file)
```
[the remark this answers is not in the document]
iprafols: I agree with the suggested changes

iprafols (commented):
Overalll I think it looks great, I have a few comments, but I think it is almost ready to merge. However, before doing that, I would test the MPI factory for ExpectedFlux classes with the different classes (not necessarily all of them, I think testing with `FixedFudgeDr16ExpectedFlux`, `TrueContinuum`,  and `MeanContinuumInterpExpectedFlux` would be enough

iprafols:
Once we fix the minor comments and are ready to do the final testing, I can easily test the class `MeanContinuumInterpExpectedFlux` since I already have a DR2 run for this class. I guess for the others we can use the normal DR2 run

